PR #8, opened 2019-01-09
validating cloud creds

joelddiaz:
add a controller to annotate the secret containing the cloud creds

possible values: mint, passthrough, insufficient for the cases where we can mint new creds, where the creds are good enough to passthrough (based on the static list of credential requests manifests), and creds that simply aren't good enough


pkg/controller/utils/utils.go
```diff
@@ -0,0 +1,163 @@
...
+		"iam:CreateAccessKey",
+		"iam:CreateUser",
+		"iam:DeleteAccessKey",
+		"iam:DeleteUser",
+		"iam:DeleteUserPolicy",
+		"iam:GetUser",
+		"iam:ListAccessKeys",
+		"iam:PutUserPolicy",
+		"iam:TagUser",
+	}
```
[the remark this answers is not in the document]
joelddiaz: Just searched through the actuator for all the aws calls and came up with that list. Will add a TODO to do a live test with only those perms...

pkg/controller/credentialsrequest/actuator/actuator.go
```diff
@@ -32,4 +32,8 @@ type Actuator interface {
 	Update(context.Context, *minterv1.CredentialsRequest) error
 	// Checks if the credentials currently exists.
 	Exists(context.Context, *minterv1.CredentialsRequest) (bool, error)
+	// NeedsUpdate will answer the question whether there is a need to update credentials
+	// based on whether any creds exist, or whether the existing creds no longer
+	// satisfy the CredentialsRequest
+	NeedsUpdate(context.Context, *minterv1.CredentialsRequest) (bool, error)
```
dgoodwin: If this is working we can probably go with it for now. 

Ultimately I'm worried this shouldn't be done in the controller and actuator interface, I don't think it's a given that the next cloud we implement will want to follow the same paradigm with passthrough creds, and it's probably better handled within the actuator itself.

However we don't have time right now, so if this is working we can go with it and investigate later.

pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -274,28 +277,45 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
 		logger.Debug("found secret namespace")
 	}
 
-	// FIXME / TODO: check for proper secret annotation
+	// Now figure out whether we need to create new creds/secrets to satisfy the CredentialsRequest
 
-	// Hand over to the actuator:
-	exists, err := r.Actuator.Exists(context.TODO(), cr)
+	updateNeeded, err := r.Actuator.NeedsUpdate(context.TODO(), cr)
```
dgoodwin: This will catch things like the access key being deleted in AWS, and the policy modified in AWS correct?

What would happen if we removed permissions, or if someone went to AWS and manually granted that access key elevated privs? My suspicion is that because we can still do what is necessary, the code will never correct the problem. If so I fear that needs fixing, teams that will use this will need to iterate on the perms they need, if they remove one they're going to expect it gets removed in AWS, and not doing so is going to cause bugs where they may think they didn't need it but actually did. Can we re-incorporate comparison of policy documents somewhere?

pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -315,6 +335,101 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
...
+				return nil, err
+			}
+		}
+	} else {
+		logger.Debug("Creating a secret")
+
+		newSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      cr.Spec.SecretRef.Name,
+				Namespace: cr.Spec.SecretRef.Namespace,
```
dgoodwin: Needs to have minterv1.AnnotationCredentialsRequest annotation.

pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -315,6 +335,101 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
...
+		}
+
+		// set controller reference
+		if err := controllerutil.SetControllerReference(cr, newSecret, scheme.Scheme); err != nil {
+			logger.WithError(err).Error("error setting controller reference on secret")
+			return nil, err
+		}
+
+		// create secret
+		if err := r.Create(context.Background(), newSecret); err != nil {
```
dgoodwin: r.Client.Create I'm guessing.

I'm nervous that this is bypassing the actuator entirely though.

pkg/controller/credentialsrequest/credentialsrequest_controller.go
joelddiaz: will change to actually check the policy docs

pkg/controller/credentialsrequest/credentialsrequest_controller.go
joelddiaz: r.Create works because we've dumped the kubeclient straight into the struct
```
type ReconcileCredentialsRequest struct {
	client.Client
	Actuator actuator.Actuator
}
```

dgoodwin: Aha I see, wasn't aware we could do that.

dgoodwin:
/lgtm

We'll refactor in a followup but we should get this in for feature free, testing here looks good so far.

PR #9, opened 2019-01-10
WIP: Integrate cloud credential operator into release image payload

dgoodwin:
Adds the manifests dir with appropriate contents for the release image. 

WIP until I can figure out how to disable everything if not in an AWS cluster.


manifests/01-deployment.yaml
```diff
@@ -0,0 +1,115 @@
...
+  strategy:
+    type: Rolling
+  template:
+    metadata:
+      labels:
+        control-plane: controller-manager
+        controller-tools.k8s.io: "1.0"
+    spec:
+      containers:
+      - image: quay.io/dgoodwin/cloud-credential-operator:latest
```
csrwng:
```suggestion
      - image: quay.io/openshift/cloud-credential-operator:latest
```

manifests/01-deployment.yaml
```diff
@@ -0,0 +1,115 @@
...
+---
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
+metadata:
+  labels:
+    control-plane: controller-manager
+    controller-tools.k8s.io: "1.0"
+  name: cloud-credential-operator
+  namespace: openshift-cloud-credential-operator
+spec:
```
csrwng:
```suggestion
spec:
  priorityClassName: system-cluster-critical
  nodeSelector:
    node-role.kubernetes.io/master: ""
```

manifests/01-deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+  ports:
+  - port: 443
+  selector:
+    control-plane: controller-manager
+    controller-tools.k8s.io: "1.0"
+---
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
```
csrwng: You should likely use a Deployment instead of DeploymentConfig. It makes it possible to run your operator before the openshift apiserver is up.

pkg/controller/controller.go
```diff
@@ -37,14 +53,69 @@ func AddToManager(m manager.Manager) error {
...
+	yamlStr := string(data)
+	log.Warn(yamlStr)
+	match, err := regexp.MatchString(`.*platform:\s*aws:.*`, yamlStr)
+	if err != nil {
+		return false, err
+	}
+	return match, nil
+}
+
+func getClient() (client.Client, error) {
```
csrwng: Your preference, but it seems simpler to create a regular kubernetes clientset instance as in:

https://github.com/openshift/master-dns-operator/blob/4560a00ecc516ffefa2f2798d2296df3feb77daf/pkg/operator/masterdns/operator.go#L111

manifests/01-deployment.yaml
dgoodwin: Changed to quay.io/openshift/origin-cloud-credential-operator:latest

pkg/controller/controller.go
dgoodwin: I was worried GetConfig wouldn't be initialized just as GetClient wasn't but I didn't actually try.

dgoodwin:
Closing to get rid of the e2e-aws-operator suite that's bad. Reopening momentarily.

PR #11, opened 2019-01-10
Integrate into release image

dgoodwin:


dgoodwin:
Ready for another review @csrwng , e2e-aws was green just a second ago but pushed some more minor changes.

csrwng:
@dgoodwin this looks good to me, is it still WIP?

dgoodwin:
/hold

@csrwng could you take a quick look again, just had to update for conflicts from Joel's PR that merged first. I'm going to request an exception for this.

pkg/controller/controller.go
```diff
@@ -37,14 +53,69 @@ func AddToManager(m manager.Manager) error {
...
+	if err != nil {
+		return false, err
+	}
+	data, ok := installCfgMap.Data["install-config"]
+	if !ok {
+		return false, fmt.Errorf("cluster-config-v1 ConfigMap did not contain install-config data")
+	}
+	yamlStr := string(data)
+	log.Warn(yamlStr)
+	match, err := regexp.MatchString(`.*platform:\s*aws:.*`, yamlStr)
```
csrwng: This feels fragile. Why not [deserialize it](https://github.com/openshift/master-dns-operator/blob/master/pkg/operator/masterdns/operator.go#L490-L500), or even better [use the infrastructure config](https://github.com/openshift/cluster-ingress-operator/blob/master/cmd/cluster-ingress-operator/main.go#L52-L56)

csrwng:
@dgoodwin just one comment, otherwise lgtm

pkg/controller/controller.go
dgoodwin: I didn't want to deserialize as it introduces a circular dependency, we'd pull the installer and the installer will soon be pulling this project for checking creds.

It's also very temporary, infrastructure config is the long term solution but it's not being populated yet, will eventually switch to that. (there should be a TODO in here somewhere I thought, I will check)

pkg/controller/controller.go
```diff
@@ -37,14 +53,69 @@ func AddToManager(m manager.Manager) error {
 		}
 	}
 	for _, f := range AddToManagerWithActuatorFuncs {
-		// TODO: cleanup hard coded AWS instantiation here:
-		awsActuator, err := awsactuator.NewAWSActuator(m.GetClient(), scheme.Scheme)
+		// Check if this is an AWS cluster and if not, add a dummy actuator:
+		// TODO: Use infrastructure type to determine this in future, it's not being populated yet:
+		// https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L11
```
dgoodwin: @csrwng TODO here ^^ Infra is definitely the way to go.

PR #13, opened 2019-01-18
Publish a ClusterOperator with conditions for status

dgoodwin:
This will expand once we have error conditions on CredentialsRequests.


pkg/controller/credentialsrequest/status.go
```diff
@@ -0,0 +1,142 @@
...
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	cloudCredOperatorNamespace = "openshift-cloud-credential-operator"
+)
+
+// syncOperatorStatus computes the operator's current status and therefrom
```
joelddiaz: s/therefrom/from there/ ?

pkg/controller/credentialsrequest/status.go
```diff
@@ -0,0 +1,142 @@
...
+	for _, cr := range credRequests {
+		if !cr.Status.Provisioned {
+			credRequestsNotProvisioned = credRequestsNotProvisioned + 1
+		}
+	}
+	if credRequestsNotProvisioned > 0 {
+		progressingCondition.Status = configv1.ConditionTrue
+		progressingCondition.Reason = "Reconciling"
+		progressingCondition.Message = fmt.Sprintf(
+			"%d/%d credentials requests provisioned",
```
joelddiaz: maybe "%d of %d credentials..." instead of the '/' to be more clear

dgoodwin:
Updated, thanks!

dgoodwin:
ready for another look @joelddiaz

dgoodwin:
Should be ready for another round of review.

pkg/controller/credentialsrequest/status.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	failingCredRequests := 0
+
+	for _, cr := range credRequests {
+		// Check for provision failure conditions:
+		foundFailure := false
+		for _, t := range failureConditionTypes {
+			failureCond := utils.FindCredentialsRequestCondition(cr.Status.Conditions, t)
+			if failureCond != nil && failureCond.Status == corev1.ConditionTrue {
+				foundFailure = true
```
csrwng: really small nit: break here

dgoodwin: done!

dgoodwin:
Finally passed, @csrwng can I get anotherr lgtm, quickly. :) There was a change to add a missing list permission for caching.

Dockerfile
```diff
@@ -10,6 +10,7 @@ WORKDIR /go/src/github.com/openshift/cloud-credential-operator
 COPY pkg/    pkg/
```
abhinavdahiya: why not `COPY . .` ?

pkg/util/clusteroperator/status.go
```diff
@@ -0,0 +1,65 @@
+package clusteroperator
```
abhinavdahiya: you can use `https://github.com/openshift/library-go/tree/master/pkg/config/clusteroperator/v1helpers`

dgoodwin:
@abhinavdahiya will look into those as a followup, this has been open too long and need to get it merged if CI will work. (spoiler alert, it won't)

PR #15, opened 2019-01-18
add conditions for credentials requests

joelddiaz:
capture error conditions into CredentialsRequest.Status.Conditions[]
    
First set of errors as follows:
InsufficientCloudCredentials
MissingTargetNamespace
CredentialsGrantFailure
CredentialsDeprovisionFailure
    
the workflow is to define failing conditions, and not bother to update the conditions whenever everything is working well. only populate a condition if it is firing (ie failing), and if the condition stops firing, then update the condition to indicate that it has been resolved.
    
this code closely follows the pattern set in the previous cluster-operator repo.


pkg/controller/credentialsrequest/credentialsrequest_controller_test.go
```diff
@@ -443,6 +444,97 @@ func TestCredentialsRequestReconcile(t *testing.T) {
...
+			mockRootAWSClient: func(mockCtrl *gomock.Controller) *mockaws.MockClient {
+				mockAWSClient := mockaws.NewMockClient(mockCtrl)
+				return mockAWSClient
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cr := getCR(c)
+				assert.NotNil(t, cr.Status)
+				assert.NotNil(t, cr.Status.Conditions)
+				condition := utils.FindCredentialsRequestCondition(cr.Status.Conditions, minterv1.MissingTargetNamespace)
+				assert.NotNil(t, condition)
```
dgoodwin: We should probably be checking the state of the condition, possibly the reason as well. Would it be worth adding test.expectedConditions with a struct encompassing the type, reason, and state, and move this validate logic into the test loop itself so it's run for every test that specifies expected conditions.

dgoodwin: Missed one use of expectedConditions I think. Looks good otherwise. Lets squash and hope for a test pass.

PR #18, opened 2019-01-21
WIP: Orphaned users investigation

dgoodwin:
Not ready yet but these are some ideas so far to help pinpoint what's going wrong with orphaned users in CI, as well as excessive API calls.

CC @smarterclayton


pkg/aws/actuator/actuator.go
```diff
@@ -639,6 +697,8 @@ func (a *AWSActuator) buildReadAWSClient(cr *minterv1.CredentialsRequest) (minte
 	}
 
 	logger.Debug("creating read AWS client")
+	//a.AWSClientBuilder(accessKeyID, secretAccessKey)
+	//return nil, fmt.Errorf("test")
```
csrwng: Delete the above comments

dgoodwin:
@csrwng updated with a couple new commits.

PR #26, opened 2019-01-30
Remove secret controller refs.

dgoodwin:
Attempted hotloop fix. These controller refs are not namespaced, but
link to a controller (the CredentialsRequest) in another namespace. This
appeared to work fine for cleanup etc, but I suspect it may be the cause
of the post-upgrade hotloop where the controller constantly reconciles
and creates, then re-syncs and sees no secret.

Instead we will delete the target secrets on CredentialsRequest deletion
ourselves.


dgoodwin:
Attempted fix for: https://github.com/openshift/cloud-credential-operator/issues/24

sjenning:
confirmed this fixes #24

PR #27, opened 2019-02-01
Fix OOM by increasing memory limits for controller.

dgoodwin:
QE hit an oom during controller startup in https://bugzilla.redhat.com/show_bug.cgi?id=1670833, I suspect due to quite a few namespaces.


dgoodwin:
Interesting thanks, I bumped that up from 70 to 150.

PR #30, opened 2019-02-13
document scaling down in-cluster bits

joelddiaz:


README.md
```diff
@@ -39,6 +39,13 @@ $ make install
 $ make run
 ```
 
+NOTE: To keep the in-cluster versions of the code from conflicting with your local copy, you should scale down the deployments for cloud-credential-operator and cluster-version-operator
+
+```
+$ oc patch -n openshift-cloud-credential-operator deployment cloud-credential-operator -p '{"spec":{"replicas": 0}}'
+$ oc patch -n openshift-cluster-version deployment cluster-version-operator -p '{"spec":{"replicas": 0}}'
```
dgoodwin: Would this look better? 

`kubectl scale -n openshift-cloud-credential-operator deployment.v1.apps/cloud-credential-operator  --replicas=0`

PR #32, opened 2019-02-14
refactor credentialsrequest controller

joelddiaz:
move actuator logic into the actuator code (and reuse existing functionality where it makes sense)

create ActuatorError to allow passing specific error conditions back to the credentialsrequest controller (to allow setting specific Conditions as needed)

allow status.conditions to be empty (nothing wrong with everything healthy)

add GetUserPolicy to list of aws actions needed for the actuator in mint-mode

unrelated deployment requests/limits (from make generate)

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -30,6 +30,7 @@ import (
 	ccaws "github.com/openshift/cloud-credential-operator/pkg/aws"
 	minteraws "github.com/openshift/cloud-credential-operator/pkg/aws"
 	actuatoriface "github.com/openshift/cloud-credential-operator/pkg/controller/credentialsrequest/actuator"
+	"github.com/openshift/cloud-credential-operator/pkg/controller/secretannotator"
 	"github.com/openshift/cloud-credential-operator/pkg/controller/utils"
 
 	openshiftapiv1 "github.com/openshift/api/config/v1"
@@ -134,9 +135,9 @@ func (a *AWSActuator) Exists(ctx context.Context, cr *minterv1.CredentialsReques
 
 }
 
-// NeedsUpdate will return whether the current credentials satisfy what's being requested
+// needsUpdate will return whether the current credentials satisfy what's being requested
 // in the CredentialsRequest
-func (a *AWSActuator) NeedsUpdate(ctx context.Context, cr *minterv1.CredentialsRequest) (bool, error) {
+func (a *AWSActuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequest) (bool, error) {
 	logger := a.getLogger(cr)
 	// If the secret simply doesn't exist, we definitely need an update
 	exists, err := a.Exists(ctx, cr)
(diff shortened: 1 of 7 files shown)


pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -366,68 +366,41 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
...
+		syncErr = r.Actuator.Create(context.TODO(), cr)
+	} else {
+		syncErr = r.Actuator.Update(context.TODO(), cr)
+	}
+	if syncErr != nil {
+		logger.Errorf("error syncing credentials: %v", err)
+
+		switch t := syncErr.(type) {
+		case actuator.ActuatorStatus:
+			logger.Debugf("errored with condition: %v", t.Reason())
```
dgoodwin: Errorf

pkg/aws/actuator/actuator.go
```diff
@@ -263,11 +264,82 @@ func (a *AWSActuator) getProviderSpec(cr *minterv1.CredentialsRequest) (*minterv
...
+	if err != nil {
+		logger.WithError(err).Error("error determining whether a credentials update is needed")
+		return &actuatoriface.ActuatorError{
+			ErrReason: minterv1.CredentialsProvisionFailure,
+			Message:   fmt.Sprintf("error determining whether a credentials update is needed: %v", err),
+		}
+	}
+
+	if !needsUpdate {
+		return nil
```
dgoodwin: log.Debug("credentials already up to date")

pkg/aws/actuator/actuator.go
```diff
@@ -263,11 +264,82 @@ func (a *AWSActuator) getProviderSpec(cr *minterv1.CredentialsRequest) (*minterv
...
+			Message:   msg,
+		}
+	}
+
+	if cloudCredsSecret.Annotations[secretannotator.AnnotationKey] == secretannotator.PassthroughAnnotation {
+		logger.Debugf("provisioning with passthrough")
+		err := a.syncPassthrough(ctx, cr, cloudCredsSecret, logger)
+		if err != nil {
+			return err
+		}
```
dgoodwin: I assume because it's just a secret copy, we safely assume that a failure in passthrough mode can be safely retried and doesn't need a condition?

pkg/aws/actuator/actuator.go
```diff
@@ -794,15 +872,20 @@ func (a *AWSActuator) loadClusterInstallConfig(logger log.FieldLogger) (*install
 	return ic, nil
 }
 
-func (a *AWSActuator) syncAccessKeySecret(cr *minterv1.CredentialsRequest, accessKey *iam.AccessKey, existingSecret *corev1.Secret, userPolicy string, logger log.FieldLogger) error {
+func (a *AWSActuator) syncAccessKeySecret(cr *minterv1.CredentialsRequest, accessKey, secretAccessKey string, existingSecret *corev1.Secret, userPolicy string, logger log.FieldLogger) error {
```
dgoodwin: Probably worth a rename from accessKey to accessKeyID

pkg/aws/actuator/actuator.go
```diff
@@ -880,6 +967,40 @@ func (a *AWSActuator) getDesiredUserPolicy(entries []minterv1.StatementEntry, us
...
+		return nil, &actuatoriface.ActuatorError{
+			ErrReason: minterv1.CredentialsProvisionFailure,
+			Message:   fmt.Sprintf("cannot proceed without cloud cred secret annotation"),
+		}
+	}
+
+	return cloudCredSecret, nil
+}
+
+func properlyAnnotatedSecret(secret *corev1.Secret) bool {
```
dgoodwin: Would isSecretAnnotated look a little better

pkg/aws/actuator/actuator.go
```diff
@@ -134,9 +135,9 @@ func (a *AWSActuator) Exists(ctx context.Context, cr *minterv1.CredentialsReques
 
 }
 
-// NeedsUpdate will return whether the current credentials satisfy what's being requested
+// needsUpdate will return whether the current credentials satisfy what's being requested
 // in the CredentialsRequest
-func (a *AWSActuator) NeedsUpdate(ctx context.Context, cr *minterv1.CredentialsRequest) (bool, error) {
+func (a *AWSActuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequest) (bool, error) {
```
dgoodwin: What purpose does needsUpdate serve? It feels like we do a lot of the same style of checking both here and in sync. Would it not make sense for sync to just go piece by piece and check if that part is ok, if not fix it?

pkg/aws/actuator/actuator.go
```diff
@@ -263,11 +264,82 @@ func (a *AWSActuator) getProviderSpec(cr *minterv1.CredentialsRequest) (*minterv
...
+	if cloudCredsSecret.Annotations[secretannotator.AnnotationKey] == secretannotator.InsufficientAnnotation {
+		msg := "cloud credentials insufficient to satisfy credentials request"
+		logger.Error(msg)
+		return &actuatoriface.ActuatorError{
+			ErrReason: minterv1.InsufficientCloudCredentials,
+			Message:   msg,
+		}
+	}
+
+	if cloudCredsSecret.Annotations[secretannotator.AnnotationKey] == secretannotator.PassthroughAnnotation {
```
dgoodwin: With these big forks in logic it might be worth doing a credPassthroughStrategy and credMintStrategy, check the annotation early in the actuator, create the right implementation, and then move any specific sync logic and needsupdate logic to the strategies.

pkg/aws/actuator/actuator.go
```diff
@@ -880,6 +967,40 @@ func (a *AWSActuator) getDesiredUserPolicy(entries []minterv1.StatementEntry, us
...
 
+func (a *AWSActuator) getCloudCredentialsSecret(ctx context.Context, logger log.FieldLogger) (*corev1.Secret, error) {
+	cloudCredSecret := &corev1.Secret{}
+	if err := a.Client.Get(ctx, types.NamespacedName{Name: rootAWSCredsSecret, Namespace: rootAWSCredsSecretNamespace}, cloudCredSecret); err != nil {
+		msg := "unable to fetch cloud cred secret"
+		logger.WithError(err).Error(msg)
+		return nil, &actuatoriface.ActuatorError{
+			ErrReason: minterv1.CredentialsProvisionFailure,
+			Message:   fmt.Sprintf("%v: %v", msg, err),
+		}
```
dgoodwin: Ah do we now get a condition if the root creds don't exist? I think I had a todo for this, so that's great if so.

pkg/aws/actuator/actuator.go
```diff
@@ -880,6 +967,40 @@ func (a *AWSActuator) getDesiredUserPolicy(entries []minterv1.StatementEntry, us
 	return string(b), nil
 }
 
+func (a *AWSActuator) getCloudCredentialsSecret(ctx context.Context, logger log.FieldLogger) (*corev1.Secret, error) {
+	cloudCredSecret := &corev1.Secret{}
+	if err := a.Client.Get(ctx, types.NamespacedName{Name: rootAWSCredsSecret, Namespace: rootAWSCredsSecretNamespace}, cloudCredSecret); err != nil {
+		msg := "unable to fetch cloud cred secret"
```
dgoodwin: + root + in there somewhere maybe. Or admin.

pkg/controller/credentialsrequest/actuator/actuator.go
```diff
@@ -45,10 +41,6 @@ func (a *DummyActuator) Exists(ctx context.Context, cr *minterv1.CredentialsRequ
 	return true, nil
 }
 
-func (a *DummyActuator) NeedsUpdate(ctx context.Context, cr *minterv1.CredentialsRequest) (bool, error) {
-	return false, nil
-}
-
```
dgoodwin: Glad to see these go off the actuator interface.

pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -443,60 +416,20 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
...
-				},
-			},
-			Data: map[string][]byte{
-				secretannotator.AwsAccessKeyName:       cloudCredSecret.Data[secretannotator.AwsAccessKeyName],
-				secretannotator.AwsSecretAccessKeyName: cloudCredSecret.Data[secretannotator.AwsSecretAccessKeyName],
-			},
-		}
+	// first unset any previous conditions
+	setInsufficientCredsCondition(cr, false)
+	setFailedToProvisionCredentialsRequest(cr, false, nil)
```
dgoodwin: Nervous about this section, if we were to get more fine grained with credentials conditions we might have some trouble maintaining this. Probably ok for now though.

pkg/aws/actuator/actuator.go
joelddiaz: the syncPassthrough() call will create/return the ActuatorError, so we will end up setting the CredentialsProvisionFailure condition

pkg/aws/actuator/actuator.go
joelddiaz: yep. any failure to provision should give us at least the CredentialsProvisionFailure condition now

pkg/controller/credentialsrequest/credentialsrequest_controller.go
joelddiaz: this was a weird one. the situation i had in mind when i structured it this way was as follows.

1) attempt to actuator.sync() fails b/c Reason/ConditionA
2) attempt to actuator.sync() fails b/c Reason/ConditionB
3) attempt to actuatior.sync() works

At the end of 3, which conditions should be set/cleared?

The approach I took was compiling a list of the conditions that actuator.sync() can return, and before setting any condition related to an error from the actuator.sync() call, just clear all actuator.sync()-possible conditions then set the one condition that is actually in the error we're presently handling.

So we're basically saying that any conditions that can be set as a result of actuator.sync() are exclusively setable only as a result of an actuator.sync() call (ie. credentials_request controller shouldn't set/unset these unless it's due to the results of an actuator.sync() call).

pkg/aws/actuator/actuator.go
joelddiaz: as discussed offline, needsUpdate() can determine whether we need an update in the case where the root creds are missing.

but if we need an update, we need to know whether to mint or passthrough which requires the existence of the root creds.

so needsUpdate() lets us succeed in the case where root creds are missing, but there is no need to update.

dgoodwin:
LGTM lets squash.

pkg/aws/actuator/actuator.go
dgoodwin: Got it, thanks!

PR #33, opened 2019-02-19
move to static list of passthrough creds

joelddiaz:
in preparation for moving the credentialsrequests manifests out of this repo (and into each coresponding repo's own location), remove the CredentialsRequest manifests and (for now) keep a static list in pkg/controller/utils/utils.go (to continue to allow annotating mint vs passthrough)

remove go-bindata dependency


pkg/controller/utils/utils.go
```diff
@@ -32,6 +31,66 @@ var (
 		"iam:SimulatePrincipalPolicy", // needed so we can verify the above list of course
 	}
 
+	CredPassthroughActions = []string{
```
dgoodwin: credPassthroughActions I think might be correct syntax. CredMintingActions above looks off too.

dgoodwin:
Looks perfect otherwise.

PR #35, opened 2019-02-27
Remove limits from pods

fabianofranz:
Removes CPU limits from pods, soon to be checked by the test in https://github.com/openshift/origin/pull/22095.

Re https://jira.coreos.com/browse/CO-262

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -53,10 +53,9 @@ spec:
         name: manager
         resources:
           limits:
-            cpu: 100m
             memory: 500Mi
           requests:
-            cpu: 100m
+            cpu: 10m
             memory: 150Mi
         ports:
         - containerPort: 9876
--- a/manifests/0000_30_cloud-credential-operator_01_deployment.yaml
+++ b/manifests/0000_30_cloud-credential-operator_01_deployment.yaml
@@ -147,10 +147,9 @@ spec:
           protocol: TCP
         resources:
           limits:
-            cpu: 100m
             memory: 500Mi
           requests:
-            cpu: 100m
+            cpu: 10m
             memory: 150Mi
       nodeSelector:
         node-role.kubernetes.io/master: ""


config/manager/deployment.yaml
```diff
@@ -52,9 +52,6 @@ spec:
         imagePullPolicy: Always
         name: manager
         resources:
-          limits:
-            cpu: 100m
-            memory: 500Mi
           requests:
             cpu: 100m
```
dgoodwin: Could you flip the requests to 10m as well, sounded like Derek is going to soon be asking for that as well so we may as well get ready.

config/manager/deployment.yaml
```diff
@@ -52,9 +52,6 @@ spec:
         imagePullPolicy: Always
         name: manager
         resources:
-          limits:
-            cpu: 100m
-            memory: 500Mi
```
dgoodwin: Should the memory limit go as well? His request seemed specific to CPU but I don't know if you can even have a memory limit and no cpu specified.

fabianofranz: The test checks for the `memory` attribute the same way it does the `cpu`, so it seems it needs to go as well.

config/manager/deployment.yaml
fabianofranz: What the test does as of now is to check it it's present and if not, it says "recommend *min* value 10m". But I had the same questions. @derekwaynecarr can you check the two comments in this PR and confirm?

config/manager/deployment.yaml
dgoodwin: Very good catch, thanks.

derekwaynecarr: i am less concerned about a memory limit than cpu limits.  The test only fails on cpu limits.

config/manager/deployment.yaml
derekwaynecarr: a follow on test will validate that we schedule no more than 2 cores of cpu requests on master nodes.  Please set this to 10m.  The cpu time is relative to other components, and 10m is fine.

derekwaynecarr: basically, we will give the api server and etcd equal cpu time by making their request equal ~150m.  we will give the controller manager and scheduler less relative time ~100m.  any operator is basically a single controller in controller manager and giving it 1/10 what we will give controller manager is fine.  The key need is that we give operators matching cpu requests so they have a fair share of cpu time relative to each other, but less than etcd/APIServer

fabianofranz: Nice, thanks for clarifying @derekwaynecarr. PR updated.

fabianofranz:
PR updated to reflect Derek's comments, @dgoodwin PTAL.

PR #36, opened 2019-02-28
pkg/controller/utils: Fix "reutrn" -> "return" comment typo

wking:
This slipped in with b46a3f86 (#8).


dgoodwin:
How the heck did you find this. :)

/lgtm

wking:
I was reading code while stumped by openshift/installer#1322 ;).

PR #37, opened 2019-03-01
Remove Machine API manifest

spangenberg:
This file lives now under install/0000_30_machine-api-operator_00_credentials-request.yaml in machine-api-operator.

FYI: https://github.com/openshift/machine-api-operator/pull/241


enxebre:
I think this file can be removed from here now that we have https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_00_credentials-request.yaml

dgoodwin:
Yeah that's correct, we should fix in your repo, and delete here.

spangenberg:
@dgoodwin right now it looks like this file gets preferred over the other one during installation. If this one gets deleted will the other file in the MAO get picked up?

dgoodwin:
It likely applies them both, so one goes in first and then the next. It should be ok to remove this one.

spangenberg:
Okay cool, I'll update the PR.

PR #38, opened 2019-03-04
Use PullPolicy IfNotPresent instead of Always.

dgoodwin:
Prevents problems caused by registry outages.


dgoodwin:
Also fixes the problems with kustomize 2.0 and this layout, and removes an unnecessary image patch.

PR #39, opened 2019-03-04
Migrate CredentialsRequest from v1beta1 to v1.

dgoodwin:
Adds a no-op conversion from v1beta1 to v1, they are identical, once the
CRD (which references multiple versions, v1 marked for storage) deploys,
existing CRDs are now returned as v1.

This just leaves the problem of the embedded provider spec/status, which
remains v1beta1. To handle this I added a migration to the actuator
interface, which attempts to load v1beta1 fields. If either succeeds, we
immediately convert it and save it back, and requeue the
CredentialsRequest.


dgoodwin:
Tests aren't done yet but this is the approach that seems to be working.

dgoodwin:
Confirmed with sttts that the no-op upgrade part is correct, we just have to make sure to update every CR before we drop v1beta1 support completely, otherwise they're still internally stored with the old v1beta1 version.

dgoodwin:
@csrwng could you take a look at this one when time allows.

pkg/controller/utils/utils.go
```diff
@@ -5,7 +5,8 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
-	ccv1beta1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1beta1"
+	minterv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+	//minterv1beta1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1beta1"
```
csrwng: remove

dgoodwin:
Oops, updated!

PR #42, opened 2019-03-13
Clusteroperator version status

joelddiaz:
vendor in new openshift/api to get updated ClusterOperator object type

as part of updating the ClusterOperator object status, also fill in the versions[] field based on the RELEASE_VERSION environment variable (update the deployment for the operator so that CVO can in-place update the RELEASE_VERSION env var)


pkg/controller/credentialsrequest/status.go
```diff
@@ -94,6 +99,17 @@ func (r *ReconcileCredentialsRequest) getOperatorState() (*corev1.Namespace, []m
...
 
+func computeClusterOperatorVersions() []configv1.OperandVersion {
+	currentVersion := os.Getenv("RELEASE_VERSION")
+	versions := []configv1.OperandVersion{
+		{
+			Name:    "operator",
+			Version: currentVersion,
+		},
+	}
+	return versions
```
dgoodwin: I assume we need to set this in our deployment per the example from Jessica's email? https://github.com/openshift/cluster-image-registry-operator/pull/225 

Then CVO replaces any occurrence it sees of 0.0.1-snapshot as a magic string, and as soon as the new deployment is up, we're reporting correct version. (hopefully I have that right)

Everything looks right here just missing the env var as far as I can see.

joelddiaz: yes, i had mistakenly directly edited manifests/0000_30_cloud-credentials-operator_01_deployment.yaml with these changes, and the 'make manifests' wiped it.

updated commit with changes to config/manager/deployment.yaml instead.

dgoodwin:
```
        {
            "apiVersion": "config.openshift.io/v1",
            "kind": "ClusterOperator",
            "metadata": {
                "creationTimestamp": "2019-03-14T13:41:37Z",
                "generation": 1,
                "name": "openshift-cloud-credential-operator",
                "resourceVersion": "7215",
                "selfLink": "/apis/config.openshift.io/v1/clusteroperators/openshift-cloud-credential-operator",
                "uid": "e38c50b9-465e-11e9-9daa-128d077c1700"
            },
            "spec": {},
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2019-03-14T13:45:37Z",
                        "message": "No credentials requests reporting errors.",
                        "reason": "NoCredentialsFailing",
                        "status": "False",
                        "type": "Failing"
                    },
                    {
                        "lastTransitionTime": "2019-03-14T13:45:37Z",
                        "message": "4 of 4 credentials requests provisioned and reconciled.",
                        "reason": "ReconcilingComplete",
                        "status": "False",
                        "type": "Progressing"
                    },
                    {
                        "lastTransitionTime": "2019-03-14T13:41:36Z",
                        "status": "True",
                        "type": "Available"
                    }
                ],
                "extension": null,
                "versions": [
                    {
                        "name": "operator",
                        "version": "0.0.1-2019-03-14-132845"
                    }
                ]
            }
        },
```

I'd say you're good. Drop WIP if you agree.

/lgtm

Thanks!

PR #43, opened 2019-03-18
detect using root user

joelddiaz:
as reported in https://github.com/openshift/installer/issues/1341 , the root aws account is not a valid user to query against with an iam:SimulatePrincipalPolicy call.

detect when the creds are the root creds (ie UserId == account number), print a warning and allow it through.

### DIFF
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -9,6 +9,7 @@ import (
 	ccaws "github.com/openshift/cloud-credential-operator/pkg/aws"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/service/iam"
 
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


pkg/controller/utils/utils.go
```diff
@@ -9,12 +9,17 @@ import (
...
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/service/iam"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 )
 
+const (
+	rootUserDetected = "Detected using root account credentials. Using the AWS account root user is not recommended: https://docs.aws.amazon.com/general/latest/gr/managing-aws-access-keys.html"
```
wking: nit: I don't think we need "Detected using root account credentials.", because "Using the AWS account root user is not recommended" seems to clearly imply that we think you are (to me anyway).  But I don't feel particularly strongly about it, and I'm not a maintainer over here anyway ;).

pkg/controller/utils/utils.go
```diff
@@ -107,12 +112,41 @@ func CheckCloudCredCreation(awsClient ccaws.Client, logger log.FieldLogger) (boo
 	return CheckPermissionsAgainstActions(awsClient, credMintingActions, logger)
 }
 
+// getClientDetails will return the *iam.User associated with the passed in credentials,
```
wking: nit: "passed in" -> "provided"?

pkg/controller/utils/utils.go
```diff
@@ -107,12 +112,41 @@ func CheckCloudCredCreation(awsClient ccaws.Client, logger log.FieldLogger) (boo
...
+// a boolean indicating whether the user is the 'root' account, and any error encountered
+// while trying to gather the info
```
wking: nit: [godoc comments should have trailing punctuation][1].

[1]: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

pkg/controller/utils/utils.go
```diff
@@ -107,12 +112,41 @@ func CheckCloudCredCreation(awsClient ccaws.Client, logger log.FieldLogger) (boo
...
+func getClientDetails(awsClient ccaws.Client) (*iam.User, bool, error) {
+	rootUser := false
+
+	user, err := awsClient.GetUser(nil)
+	if err != nil {
+		return nil, rootUser, fmt.Errorf("error querying username: %v", err)
+	}
+
+	// Detect whether the AWS account's root user is being used (unsupported)
```
wking: why "unsupported"?  AWS recommends (wisely) that you not do this, but the cluster should come up fine when launched by the account's root.  And if there are voids-your-warranty type issues vs. Red Hat support for such clusters, then we probably want to surface that at a higher-level than a library comment ;).

joelddiaz: that was earlier today when i thought we were going to error out in the face of the root user

pkg/controller/utils/utils.go
dgoodwin: +1, provided or given.

pkg/controller/utils/utils.go
dgoodwin: +1

pkg/controller/utils/utils.go
```diff
@@ -107,12 +112,40 @@ func CheckCloudCredCreation(awsClient ccaws.Client, logger log.FieldLogger) (boo
...
 func CheckPermissionsUsingQueryClient(queryClient, targetClient ccaws.Client, statementEntries []minterv1.StatementEntry, logger log.FieldLogger) (bool, error) {
-	targetUsername, err := targetClient.GetUser(nil)
+	targetUser, isRoot, err := getClientDetails(targetClient)
 	if err != nil {
-		return false, fmt.Errorf("error querying current username: %v", err)
+		return false, fmt.Errorf("error gathering AWS credentials details: %v", err)
+	}
+	if isRoot {
+		// warn about using the root creds, and just return that the creds are good enough
+		logger.Warn(rootUserDetected)
```
dgoodwin: We should probably keep the log message with the log statement. I assume at one point you planned to re-use the message somewhere, maybe a condition? Did that get scrapped for some reason?

joelddiaz: yeah, there was a version where i was checking both the queryClient and the targetClient for 'root', but i realized it wasn't necessary for the queryClient. i'll update it

dgoodwin:
/lgtm

Need to drop WIP.

PR #50, opened 2019-04-04
Bug: 1695205 Reset progressing transition time after upgrades.

dgoodwin:
CCO does not have any components to upgrade, so the first time we go to
update status and see that our running version has changed, update the
progressing last transition time to indicate something happened.

### DIFF
--- a/pkg/controller/credentialsrequest/status_test.go
+++ b/pkg/controller/credentialsrequest/status_test.go
@@ -20,12 +20,14 @@ import (
 	"context"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/controller/credentialsrequest/status_test.go
```diff
@@ -182,20 +205,58 @@ func TestClusterOperatorVersion(t *testing.T) {
...
+			assert.True(t, foundVersion, "didn't find an entry named 'operator' in the version list")
+
+			progCond := findClusterOperatorCondition(clusterop.Status.Conditions,
+				configv1.OperatorProgressing)
+			assert.NotNil(t, progCond)
+			if test.expectProgressingTransition {
+				assert.True(t, progCond.LastTransitionTime.Time.After(
+					test.currentProgressingLastTransition.Time))
+			} else {
+				assert.Equal(t, progCond.LastTransitionTime.Time.Format(time.UnixDate),
```
joelddiaz: assert.Equal() wants the expected value to be the second param, and the tested value to be the third param.

pkg/controller/credentialsrequest/status_test.go
```diff
@@ -182,20 +205,58 @@ func TestClusterOperatorVersion(t *testing.T) {
...
 			if err := fakeClient.Get(context.TODO(), client.ObjectKey{Name: cloudCredClusterOperator}, clusterop); err != nil {
 				t.Errorf("error fetching clusteroperator object: %v", err)
-			} else {
-				foundVersion := false
-				for _, version := range clusterop.Status.Versions {
-					if version.Name == "operator" {
-						foundVersion = true
-						assert.Equal(t, test.releaseVersionEnv, version.Version)
-					}
+				return
```
joelddiaz: nit. you can probably just assert.NoError() on that fakeClient.Get() call above

csrwng:
sorry @joelddiaz :)

csrwng:
cancel hold when you're happy with it

dgoodwin:
updated, thanks!

PR #51, opened 2019-04-05
Begin making use of Infrastructure config.

dgoodwin:
Vendors latest config API.
Uses Infrastructure.Status.Platform for AWS check instead of deprecated config map.
Uses InfrastructureName for AWS user agent to help track down hotlooping clusters if it resurfaces.

Deprecated configmap still in use in some areas, will work on this over time. (user tags might need to get axed)

Also will do a followup PR to use the InfrastructureName to properly tag the user creds, this is not included in this one.


dgoodwin:
Forgot the RBAC, crashed on startup which meant no nodes which meant no pod logs, tricky to diagnose.

dgoodwin:
@joelddiaz ready for another look, just added a couple rbac lines.

PR #54, opened 2019-04-09
Fix the hotloop on certain types of errors.

dgoodwin:


pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -421,22 +421,15 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
 		cr.Status.LastSyncGeneration = origCR.Generation
 	}
 
+	// TODO: if syncErr is not nil, and this updateStatus runs, we hotloop.
+	// If we remove the updateStatus call, we do not hotloop.
 	err = r.updateStatus(origCR, cr, logger)
```
dgoodwin: @joelddiaz @csrwng weirdest thing i've seen yet... ^^^

dgoodwin: Answer was right infront of my face, Cesar pointed out on slack, it's because the update status is seeing a change, and then re-syncing the cluster every error, aka our likely hotloop that was eating all AWS requests.

pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -427,29 +427,23 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
...
-
-	return reconcile.Result{}, nil
+	return reconcile.Result{}, syncErr
 }
 
 func (r *ReconcileCredentialsRequest) updateActuatorConditions(cr *minterv1.CredentialsRequest, reason minterv1.CredentialsRequestConditionType, conditionError error) {
 
-	// first unset any previous conditions
-	setInsufficientCredsCondition(cr, false)
-	setFailedToProvisionCredentialsRequest(cr, false, nil)
```
dgoodwin: Fixed now, problem was these two lines causing condition timestamps to flap even though the same error was occurring over and over.

PR #55, opened 2019-04-13
Make AWSProviderCodec generic for other clouds

juan-lee:
AWSProviderCodec is currently only usable in the AWS actuator even
though other cloud providers could make use of the provided
functionality.

This change replaces AWSProviderCodec with ProviderCodec that can be
used with different cloud provider spec and status implementations.
ProviderCodec uses a similar approach to the controller-runtime client
interfaces by writing to the `out runtime.Object` instead of returning a
separate value.

The upcoming PR for the azure actuator will depend on ProviderCodec.


dgoodwin:
/lgtm

Thanks @juan-lee 

/test e2e-aws

PR #57, opened 2019-04-17
Azure Provider Passthrough Actuator

juan-lee:
This change brings passthrough behavior to the azure provider. In other
words, root credentials which have contributor or greater permissions to
the openshift cluster's resource group on azure are reused for each
credentials request. The installer ensures that the root credential is
at least contributor as part of initial provisioning.

The next change in this series will introduce the credential minting
actuator.


pkg/apis/cloudcredential/v1/azure_types.go
```diff
@@ -0,0 +1,55 @@
...
+	Scope string `json:"scope"`
+}
+
+// AzureProviderStatus contains the status of the credentials request in Azure.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type AzureProviderStatus struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Name is the service principal created in Azure for these credentials.
+	Name string `json:"name"`
```
dgoodwin: This looked slightly ambiguous when I saw it used in code, would it be worth going to ServicePrincipal or ServicePrincipalName?

juan-lee: Looking at this again, I completely agree. I also changed ClientID to AppID to match the names in our azure sdk/api.

pkg/azure/client.go
```diff
@@ -0,0 +1,75 @@
...
+
+	minterv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+	"github.com/openshift/cloud-credential-operator/pkg/controller/credentialsrequest/actuator"
+	"github.com/openshift/cloud-credential-operator/pkg/controller/secretannotator"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	RootSecretNamespace = "openshift-config"
+	RootSecretName      = "azure-creds"
```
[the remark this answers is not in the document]
juan-lee: @joelddiaz tipped me off. 😀

pkg/controller/secretannotator/secretannotator_controller.go
```diff
@@ -63,6 +63,9 @@ const (
 
 	AwsAccessKeyName       = "aws_access_key_id"
 	AwsSecretAccessKeyName = "aws_secret_access_key"
+
+	AzureClientID     = "azure_client_id"
+	AzureClientSecret = "aws_secret_access_key"
```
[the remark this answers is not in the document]
juan-lee: good catch!

pkg/azure/actuator_test.go
```diff
@@ -0,0 +1,53 @@
...
+		errRegexp string
+	}{
+		{validRootSecret, ""},
+		{rootSecretBadAnnotation, "invalid mode"},
+		{rootSecretNoAnnotation, "cannot proceed without cloud cred secret annotation.*"},
+	}
+
+	for _, tt := range tests {
+		t.Run("", func(t *testing.T) {
+			f := newFakeClient(t)
```
joelddiaz: i see the custom fake client you made, and i'm wondering why you didn't go with the fake client you get from "sigs.k8s.io/controller-runtime/pkg/client/fake"

pkg/controller/controller.go
```diff
@@ -52,22 +53,29 @@ func AddToManager(m manager.Manager) error {
...
 				return err
 			}
-		} else {
+		case configv1.AzurePlatformType:
+			log.Info("initializing Azure actuator")
+			a, err = azure.NewActuator(m.GetClient())
+			if err != nil {
+				return err
+			}
+		default:
```
joelddiaz: maybe update the log message below to something like "initializing no-op actuator (not a supported platform)"

pkg/azure/actuator_test.go
juan-lee: didn't know about that one. Thank you for pointing it out. I'll certainly use it again in the future.

pkg/azure/actuator_test.go
```diff
@@ -0,0 +1,53 @@
...
+		in        corev1.Secret
+		errRegexp string
+	}{
+		{validRootSecret, ""},
+		{rootSecretBadAnnotation, "invalid mode"},
+		{rootSecretNoAnnotation, "cannot proceed without cloud cred secret annotation.*"},
+	}
+
+	for _, tt := range tests {
+		t.Run("", func(t *testing.T) {
```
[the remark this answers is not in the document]
juan-lee: no particular reason, just an oversight. Thanks for pointing it out.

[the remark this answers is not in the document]
joelddiaz: i see you added the name field, but the t.Run() still has the empty string as the first parameter instead of tt.name

joelddiaz:
@juan-lee i think just adding that tt.name to the t.Run() call and we're good to go

juan-lee:
> @juan-lee i think just adding that tt.name to the t.Run() call and we're good to go

rookie mistake, added the code, but didn't use it :P thanks!

PR #58, opened 2019-04-19
provide pod logs via API on failures

deads2k:
When pods fail, you normally get logs, but if other bits of the cluster fail they aren't always available.  This provides a snippet on crashes.


deads2k:
network is not ready: [runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni config uninitialized]

/retest

smarterclayton:
/approve
/lgtm

all operators should report terminationMessagePath fallback

PR #63, opened 2019-05-03
Populate all of the azure credential secret

juan-lee:
The root azure credential has more information about the identity
context that is needed for the ingress operator.

### DIFF
--- a/pkg/controller/secretannotator/secretannotator_controller.go
+++ b/pkg/controller/secretannotator/secretannotator_controller.go
@@ -64,8 +64,13 @@ const (
 	AwsAccessKeyName       = "aws_access_key_id"
 	AwsSecretAccessKeyName = "aws_secret_access_key"
 
-	AzureClientID     = "azure_client_id"
-	AzureClientSecret = "azure_client_secret"
+	AzureClientID       = "azure_client_id"
+	AzureClientSecret   = "azure_client_secret"
+	AzureRegion         = "azure_region"
+	AzureResourceGroup  = "azure_resourcegroup"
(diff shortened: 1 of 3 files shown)


pkg/controller/secretannotator/secretannotator_controller.go
```diff
@@ -64,8 +64,13 @@ const (
...
 
-	AzureClientID     = "azure_client_id"
-	AzureClientSecret = "azure_client_secret"
+	AzureClientID       = "azure_client_id"
+	AzureClientSecret   = "azure_client_secret"
+	AzureRegion         = "azure_region"
+	AzureResourceGroup  = "azure_resourcegroup"
+	AzureResourcePrefix = "azure_resource_prefix"
+	AzureSubscriptionID = "azure_subscription_id"
+	AzureTenantID       = "azure_tenant_id"
```
dgoodwin: Only thing I'm curious about is if these belong more closely to the credential, than to the cluster you are creating.

Region for example in AWS is a property on the ClusterDeployment AWS config, rather than the credential, and the credential could be used to provision clusters in many regions. 

Do all these properties align only to one set of credentials?

juan-lee: The credential can be used for multiple clusters in different resource groups. However, in the context of the particular cluster in which this secret is deployed, the subscription and other properties are part of the root credential role's scope.

dgoodwin: Whoops I am confusing the other project we think about all day, forgot we're dealing with a component in-cluster only here.

PR #64, opened 2019-05-03
Fix tolerations

ravisantoshgudimetla:
In 4.1, we have taint-based evictions enabled.  This changes logic in the node lifecycle controller in the kube-controller-manager significantly.

Historically, the node lifecycle controller would directly evict pods from node that had a `Ready` condition of `False` or `Unknown` after a pod eviction timeout set by the `--pod-eviction-timeout flag` on the kube-controller-manager.  This setting applied to all pods cluster-wide.  The default was `5m`.

With taint-based evictions, all the node lifecycle controller does is taint the node with `node.kubernetes.io/unreachable` and/or `node.kubernetes.io/not-ready` taint with `NoExecute` effect.  This would normally result in the immediate eviction of all pods that don't tolerate those taints, breaking the old behavior with pod eviction timeouts.  Enter `DefaultTolerationSeconds` mutating admission plugin.

https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/plugin/pkg/admission/defaulttolerationseconds/admission.go

This plugin will add a default `NoExecute` toleration for `node.kubernetes.io/unreachable` and/or `node.kubernetes.io/not-ready` taints with a `tolerationSeconds` of 5m (300s) as long as no such toleration is already specified in the pod spec.  This restores the old pod eviction tiemout behavior.

One of the intended effects of this change is the make the pod eviction timeout a pod-level property.  Different applications require different timeouts depending on their design and having it controlled at a cluster level before was not optimal.

The `DefaultTolerationSeconds` plugin has a flag that allows adjusting the defaults for `tolerationSeconds`
https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/plugin/pkg/admission/defaulttolerationseconds/admission.go#L34-L40

We might expose this tunable for user control in the future and do not want the cluster control plane components to be subject to it.

Thus, this PR explictly defines a `NoExecute` toleration for `node.kubernetes.io/unreachable` and/or `node.kubernetes.io/not-ready` taints with a `tolerationSeconds` generically appropriate for cluster components.

Once these changes are in across all components, this e2e will enforce it in the future
https://github.com/openshift/origin/pull/22752 

/cc @sjenning @smarterclayton @derekwaynecarr

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -48,7 +48,17 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
       tolerations:
-      - operator: Exists
+      - key: "node-role.kubernetes.io/master"
+        operator: "Exists"
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/unreachable"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
       containers:
       - env:
         - name: RELEASE_VERSION
--- a/manifests/01_deployment.yaml
+++ b/manifests/01_deployment.yaml
@@ -170,4 +170,14 @@ spec:
       priorityClassName: system-cluster-critical
       terminationGracePeriodSeconds: 10
       tolerations:
-      - operator: Exists
+      - key: "node-role.kubernetes.io/master"
+        operator: "Exists"
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/unreachable"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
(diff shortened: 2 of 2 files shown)


ravisantoshgudimetla:
`May 06 23:35:30.315 W ns/openshift-machine-config-operator pod/etcd-quorum-guard-86bcc896bd-z52fx 0/6 nodes are available: 2 node(s) didn't match node selector, 2 node(s) didn't match pod affinity/anti-affinity, 2 node(s) didn't satisfy existing pods anti-affinity rules, 2 node(s) were unschedulable. (49 times)
May 06 23:36:39.391 E clusterversion/version changed Failing to True: ClusterOperatorDegraded: Cluster operator machine-config is reporting a failure: Failed to resync 0.0.1-2019-05-06-214309 because: timed out waiting for the condition during syncRequiredMachineConfigPools: error pool master is not ready, retrying. Status: (pool degraded: true total: 3, ready 1, updated: 1, unavailable: 1)`

/cc  @RobertKrawitz

PR #67, opened 2019-05-08
Azure secret annotator

serbrech:
The current azure passthrough setup forgot to implement the `secretannotator` for azure. 
As a result, no credentials where ever delivered in response to a credential request.

To resolve that : 

- added the azure implementation for the secret-annotator
  - moved aws in its own package `pkg/controller/secretannotator/aws`
  - added `pkg/controller/secretannotator/azure`
  - made `pkg/controller/secretannotator/controller.go` decide on which implementation to run depending on the platform

to make this refactoring possible, some parts had to move around :

- added `pkg/controller/platform` that retrieves the `platformtype` from a `Manager`
- added a `pkg/controller/secretannotator/constants` package to make it possible to share variables across the different packages without circular dependencies
- moved aws secretannotator tests under aws, and azure ones under azure

The code structure has changed quite a bit, but apart from the switch case in secretannotator, the aws logic is untouched, and there is a single place where the code path splits, which should reduce risks of impacting the existing.

@juan-lee @abhinavdahiya

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -1142,3 +1142,21 @@ func (a *AWSActuator) loadClusterUUID(logger log.FieldLogger) (configv1.ClusterI
 	logger.WithField("clusterID", clusterVer.Spec.ClusterID).Debug("found cluster ID")
 	return clusterVer.Spec.ClusterID, nil
 }
+
+func isAWSCredentials(providerSpec *runtime.RawExtension) (bool, error) {
+	codec, err := minterv1.NewCodec()
+	if err != nil {
+		return false, err
+	}
+	unknown := runtime.Unknown{}
+	err = codec.DecodeProviderSpec(providerSpec, &unknown)
+	if err != nil {
+		return false, err
+	}
+	isAWS := unknown.Kind == reflect.TypeOf(minterv1.AWSProviderSpec{}).Name()
+	if !isAWS {
+		log.WithField("kind", unknown.Kind).
+			Info("actuator handles only azure credentials")
+	}
+	return isAWS, nil
+}
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -120,6 +120,9 @@ func (a *AWSActuator) Exists(ctx context.Context, cr *minterv1.CredentialsReques
 	logger := a.getLogger(cr)
 	logger.Debug("running Exists")
 	var err error
+	if isAWS, err := isAWSCredentials(cr.Spec.ProviderSpec); !isAWS {
+		return false, err
+	}
 	awsStatus, err := DecodeProviderStatus(a.Codec, cr)
 	if err != nil {
 		return false, err
@@ -252,7 +255,9 @@ func (a *AWSActuator) Update(ctx context.Context, cr *minterv1.CredentialsReques
 }
 
 func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) error {
-
+	if isAWS, err := isAWSCredentials(cr.Spec.ProviderSpec); !isAWS {
+		return err
+	}
 	logger := a.getLogger(cr)
 	logger.Debug("running sync")
 
@@ -572,6 +577,9 @@ func (a *AWSActuator) updateProviderStatus(ctx context.Context, logger log.Field
 
 // Delete the credentials. If no error is returned, it is assumed that all dependent resources have been cleaned up.
 func (a *AWSActuator) Delete(ctx context.Context, cr *minterv1.CredentialsRequest) error {
+	if isAWS, err := isAWSCredentials(cr.Spec.ProviderSpec); !isAWS {
+		return err
+	}
 	logger := a.getLogger(cr)
 	logger.Debug("running Delete")
 	var err error
@@ -909,7 +917,7 @@ func (a *AWSActuator) getCloudCredentialsSecret(ctx context.Context, logger log.
 	}
 
 	if !isSecretAnnotated(cloudCredSecret) {
-		logger.WithField("secret", fmt.Sprintf("%s/%s", annotatorconst.CloudCredSecretNamespace, annotatorconst.CloudCredSecretName)).Error("cloud cred secret not yet annotated")
+		logger.WithField("secret", fmt.Sprintf("%s/%s", rootAWSCredsSecretNamespace, rootAWSCredsSecret)).Error("cloud cred secret not yet annotated")
 		return nil, &actuatoriface.ActuatorError{
 			ErrReason: minterv1.CredentialsProvisionFailure,
 			Message:   fmt.Sprintf("cannot proceed without cloud cred secret annotation"),
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -28,7 +28,8 @@ import (
 	ccaws "github.com/openshift/cloud-credential-operator/pkg/aws"
 	minteraws "github.com/openshift/cloud-credential-operator/pkg/aws"
 	actuatoriface "github.com/openshift/cloud-credential-operator/pkg/controller/credentialsrequest/actuator"
-	"github.com/openshift/cloud-credential-operator/pkg/controller/secretannotator"
+	awsannotator "github.com/openshift/cloud-credential-operator/pkg/controller/secretannotator/aws"
+	annotatorconst "github.com/openshift/cloud-credential-operator/pkg/controller/secretannotator/constants"
 	"github.com/openshift/cloud-credential-operator/pkg/controller/utils"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -281,7 +282,7 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
 		return err
 	}
 
-	if cloudCredsSecret.Annotations[secretannotator.AnnotationKey] == secretannotator.InsufficientAnnotation {
+	if cloudCredsSecret.Annotations[annotatorconst.AnnotationKey] == annotatorconst.InsufficientAnnotation {
 		msg := "cloud credentials insufficient to satisfy credentials request"
 		logger.Error(msg)
 		return &actuatoriface.ActuatorError{
@@ -290,13 +291,13 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
 		}
 	}
 
-	if cloudCredsSecret.Annotations[secretannotator.AnnotationKey] == secretannotator.PassthroughAnnotation {
+	if cloudCredsSecret.Annotations[annotatorconst.AnnotationKey] == annotatorconst.PassthroughAnnotation {
 		logger.Debugf("provisioning with passthrough")
 		err := a.syncPassthrough(ctx, cr, cloudCredsSecret, logger)
 		if err != nil {
 			return err
 		}
-	} else if cloudCredsSecret.Annotations[secretannotator.AnnotationKey] == secretannotator.MintAnnotation {
+	} else if cloudCredsSecret.Annotations[annotatorconst.AnnotationKey] == annotatorconst.MintAnnotation {
 		logger.Debugf("provisioning with cred minting")
 		err := a.syncMint(ctx, cr, infraName, logger)
 		if err != nil {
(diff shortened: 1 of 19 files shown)


pkg/controller/utils/platform.go
```diff
@@ -0,0 +1,43 @@
+package utils
```
abhinavdahiya: Util packages have tendency to hide wrong package structuring and I think we should avoid it .

`pkg/controller/platform` seems like a better fit for this package. Then you could have platform.Type func part of the api.

serbrech: yes, that's WIP and an intermediate step to refactor the existing unflexible AWS-only implementation

pkg/controller/secretannotator/azure/reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	controllerName      = "secretannotator"
```
dgoodwin: Should this use the const you made public in pkg/controller/secretannotator/ ?

pkg/controller/secretannotator/secretannotator_controller.go
```diff
@@ -60,156 +50,22 @@ const (
...
-		r.updateSecretAnnotations(secret, InsufficientAnnotation)
-		return fmt.Errorf("failed checking create cloud creds: %v", err)
-	}
-
-	if cloudCheckResult {
-		r.logger.Info("Verified cloud creds can be used for minting new creds")
-		return r.updateSecretAnnotations(secret, MintAnnotation)
+	switch platformType {
+	case configv1.AzurePlatformType:
+		return aws.Add(mgr, azure.NewReconciler(mgr))
```
dgoodwin: This looks like it should be azure.Add.

pkg/controller/secretannotator/aws/reconciler.go
```diff
@@ -0,0 +1,183 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	ccaws "github.com/openshift/cloud-credential-operator/pkg/aws"
+	"github.com/openshift/cloud-credential-operator/pkg/controller/utils"
+)
+
+const (
+	controllerName = "secretannotator"
```
dgoodwin: Same as below.

serbrech:
Hey @dgoodwin thanks for reviewing! 
The details are still moving around to be aligned with the refactoring of aws/azure as 2 different actuators. What I really wanted you to look into for now is if the way I am refactoring to handle multiple platform is aligned with your ideas.

dgoodwin:
Approach looks fine.

pkg/controller/secretannotator/azure/reconciler.go
serbrech: done

serbrech:
@dgoodwin much more ready for review now.
I'm gonna test it in our environment to verify the behavior manually.

pkg/azure/actuator.go
```diff
@@ -58,28 +62,58 @@ func (a *Actuator) IsValidMode() error {
 	return errors.New("invalid mode")
 }
 
+func isAzureCredentials(providerSpec *runtime.RawExtension) (bool, error) {
+	codec, err := minterv1.NewCodec()
+	if err != nil {
+		return false, err
+	}
```
serbrech: Decode as unknown to check for Kind.

pkg/azure/actuator.go
```diff
@@ -58,28 +62,58 @@ func (a *Actuator) IsValidMode() error {
...
+		log.WithField("kind", unknown.Kind).
+			Info("actuator handles only azure credentials")
+	}
+	return isAzure, nil
+}
+
 func (a *Actuator) Create(ctx context.Context, cr *minterv1.CredentialsRequest) error {
+	if isAzure, err := isAzureCredentials(cr.Spec.ProviderSpec); !isAzure {
+		return err
+	}
```
serbrech: If it is not azure credential, we return the error.
if it managed to decode but found a different Kind than expected, we return nil and the operation is a NoOp.
if it failed to decode (only if the schemes are wrong) we just return the error, which gets handled/logged at a higher level

serbrech:
Unless a higher level of filtering happens, the same logic needs to be implemented in AWS to not get errors for azure cred request 
6d421cf691c6

serbrech:
```
INFO[0000] setting up client for manager                
INFO[0000] setting up manager                           
INFO[0004] registering components                       
INFO[0004] setting up scheme                            
INFO[0004] setting up controller                        
INFO[0009] Setting up secret annotator. Platform Type is Azure 
INFO[0014] initializing Azure actuator                  
INFO[0014] starting the cmd                             
INFO[0016] validating cloud cred secret                  controller=secretannotator
INFO[0016] Platform is azure: using passthrough          controller=secretannotator
INFO[0016] syncing credentials request                   controller=credreq cr=openshift-cloud-credential-operator/cloud-credential-operator-iam-ro
INFO[0016] version has changed, updating progressing condition lastTransitionTime  new="[{operator 4.1.0-0.okd-2019-05-13-204940}]" old="[{operator local}]"
INFO[0016] syncing credentials request                   controller=credreq cr=openshift-cloud-credential-operator/openshift-image-registry
INFO[0016] syncing credentials request                   controller=credreq cr=openshift-cloud-credential-operator/openshift-ingress
INFO[0016] syncing credentials request                   controller=credreq cr=openshift-cloud-credential-operator/openshift-machine-api
INFO[1918] syncing credentials request                   controller=credreq cr=openshift-cloud-credential-operator/openshift-ingress
INFO[1918] syncing credentials request                   controller=credreq cr=openshift-cloud-credential-operator/openshift-ingress
ERRO[1918] error syncing credentials: secrets "cloud-credentials" already exists  controller=credreq cr=openshift-cloud-credential-operator/openshift-ingress secret=openshift-ingress-operator/cloud-credentials
ERRO[1918] unexpected error while syncing credentialsrequest: secrets "cloud-credentials" already exists  controller=credreq cr=openshift-cloud-credential-operator/openshift-ingress secret=openshift-ingress-operator/cloud-credentials
INFO[1919] syncing credentials request                   controller=credreq cr=openshift-cloud-credential-operator/openshift-ingress
INFO[1920] syncing credentials request                   controller=credreq cr=openshift-cloud-credential-operator/openshift-ingress
INFO[2097] syncing credentials request                   controller=credreq cr=openshift-cloud-credential-operator/openshift-ingress
INFO[2097] actuator handles only azure credentials       kind=AWSProviderSpec
INFO[2097] actuator handles only azure credentials       kind=AWSProviderSpec
INFO[2097] status has changed, updating                  controller=credreq cr=openshift-cloud-credential-operator/openshift-ingress secret=openshift-ingress-operator/cloud-credentials
INFO[2097] syncing credentials request                   controller=credreq cr=openshift-cloud-credential-operator/openshift-ingress
```

```
KUBECONFIG=~/tmp/os4j/auth/kubeconfig kubectl get secrets/cloud-credentials --namespace openshift-ingress-operator
NAME                TYPE     DATA   AGE
cloud-credentials   Opaque   7      29m
```

serbrech:
Can I get a review on this? @dgoodwin

pkg/azure/actuator.go
dgoodwin: Way better than what I suggested, nice.

serbrech: I think it could be pushed up to the predicates function, or even above that when we register the actuators. but that requires more refactoring, so this should do for now.
Do we let the AWS implementation log errors on azure creds for now?

dgoodwin: Are azure cred requests landing in the release image yet? (these come from each component's repo that needs creds for AWS)

serbrech: in Ingress Operator PR :
https://github.com/openshift/cluster-ingress-operator/pull/231/files

pkg/controller/secretannotator/secretannotator_controller.go
```diff
@@ -17,199 +17,32 @@ limitations under the License.
...
-
-	secret := &corev1.Secret{}
-	err := r.Get(context.Background(), request.NamespacedName, secret)
-	if err != nil {
-		r.logger.Debugf("secret not found: %v", err)
-		return reconcile.Result{}, err
+	switch platformType {
+	case configv1.AzurePlatformType:
+		return azure.Add(mgr, azure.NewReconciler(mgr))
+	case configv1.AWSPlatformType:
```
joelddiaz: go doesn't have automatic fallthroughs. need to put an explicit 'fallthrough' statement. right now on aws i get 
```
INFO[0001] setting up controller                        
INFO[0003] Setting up secret annotator. Platform Type is AWS 
ERRO[0003] platform not supportedunable to register controllers to the manager 
exit status 1
```

pkg/controller/secretannotator/constants/const.go
```diff
@@ -0,0 +1,25 @@
+package constants
+
+const (
+	ControllerName = "secretannotator"
+
+	// TODO: dynamically detect which environment we're running on
+	CloudCredSecretName      = "aws-creds"
```
joelddiaz: this const can likely go, as it should only be down in the new aws secret annotator controller and actuator

pkg/controller/secretannotator/aws/reconciler.go
```diff
@@ -0,0 +1,166 @@
...
+
+	ccaws "github.com/openshift/cloud-credential-operator/pkg/aws"
+	"github.com/openshift/cloud-credential-operator/pkg/controller/secretannotator/constants"
+	"github.com/openshift/cloud-credential-operator/pkg/controller/utils"
+)
+
+const (
+	// TODO: dynamically detect which environment we're running on
+	AWSCloudCredSecretName   = "aws-creds"
+	CloudCredSecretNamespace = "kube-system"
```
joelddiaz: remove this and use the shared consts instead

pkg/controller/secretannotator/azure/reconciler.go
```diff
@@ -0,0 +1,100 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/openshift/cloud-credential-operator/pkg/controller/secretannotator/constants"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	cloudCredSecretName      = "azure-credentials"
+	cloudCredSecretNamespace = "kube-system"
```
joelddiaz: probably should use the shared consts

pkg/controller/secretannotator/secretannotator_controller.go
serbrech: ah thanks

serbrech: Not fond of the `fallthrough`keyword, I explicitely call the aws one in both cases instead

serbrech:
@joelddiaz @dgoodwin Added the same check for the aws actuator as for the azure one.
I suggest you find a better structure to make filter the cred request at a higher level so that it's more generic (identify the request in the watch predicate or something similar).
But I'd like to keep this out of the scope of this PR which is already quite big in terms of refactoring.

pkg/aws/actuator/actuator.go
```diff
@@ -1142,3 +1142,21 @@ func (a *AWSActuator) loadClusterUUID(logger log.FieldLogger) (configv1.ClusterI
...
+	}
+	unknown := runtime.Unknown{}
+	err = codec.DecodeProviderSpec(providerSpec, &unknown)
+	if err != nil {
+		return false, err
+	}
+	isAWS := unknown.Kind == reflect.TypeOf(minterv1.AWSProviderSpec{}).Name()
+	if !isAWS {
+		log.WithField("kind", unknown.Kind).
+			Info("actuator handles only azure credentials")
```
joelddiaz: s/azure/aws/

serbrech: the copy pasta fallacies. fixed in amended commit.

serbrech:
@dgoodwin you need to cancel your hold for this to go further

PR #70, opened 2019-05-24
Add OpenStack actuator

Fedosin:
This commit adds a simple OpenStack actuator that syncs cloud credentials.


pkg/openstack/actuator.go
```diff
@@ -0,0 +1,260 @@
...
+		} else {
+			decoded := string(keyBytes)
+			existingClouds = string(decoded)
+			logger.Debug("found clouds.yaml in target secret")
+		}
+	}
+	return existingSecret, existingClouds
+}
+
+func (a *OpenStackActuator) getCloudCredentialsSecret(ctx context.Context, logger log.FieldLogger) (*corev1.Secret, string, error) {
```
dgoodwin: Could you rename this to getRootCloudCredentialsSecret.

pkg/openstack/actuator.go
```diff
@@ -0,0 +1,260 @@
...
+		logger.WithError(err).Error(msg)
+		return nil, "", &actuatoriface.ActuatorError{
+			ErrReason: minterv1.CredentialsProvisionFailure,
+			Message:   fmt.Sprintf("%v: %v", msg, err),
+		}
+	}
+
+	keyBytes, ok := cloudCredSecret.Data["clouds.yaml"]
+	if !ok {
+		logger.Warning("secret did not have expected key: clouds.yaml, will be regenerated")
```
dgoodwin: This looks inaccurate, nothing would regenerate the root creds secret would it? Probably a hard error at this point.

pkg/apis/cloudcredential/v1/openstack_types.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// TODO: these types should eventually be broken out, along with the actuator,
+// to a separate repo.
+
+// OpenStackProviderSpec the specification of the credentials request in OpenStack.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type OpenStackProviderSpec struct {
+	metav1.TypeMeta `json:",inline"`
+}
```
dgoodwin: The empty spec here is interesting. For AWS we did two modes, minting where we create the creds you specifically asked for, and passthrough where if we don't have enough permission to mint we just copy the root creds everywhere. 

With the empty spec here and the pure copying this is similar to what MS did for Azure's initial cut which is just passthrough. There was an additional bit of functionality where we check and annotate the root creds secret with which mode they support which you can see here: https://github.com/openshift/cloud-credential-operator/tree/master/pkg/controller/secretannotator. Ideally it would be good to do the same for OpenStack even though we're just always saying passthrough, we'd be a little closer to a full implementation.

Do we plan to fill this out with the ability to mint fine grained credentials/permissions in the future?

Fedosin: Hi! Yes, absolutely we have these plans. I decided to keep things simple at first and just copy the data from the system secret, because it allows us to easily test our patches for operators.
When the codebase is stable enough I'm going to implement minting of fine grained credentials, like it's done with Azure, i.e. with Rolebindings in the spec.

pkg/openstack/actuator.go
Fedosin: Done

pkg/openstack/actuator.go
Fedosin: Done

pkg/apis/cloudcredential/v1/openstack_types.go
dgoodwin: Excellent, go for AWS as an example when you get to that point, Azure doesn't have full minting yet they're just passthrough as well so far.

PR #71, opened 2019-06-14
generate deployment and crd into the right filename

joelddiaz:
remove the memory limit as originally intented in #68.

the deployment yaml changes are a no-op. just changes the order of the keys.

update the location where we copy the crd to (match the current filenames in manifets/*).

update gitignore to cover vscode directories.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,8 @@ manifests:
 	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd
 	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go rbac --name cloud-credential-operator
 	# kustomize and move to manifests dir for release image:
-	kustomize build config > manifests/0000_30_cloud-credential-operator_01_deployment.yaml
-	cp config/crds/cloudcredential_v1_credentialsrequest.yaml manifests/0000_30_cloud-credential-operator_00_v1_crd.yaml
+	kustomize build config > manifests/01_deployment.yaml
+	cp config/crds/cloudcredential_v1_credentialsrequest.yaml manifests/00_v1_crd.yaml
 
 # Run go fmt against code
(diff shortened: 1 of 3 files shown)


config/manager/deployment.yaml
```diff
@@ -67,6 +67,8 @@ spec:
         imagePullPolicy: IfNotPresent
         name: manager
         resources:
+          limits:
+            memory: 500Mi
```
dgoodwin: This was explicitly taken out in a recent commit, we don't want a memory limit anymore. Are we doing anything with kubebuilder that regenerates this file? Afaik it's just an initial one time generation, after which we have full control.

dgoodwin: Hmm, I wonder if we properly remove the memory limit, was it missed in 01_deployment.yaml? It should be gone in both it sounds like.

Makefile
```diff
@@ -40,8 +40,8 @@ manifests:
 	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd
 	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go rbac --name cloud-credential-operator
 	# kustomize and move to manifests dir for release image:
-	kustomize build config > manifests/0000_30_cloud-credential-operator_01_deployment.yaml
-	cp config/crds/cloudcredential_v1_credentialsrequest.yaml manifests/0000_30_cloud-credential-operator_00_v1_crd.yaml
+	kustomize build config > manifests/01_deployment.yaml
+	cp config/crds/cloudcredential_v1_credentialsrequest.yaml manifests/00_v1_crd.yaml
```
dgoodwin: The manifests are used for CVO release image ordering I thought, and the filenames were specific to the ordering we needed. Are we sure this is safe to change now? Maybe something has changed I'm not aware of.

PR #74, opened 2019-06-19
Update sigs.k8s.io/controller-runtime to release-2.0 and bump k8s to 1.14.1

ingvagabund:
There is already a bunch of other components imported in the installer that require newer version of controller-runtime. Once they get updated, installer will require older and new version of controller-runtime at the same time which are not compatible.

Controller name got renamed from `credentialsrequest-controller` to `credentialsrequest_controller`. Newer controller-runtime uses the controller name as prefix in metric names when they are registered. Unfortunately, a metric name is not allowed to contain `-`, only `_`. Thus the change. Not ideal though.


dgoodwin:
/lgtm

Thanks @ingvagabund

PR #77, opened 2019-06-27
add support for minting creds on GCP

joelddiaz:
- add GCP provider config
- vendor GCP dependencies
- add GCP secret annotator
- add GCP actuator (initial support for mint-mode only)
- add GCP mock client


pkg/gcp/actuator/actuator.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+
+	svcAcct, err := getServiceAccount(gcpClient, gcpStatus.ServiceAccountID)
+	if err != nil {
+		return fmt.Errorf("error getting service account details: %v", err)
+	}
+
+	if err := removeAllPolicyBindingsFromServiceAccount(gcpClient, svcAcct); err != nil {
+		return fmt.Errorf("error removing service account policy bindings: %v", err)
+	}
```
dgoodwin: Is this required before it will let you delete the service account?

pkg/gcp/actuator/policy.go
```diff
@@ -0,0 +1,248 @@
...
+
+func addPolicyBinding(policy *cloudresourcemanager.Policy, role *iamadminpb.Role, svcAcct *iamadminpb.ServiceAccount) bool {
+	for i, binding := range policy.Bindings {
+		if binding.Role == role.Name {
+			return addServiceAccountToBinding(svcAcct, policy.Bindings[i])
+		}
+	}
+
+	// if we didn't find an existing binding entry, then make one
+	createServiceAccountRoleBinding(policy, role, svcAcct)
```
dgoodwin: What is the difference between role binding and policy binding?

pkg/controller/utils/gcp/utils.go
```diff
@@ -0,0 +1,174 @@
...
+	if len(disallowedPerms) > 0 {
+		logger.Warn("Detected some unallowed permissions: %s", disallowedPerms)
+	}
+
+	return len(disallowedPerms) == 0, nil
+
+}
+
+// CheckServicesEnabled will take a list of GCP permissions, and see whether each permissions'
+// related API is enabled.
```
dgoodwin: What does it mean for APIs to be disabled in this context? Is this for a particular user and something in addition to permission checking?

pkg/gcp/actuator/actuator.go
```diff
@@ -0,0 +1,597 @@
...
+				Message:   fmt.Sprintf("%s: %v", msg, err),
+			}
+		}
+	}
+
+	return nil
+}
+
+func (a *Actuator) syncPassthrough(ctx context.Context, cr *minterv1.CredentialsRequest, cloudCredsSecret *corev1.Secret, logger log.FieldLogger) error {
+	return fmt.Errorf("PASSTHROUGH ON GCP NOT YET IMPLEMENTED")
```
dgoodwin: I added a Jira for this for next sprint.

pkg/gcp/actuator/actuator.go
```diff
@@ -0,0 +1,597 @@
...
+		if status.Code(err) == codes.NotFound {
+			logger.WithField("serviceaccount", gcpStatus.ServiceAccountID).Debug("service account does not exist, creating")
+			if rootGCPClient == nil {
+				return fmt.Errorf("no root GCP client available, cred secret may not exist: %s/%s", rootGCPCredsSecretNamespace, rootGCPCredsSecret)
+			}
+		} else {
+			return fmt.Errorf("error checking for existing service account: %v", err)
+		}
+
+		svcAcctName, err := generateNameWithFieldLimits(infraName, 50, cr.Name, 49) // 100 total char limit
```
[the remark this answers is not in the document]
joelddiaz: basically reusing the field name chopping, and discarding the random suffix since the service account name field doesn't need any random chars to make the field unique (so we preserve more of the infraName+credRequestName.

i can add comments to clarify, but would it be better to just split the generateNameWithFieldLimits into two functions. one for chopping up the individual fields, and one for adding the random suffix?

pkg/gcp/actuator/actuator.go
joelddiaz: Sadly, yes. GCP lets you remove a service account without undoing the policy bindings. For example, in the following output the serviceaccount "test-serviceaccount" has been deleted, but the policy binding still exists.
```
[jdiaz@minigoomba ~]$ gcloud projects get-iam-policy openshift-hive-dev --format=json | jq  '.bindings[] | select(.role == "roles/appengine.codeViewer")'
{
  "members": [
    "serviceAccount:test-serviceaccount@openshift-hive-dev.iam.gserviceaccount.com"
  ],
  "role": "roles/appengine.codeViewer"
}
```
So this policy cleanup is needed to avoid leaving cruft around GCP.

pkg/gcp/actuator/policy.go
joelddiaz: A policy contains a list of bindings, and each binding entry is what ties a specific role to a list of users/members:
Here is a list of binding entries for everything related to AppEngine:

```
[jdiaz@minigoomba ~]$ gcloud projects get-iam-policy openshift-hive-dev --format=json | jq  '.bindings[] | select(.role | contains("roles/appengine"))'
{
  "members": [
    "serviceAccount:jdiaz-tect-6-my-test-cre-4jl2w@openshift-hive-dev.iam.gserviceaccount.com",
    "serviceAccount:test-serviceaccount-12345@openshift-hive-dev.iam.gserviceaccount.com"
  ],
  "role": "roles/appengine.appAdmin"
}
{
  "members": [
    "serviceAccount:jdiaz-tect-6-my-test-cre-4jl2w@openshift-hive-dev.iam.gserviceaccount.com"
  ],
  "role": "roles/appengine.appViewer"
}
```
Two binding entries above. You can see that the appAdmin role is bound to two service accounts, and the appViewer role is bound to one service account.

If I've used the term 'policy binding' somewhere, that's probably a mistake.

pkg/controller/utils/gcp/utils.go
joelddiaz: GCP has this idea of enabling/disabling individual service APIs at the project level. I can have complete admin permissions on a project, and still not have API access to different services until the API is enabled.

```
[jdiaz@minigoomba ~]$ gcloud spanner instances list
API [spanner.googleapis.com] not enabled on project

[jdiaz@minigoomba ~]$ gcloud services list --available --format='table(config.name,state)' --filter='config.name=spanner.googleapis.com'
NAME                    STATE
spanner.googleapis.com  DISABLED
```

pkg/apis/cloudcredential/v1/gcp_types.go
```diff
@@ -0,0 +1,38 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TODO: these types should eventually be broken out, along with the actuator, to a separate repo.
+
+// GCPProviderSpec contains the required information to create a service account with policy bindings in GCP.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type GCPProviderSpec struct {
+	metav1.TypeMeta `json:",inline"`
+	PredefinedRoles []string `json:"predefinedRoles"`
```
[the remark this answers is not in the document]
joelddiaz: done

pkg/gcp/actuator/actuator.go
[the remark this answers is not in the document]
joelddiaz: split into two functions and added comments when generating the strings to indicate the field limits we're working with

pkg/gcp/actuator/actuator.go
```diff
@@ -0,0 +1,597 @@
...
+
+	err = a.Client.Status().Update(ctx, cr)
+	if err != nil {
+		logger.WithError(err).Error("error updating credentials request status")
+		return err
+	}
+	return nil
+}
+
+// getCloudCredentialsSecret will return the cluster's GCP cloud cred secret if it exists and is properly annotated
```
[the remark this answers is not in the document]
joelddiaz: done

pkg/gcp/actuator/actuator.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+	unknown := runtime.Unknown{}
+	err = codec.DecodeProviderSpec(providerSpec, &unknown)
+	if err != nil {
+		return false, err
+	}
+	isGCP := unknown.Kind == reflect.TypeOf(minterv1.GCPProviderSpec{}).Name()
+	if !isGCP {
+		log.WithField("kind", unknown.Kind).
+			Info("actuator handles only gcp credentials")
```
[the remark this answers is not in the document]
joelddiaz: done

pkg/gcp/actuator/actuator.go
joelddiaz: I guess I didn't really answer your question. It is **not** required to remove the role bindings from the policy before deleting the service account, but it will then leave the role binding entries for the now non-existent service account.
So we really should be cleaning up the policy before removing the service account.

pkg/gcp/actuator/actuator.go
```diff
@@ -0,0 +1,602 @@
...
+	gcpStatus, err := decodeProviderStatus(a.Codec, cr)
+	if err != nil {
+		return err
+	}
+
+	if gcpStatus.ServiceAccountID == "" {
+		// The service account id has a max length of 30 chars
+		// split it into 12-11-5 where the resuling string becomes:
+		// <infraName chopped to 12 chars>-<crName chopped to 11 chars>-<random 5 chars>
+		svcAcctID, err := generateUniqueNameWithFieldLimits(infraName, 12, cr.Name, 11)
```
dgoodwin: Whats the situation for cleanup right now, I assume we're leaking resources until we work out something with installer team?

dgoodwin: And does deleting a CR cleanup the objects?

joelddiaz: deleting the CR will delete all the created objects in the cloud, but if the cluster gets nuked (like the AWS uninstall process works), then we'll have the remnant objects to deal with

PR #78, opened 2019-06-27
fix cloud name typo

joelddiaz:


dgoodwin:
Why is this not merging?

/lgtm
/approve

PR #79, opened 2019-07-01
ignore credentialsRequests for other clouds

joelddiaz:
currently COO will go down a codepath for CredentialsRequest objects that are intended for a different cloud such that the Status.Provisioned will show True even though the credrequest was ignored.
    
fix up the credentialsrequest controller to actively ignore credentialsrequests for different cloud platforms than the one the cluster has been deployed to.
    
update the status calculation to also ignore the invalid credentialsrequests objects when reporting the cluster operator status.

TODO: add condition to credRequests for other clouds/platforms

this PR builds on top of #77

### DIFF
--- a/pkg/apis/cloudcredential/v1/credentialsrequest_types.go
+++ b/pkg/apis/cloudcredential/v1/credentialsrequest_types.go
@@ -131,6 +131,12 @@ const (
 	// CredentialsDeprovisionFailure is true whenever there is an error when trying
 	// to clean up any previously-created cloud resources
 	CredentialsDeprovisionFailure CredentialsRequestConditionType = "CredentialsDeprovisionFailure"
+	// Ignored is true when the CredentialsRequest's ProviderSpec is for
+	// a different infrastructure platform than what the cluster has been
+	// deployed to. This is normal as the release image contains CredentialsRequests for all
+	// possible clouds/infrastructure, and cloud-credential-operator will only act on the
+	// CredentialsRequests where the cloud/infra matches.
+	Ignored CredentialsRequestConditionType = "Ignored"
 )
 
 func init() {
--- a/pkg/controller/credentialsrequest/credentialsrequest_controller.go
+++ b/pkg/controller/credentialsrequest/credentialsrequest_controller.go
@@ -30,6 +30,8 @@ import (
 	"github.com/openshift/cloud-credential-operator/pkg/controller/internalcontroller"
 	"github.com/openshift/cloud-credential-operator/pkg/controller/utils"
 
+	configv1 "github.com/openshift/api/config/v1"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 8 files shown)


pkg/apis/cloudcredential/v1/credentialsrequest_types.go
```diff
@@ -131,6 +131,10 @@ const (
 	// CredentialsDeprovisionFailure is true whenever there is an error when trying
 	// to clean up any previously-created cloud resources
 	CredentialsDeprovisionFailure CredentialsRequestConditionType = "CredentialsDeprovisionFailure"
+	// OtherInfrastrucutre is true when the CredentialsRequest's ProviderSpec is for
+	// a different infrastructure platform than what the cluster has been
+	// deployed to.
+	OtherInfrastructure CredentialsRequestConditionType = "OtherInfrastructure"
```
dgoodwin: WDYT about renaming the condition to Ignored, reason "ProviderMismatch" or "InfrastructureMismatch" or "PlatformMismatch"

dgoodwin: We should also expand on the doc comment just to note that this is normal in that we deploy creds for all clouds in the release image and only act on those if the cloud infra matches.

pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -284,6 +290,22 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
 	origCR := cr
 	cr = cr.DeepCopy()
 
+	// Ignore CR if it's for a different cloud
+	differentCloud, err := isCRForDifferentCloud(cr, r.platformType)
```
dgoodwin: flipping the boolean response here might be a little more readable, something like "matchesPlatform" returning true.

pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -532,6 +554,17 @@ func setCredentialsDeprovisionFailureCondition(cr *minterv1.CredentialsRequest,
 		status, reason, msg, updateCheck)
 }
 
+func setOtherInfrastructureCondition(cr *minterv1.CredentialsRequest, failed bool, clusterPlatform configv1.PlatformType) {
```
dgoodwin: setIgnoredCondition if you're ok with the above suggestion.

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
joelddiaz: yes, that makes way more sense. done.

dgoodwin:
Looks good lets just squash and get CI clear.

pkg/controller/credentialsrequest/credentialsrequest_controller.go
joelddiaz: done

PR #86, opened 2019-07-08
implement GCP passthrough mode (and add service API checks)

joelddiaz:
- [x] handle the case when the root creds are in passthrough mode.
- [x] add checks for passthrough and mint mode where we check that the service APIs are enabled/available before proceeding to process the CredentialsRequest.

Note that passthrough mode is handled differently than in AWS. In AWS we work with a static list of permissions that are needed for the cluster to run with. For GCP passthrough mode, the decision on whether we can satisfy a CredentialsRequest is dynamic. All we absolutely need for passthrough mode is the ability to list service APIs (to determine whether any particular service API is enabled), the ability to get the details for a specific role (so we can determine whether the role exists and the permissions attached to that role), and the ability to get the details of a project (so we can get the project number for a given project name). All the other permissions attached to the root creds in passthrough mode would pass/fail a permissions check during the TestIamPermissions() calls.

### DIFF
--- a/pkg/controller/utils/gcp/utils.go
+++ b/pkg/controller/utils/gcp/utils.go
@@ -29,6 +29,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 var (
@@ -52,14 +53,26 @@ var (
 		"resourcemanager.projects.setIamPolicy",
 	}
 
+	// CredPassthroughPermissions is a list of GCP permissions needed to run in passthrough mode.
 	CredPassthroughPermissions = []string{
 		// Query API availability
 		"serviceusage.services.list",
+		// Query getting current project details
 		"resourcemanager.projects.get",
+		// Query role existence and permissions attached to roles
+		"iam.roles.get",
 	}
 
 	credentailRequestScheme = runtime.NewScheme()
 	credentialRequestCodec  = serializer.NewCodecFactory(credentailRequestScheme)
+
+	invalidPermsForProjectScopedPermissionsTesting = []string{
+		// checking whether we have permissions resourcemanager.projects.list returns an error
+		// (as it is invalid when checking against a project resource),
+		// but listing projects is always allowed (it simply returns a list of projects you
+		// have projects.get permissions for)
+		"resourcemanager.projects.list",
+	}
 )
 
 const (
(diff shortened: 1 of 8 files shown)


pkg/gcp/actuator/actuator.go
```diff
@@ -348,37 +431,59 @@ func (a *Actuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequ
...
-			return true, nil
+			return serviceAPIsEnabled, true, nil
 		}
 
 	} else {
 		// passthrough-specifc check here
-		return true, fmt.Errorf("PASSTHROUGH NOT IMPLEMENTED")
+
+		// build client with the target secret creds, then check permissions
+		testClient, err := a.buildClientFromCRSecret(cr)
```
dgoodwin: If we're doing passthrough here, why do we go to the target secret rather than using the root creds?

Just thinking about scenario where the root creds have been rotated and the old ones are no longer valid,  would be fail here because the target secret is still on the old creds and then we fail to copy over the new? (if so unit test would be good here)

pkg/gcp/actuator/actuator.go
```diff
@@ -348,37 +431,59 @@ func (a *Actuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequ
...
 	}
 
 	// If we've made it this far, then there are no updates needed
-	return false, nil
+	return serviceAPIsEnabled, false, nil
+}
+
+func (a *Actuator) buildClientFromCRSecret(cr *minterv1.CredentialsRequest) (ccgcp.Client, error) {
+	authJSON, err := loadCredsFromSecret(a.Client, cr.Spec.SecretRef.Namespace, cr.Spec.SecretRef.Name)
+	fmt.Printf("GOT %s\n", string(authJSON))
```
dgoodwin: leftover debugging.

pkg/gcp/actuator/actuator.go
joelddiaz: This is due to the dynamic permissions checking for GCP (as opposed to the static-mode we implemented for AWS). 'passthrough' basically reads in GCP as 'these creds have the potential to be passed through depending on the permissions being requested' versus in AWS where we are blindly assuming that the AWS secret with the passthrough annotation is good enough (based on the static list of perms in https://github.com/openshift/cloud-credential-operator/blob/master/pkg/aws/utils.go#L35 .

The thinking was that when we're checking whether an update is needed, if there is an existing secret, and it already has good-enough permissions to satisfy the credRequest, then just leave the secret alone as there is no update needed.

joelddiaz: To be clear, we could absolutely check against the cluster root creds, and then check whether the cluster root creds content is the same as the already-existing CR secret content, and decide needsUpdate() that way.

joelddiaz: If the existing target secret fails, then we'd return that we do indeed need to update, and that code path will handle checking the root creds before syncing the new secret contents.

pkg/gcp/actuator/actuator.go
joelddiaz: whoops. fixed.

pkg/controller/utils/gcp/utils.go
```diff
@@ -110,14 +115,31 @@ func checkServicesAndPermissions(gcpClient ccgcp.Client, permissionsList []strin
 	return servicesEnabled, nil
 }
 
+func filterPermissionsList(list []string) []string {
```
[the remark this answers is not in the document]
joelddiaz: i made it more generic and have now defined a list of permissions we know fail the TestIamPermissions() call when done in project scope

pkg/controller/utils/gcp/utils.go
```diff
@@ -110,14 +115,31 @@ func checkServicesAndPermissions(gcpClient ccgcp.Client, permissionsList []strin
...
+	filteredPerms := []string{}
+	for _, perm := range list {
+		// checking whether we have permissions resourcemanager.projects.list returns an error
+		// (as it is invalid when checking against a project resource),
+		// but listing projects is always allowed (it simply returns a list of projects you
+		// have projects.get permissions for)
+		if perm == "resourcemanager.projects.list" {
```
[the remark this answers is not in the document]
joelddiaz: done.

pkg/gcp/actuator/actuator.go
joelddiaz: Re-read your original comment, and for cert rotation, you're absolutely right. We should check the root creds, and then make sure those are synced to the CR target secret. Will update.

joelddiaz: ok. changed the logic to check the root creds, and then whether the CR target secret data is out-of-date.

dgoodwin:
LGTM, lets squash, thanks!

PR #87, opened 2019-07-09
added NewClientSession to AWS client for users that require session tokens

fool65c:
Overview: Added function NewClientSession in the aws client.  This function allows for the session token to be passed in and on to the aws api.

Reason for adding: The project [Openshift/installer](https://github.com/openshift/installer) uses cloud-credential-operator to [create an AWS client](https://github.com/openshift/installer/blob/4d88d536bc0f27c03713fcb4feeb27d5fae8a023/pkg/asset/installconfig/aws/permissions.go#L198).  The goal is to add this functionality here then modify the installer to call this new function

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -102,11 +102,11 @@ func NewClient(accessKeyID, secretAccessKey []byte, infraName string) (Client, e
 }
 
 // NewClientSession allows for clients with session tokens to connect
-func NewClientWithSessionToken(accessKeyID, secretAccessKey []byte, sessionToken string, infraName string) (Client, error) {
+func NewClientWithSessionToken(accessKeyID, secretAccessKey []byte, sessionToken, infraName string) (Client, error) {
 	awsConfig := &awssdk.Config{}
 
 	awsConfig.Credentials = credentials.NewStaticCredentials(
-		string(accessKeyID), string(secretAccessKey), sessionToken)
+		string(accessKeyID), string(secretAccessKey), string(sessionToken))
 
 	s, err := session.NewSession(awsConfig)
 	if err != nil {
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -98,11 +98,11 @@ func (c *awsClient) TagUser(input *iam.TagUserInput) (*iam.TagUserOutput, error)
 
 // NewClient creates our client wrapper object for the actual AWS clients we use.
 func NewClient(accessKeyID, secretAccessKey []byte, infraName string) (Client, error) {
-	return NewClientSession(accessKeyID, secretAccessKey, "", infraName)
+	return NewClientWithSessionToken(accessKeyID, secretAccessKey, "", infraName)
 }
 
 // NewClientSession allows for clients with session tokens to connect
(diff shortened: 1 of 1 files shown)


pkg/aws/client.go
```diff
@@ -98,10 +98,15 @@ func (c *awsClient) TagUser(input *iam.TagUserInput) (*iam.TagUserOutput, error)
 
 // NewClient creates our client wrapper object for the actual AWS clients we use.
 func NewClient(accessKeyID, secretAccessKey []byte, infraName string) (Client, error) {
+	return NewClientSession(accessKeyID, secretAccessKey, "", infraName)
+}
+
+// NewClientSession allows for clients with session tokens to connect
+func NewClientSession(accessKeyID, secretAccessKey []byte, sessionToken string, infraName string) (Client, error) {
```
dgoodwin: Sorry small nit, could we call this NewClientWithSessionToken?

And an even smaller nit, you can do "sessionToken, infraName string" in the signature.

dgoodwin:
Can you elaborate on the longer term picture? Once the cluster comes up we mint credentials for a variety of components, and beyond that we assume that credentials may change, or be added, or need to be fixed / replaced in some way. These short lived session tokens would (a) need to be passed through to the operator running in-cluster to mint the component creds, and that might work fine for install but (b) it seems like we'd be dead in the water throughout the lifecycle of the cluster, the token would expire, and we have no way to automatically get a new one as it requires 2FA?

Let me know if I've got any of that wrong, I am curious how we can support this, it didn't look possible last we looked.

pkg/aws/client.go
fool65c: updates

fool65c:
You are correct, I took a look at the installer again and found where they are passing the creds as secretes.  Which means this would only work for the install and have long term side effects.  

We are pulling these credentials from vault, so maybe there is a pathway there.

dgoodwin:
That's interesting, could we discuss in the group g arch call tomorrow? Also who is this. :)

fool65c:
Hey this is Kevin, I work at a finical firm where I'm trying to get some openshift adoption while still playing within the rules 😄 of our accounts

dgoodwin:
Oh my mistake I assumed you were RH. Great thanks for trying to address this.

If you would like to explore adding vault support I would be happy to help see if we could fit it in. If so maybe just open an issue and we can start discussing there, I don't know a lot about vault so might have some basic questions.

I'll start asking around folks and see what they think in any case.

dgoodwin:
We've also played with the idea of a standalone utility. We do support removing the root creds from the cluster once it's up, everything should hum along fine until credentials changes are needed likely in the event of an OpenShift release upgrade, at which point the operator will start reporting that there's a problem because we can't fix things as we don't have root creds.

We could work out a standalone CLI utility that runs the credentials reconcile loop with credentials provided by the admin that would then get the creds back into sync.

If you wanted to go this route we can explore that as well.

dgoodwin:
Stale over a year, probably need to close this. Manual mode is available now and some work is being discussed around STS token support.

/close

PR #89, opened 2019-07-11
Bug 1729021: utilize the permissionBoundary of the current user if available

balpert89:
Our IAM root user has a permission boundary set. This means, every IAM user created by that root user has to have the permission boundary set.

We experienced failure in the installation of Openshift 4 on AWS because the user object to be created fails to pull the permission boundary of the provided user that is used to do the installation.

### DIFF
--- a/pkg/controller/credentialsrequest/credentialsrequest_controller_test.go
+++ b/pkg/controller/credentialsrequest/credentialsrequest_controller_test.go
@@ -165,6 +165,7 @@ func TestCredentialsRequestReconcile(t *testing.T) {
 			},
 			mockRootAWSClient: func(mockCtrl *gomock.Controller) *mockaws.MockClient {
 				mockAWSClient := mockaws.NewMockClient(mockCtrl)
+				mockGetUser(mockAWSClient)
 				mockCreateUser(mockAWSClient)
 				mockPutUserPolicy(mockAWSClient)
 				mockCreateAccessKey(mockAWSClient, testAWSAccessKeyID, testAWSSecretAccessKey)
@@ -219,6 +220,7 @@ func TestCredentialsRequestReconcile(t *testing.T) {
 			},
 			mockRootAWSClient: func(mockCtrl *gomock.Controller) *mockaws.MockClient {
 				mockAWSClient := mockaws.NewMockClient(mockCtrl)
+				mockGetUser(mockAWSClient)
 				mockCreateUser(mockAWSClient)
 				mockPutUserPolicy(mockAWSClient)
 				mockCreateAccessKey(mockAWSClient, testAWSAccessKeyID, testAWSSecretAccessKey)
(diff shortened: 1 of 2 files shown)


FlorinPeter:
@joelddiaz can you please take a look and give us an /ok-to-test ?

dgoodwin:
@joelddiaz has been digging in, this has opened the door to a lot of questions and new territory, we'll get this issue updated ASAP.

/ok-to-test

joelddiaz:
Can we get a test case that mocks up receiving root user details with the permissions boundary defined?

I think we'll eventually want to build on that (in another PR) as i think it would be wise to have a user creation process (when we're dealing with permissions boundaries) that goes like:
1) create user (with permissions boundary)
2) apply custom policy
3) (additional work in a future PR) check that the new user+policy can perform the actions (ie not blocked by permissions boundary)

< Totally new to permissions boundary disclaimer here >
From my understanding of these permissions boundaries, I can see the possibility that we can create a user, attach a policy of s3:*, and that succeeds. But then when the user actually tries to perform s3:* operations, it can be denied via the contents of the permissions boundary.

FlorinPeter:
> Can we get a test case that mocks up receiving root user details with the permissions boundary defined?

I think @balpert89 can try to help here but it will be nice if you can help him out as I am on vacation for the next 2 weeks.

> < Totally new to permissions boundary disclaimer here >
> From my understanding of these permissions boundaries, I can see the possibility that we can create a user, attach a policy of s3:_, and that succeeds. But then when the user actually tries to perform s3:_ operations, it can be denied via the contents of the permissions boundary.

Yes that's right, we have an managed AWS company account and those boundaries makes sure that a user can only access AWS objects if the MFA is set or a user is restricted to a particular region.

joelddiaz:
> > Can we get a test case that mocks up receiving root user details with the permissions boundary defined?
> 
> I think @balpert89 can try to help here but it will be nice if you can help him out as I am on vacation for the next 2 weeks.
> 
Absolutely. I'll put together something explaining what I think is necessary for the test case.

joelddiaz:
here's what i had in mind for a test case (i haven't updated the existing test cases...they just need a GetUser() mock where needed): https://gist.github.com/joelddiaz/8b352fb84296acdab48c8fa9781310c0

balpert89:
Hi @joelddiaz 

sorry for being blunt: I have tried to incorporate the test case into the test suite but failed :-(

The test case for `mockCreateUserWithPermissionsBoundary` fails because the returned object is not like the expected one (as `PermissionsBoundary` is missing):
```
    --- FAIL: TestCredentialsRequestReconcile/pass_along_any_existing_permissions_boundary (0.00s)
        actuator.go:1060: Unexpected call to *mock.MockClient.CreateUser([{
              UserName: "mycluster-test-aws-user"
            }]) at /home/balpert/go/src/cloud-credential-operator/vendor/github.com/openshift/cloud-credential-operator/pkg/aws/mock/client_generated.go:54 because: 
            Expected call at /home/balpert/go/src/cloud-credential-operator/pkg/controller/credentialsrequest/credentialsrequest_controller_test.go:1177 doesn't match the argument at index 0.
            Got: {
              UserName: "mycluster-test-aws-user"
            }
            Want: is equal to {
              PermissionsBoundary: "some:boundary:ARN:1234",
              UserName: "mycluster-test-aws-user"
            }
```

I am fairly new to this topic, could you give me some headsup? Where is the `CreateUserInput` created that is compared against the expected one in the mock function? How can I influence that? Is there a documentation for the testing framework, so I can start understanding it better?

Thank you, I appreciate any support :-)

joelddiaz:
> I am fairly new to this topic, could you give me some headsup? Where is the `CreateUserInput` created that is compared against the expected one in the mock function? How can I influence that? Is there a documentation for the testing framework, so I can start understanding it better?

The mockCreateUserWithPermissionsBoundary() is just a helper function that can be used in any test case where we want to ensure that the aws iam:CreateUser() call was called with the permissions boundary values filled out for the new user.

We use that function in addition to mockGetUserWithPermissionsBoundary(), so that when the new code does the iam:GetUser(), it gets a response of a user with permissions boundaries defined.

The CreateUserInput is in the code you added https://github.com/openshift/cloud-credential-operator/blob/05f3afc9fe0eb76dfe5663a627d402c52cf598ed/pkg/aws/actuator/actuator.go#L1057-L1067 , so when it does the GetUser(), it now gets the user with the permissions boundary, and that forces it down your new code that adds the

When the test case ("pass along any existing permissions boundary") is run, it calls into the Reconcile() function https://github.com/balpert89/cloud-credential-operator/blob/master/pkg/controller/credentialsrequest/credentialsrequest_controller_test.go#L824 , it will eventually call down into the AWS actuator code and go down your new code that tests for existing permissions boundaries.

I just pulled down your branch, and applied my patch, and it passed for me. You can see the commit in my branch I just pushed out: https://github.com/joelddiaz/cloud-credential-operator/commit/78a7b75c5162f91e4d35fb5bb0422c593806fce7

```
[jdiaz@minigoomba cloud-credential-operator (permissionsboundary $%)]$ go test ./pkg/controller/credentialsrequest/ -run "TestCredentialsRequestReconcile" -count 1
ok      github.com/openshift/cloud-credential-operator/pkg/controller/credentialsrequest        0.034s
```

I hope that explanation helped.

balpert89:
Hi @joelddiaz thanks for your explanation, it made a lot clearer and I have managed to also test this. I have included the commit and rebased the master branch, now the only test missing is `tide`. How can I start that? I guess it can be done only by you ;-)

pkg/controller/credentialsrequest/credentialsrequest_controller_test.go
```diff
@@ -802,6 +804,90 @@ func TestCredentialsRequestReconcile(t *testing.T) {
...
+						string(targetSecret.Data["aws_secret_access_key"]))
+				}
+				cr := getCR(c)
+				assert.True(t, cr.Status.Provisioned)
+				assert.Equal(t, int64(testCRGeneration), int64(cr.Status.LastSyncGeneration))
+				assert.NotNil(t, cr.Status.LastSyncTimestamp)
+			},
+		},
+		{
+			name: "pass along any existing permissions boundary",
```
joelddiaz: the "pass along any existing permissions boundary" test case is repeated twice. once here (line 850) and up above in line 808. just remove one of these redundant test cases

balpert89:
Hi @joelddiaz can you please trigger the `tide` test?

FlorinPeter:
@joelddiaz how can we move this forward?

dgoodwin:
Joel was out last week but back today, but based on his last comments

/lgtm

FlorinPeter:
@joelddiaz @dgoodwin thx. We also need this merged in 4.1. 
Is it possible to cherry pick this and to get this in the next z-release?

dgoodwin:
That may be a tough sell as the bar is high for 4.1, but I will ask.

PR #90, opened 2019-07-12
Revert "Bug 1726451: Add clusteroperator version level gate."

dgoodwin:
This reverts commit 4e137365484bdc5ea86ee5b0cac2fe60b3d3d36c.

We're seeing increased error rates as a result of this change. It is likely the upgrade bug we've seen in the wild but cannot reproduce, only now blocking upgrade because of the level gate. We suspect this was caused by the lack of leader election, but we did not backport leader election along with this change. We will try again next week with both changes.

### DIFF
--- a/manifests/10_cluster-operator.yaml
+++ b/manifests/10_cluster-operator.yaml
@@ -2,8 +2,3 @@ apiVersion: config.openshift.io/v1
 kind: ClusterOperator
 metadata:
   name: cloud-credential
-status:
-  versions:
-  - name: operator
-    version: "0.0.1-snapshot"
-


crawford:
I'm going to force merge this since it's a simple revert and removing the branch block is potentially going to introduce undesirable changes from other repos.

wking:
> We will try again next week with both changes.

Is this just a question of cherrypicking both #82 and #85?  It doesn't seem to have happened yet.

dgoodwin:
I suspect that it's just a matter of taking both but no it hasn't happened yet. I was not able to make sense of testgrid to be confident the problem is resolved in 4.2 but I haven't heard or seen any results where it's still occuring, so those two patches combined likely fix the problem.

However I'm not positive it's worth pursuing in 4.1, it doesn't seem to be causing major issues and I believe it can be worked around trivially.

wking:
What's the workaround?  Having the CVO ignore failed upgrades sounds pretty bad.

dgoodwin:
There's no actual upgrade involved here, it's just the operator reporting it's env var. We *think* (never been able to reproduce) that it was caused by lack of leader election, old pod still running and if the timing is right reports the old version. We have logs showing the new pod fighting with something to get the version it wants set.

So dangerous as that may be, in this case I don't think it's particularly dangerous, that old pod goes away and the problem should clear itself up next reconcile of a cred request. (30 mins?) 

Leader election appears to have fixed in 4.2, as far as we know.

PR #91, opened 2019-07-15
Support AWS temporary credentials

Vlatombe:
Similar to #87 

Context: in lots of AWS deployments, auth to AWS is done through temporary credentials. Either by [assuming a role](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user.html) from an IAM user, or from an instance running with an [instance profile](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html). Using root user, or long-term access keys is not recommended (see [Best Practices for managing AWS Access Keys](https://docs.aws.amazon.com/general/latest/gr/aws-access-keys-best-practices.html)).

This PR introduces the following changes:
* a variant of `NewClient` taking `*credentials.Value`
* `CheckPermissionsUsingQueryClient` has been changed in order to accept assumed roles. It uses `GetCallerIdentity` instead of `GetUser` as this is a more versatile way to determine auth context. Unfortunately the call to `SimulatePrincipalPolicy` is not reliable when using an assumed role (in my case, the simulation claims `ec2:RunInstances` is denied by AWS Organizations, but in reality it works just fine) , so the validation is skipped in such context.

### DIFF
--- a/pkg/aws/utils.go
+++ b/pkg/aws/utils.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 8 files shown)


pkg/aws/utils.go
```diff
@@ -111,41 +112,44 @@ func CheckCloudCredCreation(awsClient Client, logger log.FieldLogger) (bool, err
...
 		// warn about using the root creds, and just return that the creds are good enough
 		logger.Warn("Using the AWS account root user is not recommended: https://docs.aws.amazon.com/general/latest/gr/managing-aws-access-keys.html")
 		return true, nil
 	}
+	assumedRole, err := isAssumedRole(arn)
+	if err != nil {
+		return false, fmt.Errorf("error checking arn: %v", err)
+	}
+	if assumedRole {
+		logger.Warnf("Using assumed role %s, cannot validate permissions beforehand.", *arn)
```
joelddiaz: @dgoodwin this is an interesting development. The installer uses this code-path today to do the preinstall checks to better judge how much of a chance we have for a successful install (at least permissions-wise success).

Vlatombe:
FTR, with these patches I get somewhat of a successful installation, but then I get stuck because the whole `cloud-credential-operator` logic relies on managing users from the provided 'root' user.

Out of curiosity, is there any reason you didn't consider using roles instead to implement this operator? It is pretty easy to assume roles within instances, which gives you temporary credentials, without having to mess with access keys

joelddiaz:
> FTR, with these patches I get somewhat of a successful installation, but then I get stuck because the whole `cloud-credential-operator` logic relies on managing users from the provided 'root' user.
> 
We call it 'root' simply because it has permissions to create users with arbitrary policies, but the 'root' user needs these specific permissions https://github.com/openshift/cloud-credential-operator/blob/master/pkg/aws/utils.go#L21-L33 .

You can also try an install with the 'passthrough' permissions. No new users created, just reusing/passing through of the existing creds to the components that need them.

> Out of curiosity, is there any reason you didn't consider using roles instead to implement this operator? It is pretty easy to assume roles within instances, which gives you temporary credentials, without having to mess with access keys

I believe one of the reasons is that the cloud-cred-operator runs on worker nodes, and regular worker nodes get minimal permissions (ec2:DescribeInstance is all I see).

Vlatombe:
My overall problem is my AWS account is part of an AWS organization with overall policies which 'break' the policy simulator. (For example validation always fails on  `ec2:RunInstances` when in fact the user/role can perfectly run instances). Maybe some alternative would be to have a flag to disable permissions validation altogether.

joelddiaz:
> My overall problem is my AWS account is part of an AWS organization with overall policies which 'break' the policy simulator. (For example validation always fails on `ec2:RunInstances` when in fact the user/role can perfectly run instances). Maybe some alternative would be to have a flag to disable permissions validation altogether.

Help me understand with a little more details. Are you saying you've run the installer with some AWS credentials (which can in fact perform ec2:RunInstance), and the installer is failing to proceed b/c it says you're missing permissions? Or the installer ran fine, but as the cluster is coming up it is erroring out due to cloud-cred-operator saying it cannot satisfy some CredentialsRequest objects?

Vlatombe:
With the change in this PR, I can use an assumed role to perform the installation itself. It completes the bootstrap and waits forever for some components to be up. The control plane is accessible and `cloud-cred-operator` complains that it is not able to satisfy requests because of missing permissions.

joelddiaz:
It's not clear to me with the current approach, how temporary creds (ie max 1hr lifetime) can be used. Off the top of my head, the in-cluster registry needs S3, the ingress operator needs ec2 access (for managing load balancers), and the machine-controller (not sure about the name, but the operator that spins up new instances) all need consistent access to interact with AWS.

Vlatombe:
What I see usually done in AWS setups to avoid dealing with long-term access keys:
* Set up roles per service with appropriate privileges
* `cloud-cred-operator` runs on a VM whose instance profile allows it to get temporary credentials (through instance metadata api) to assume other roles and hand out temporary credentials from it.
* Possibly, the assumed roles can be protected using an [external id](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html) that could be stored in a k8s secret in `cloud-cred-operator` namespace.

PR #93, opened 2019-07-25
[gcp] Allow skipping service checks

csrwng:
Introduces support for the GCP_SKIP_SERVICE_CHECK environment variable
and makes it default to "true" for the operator deployment.

### DIFF
--- a/pkg/apis/cloudcredential/v1/gcp_types.go
+++ b/pkg/apis/cloudcredential/v1/gcp_types.go
@@ -29,6 +29,10 @@ type GCPProviderSpec struct {
 	// PredefinedRoles is the list of GCP pre-defined roles
 	// that the CredentialsRequest requires.
 	PredefinedRoles []string `json:"predefinedRoles"`
+	// SkipServiceCheck can be set to true to skip the check whether the requested roles
+	// have the necessary services enabled
+	// +optional
+	SkipServiceCheck bool `json:"skipServiceCheck,omitempty"`
 }
 
 // GCPProviderStatus contains the status of the GCP credentials request.
--- a/pkg/gcp/actuator/actuator.go
+++ b/pkg/gcp/actuator/actuator.go
@@ -318,9 +318,10 @@ func (a *Actuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequest
 	if err != nil {
 		return err
 	}
-	if !serviceAPIsEnabled {
+	if !gcpSpec.SkipServiceCheck && !serviceAPIsEnabled {
 		return fmt.Errorf("not all required service APIs are enabled")
 	}
+	serviceAPIsEnabled = true
 
 	// Create service account if necessary
 	var serviceAccount *iamadminpb.ServiceAccount
@@ -413,9 +414,10 @@ func (a *Actuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequ
 		return false, true, fmt.Errorf("error checking whether service APIs are enabled: %v", err)
 	}
 
-	if !serviceAPIsEnabled {
+	if !gcpSpec.SkipServiceCheck && !serviceAPIsEnabled {
 		return serviceAPIsEnabled, true, nil
 	}
+	serviceAPIsEnabled = true
 
 	// If the secret simply doesn't exist, we definitely need an update
 	exists, err := a.Exists(ctx, cr)


csrwng:
@dgoodwin ptal
/cc @abhinavdahiya

abhinavdahiya:
Hmm.. do we want each credential request to do this..?

and I think we should log that service check failed but skipped as requested.

abhinavdahiya:
The cred request failed for:
```
kind: CredentialsRequest
metadata:
  creationTimestamp: "2019-07-25T21:20:16Z"
  finalizers:
  - cloudcredential.openshift.io/deprovision
  generation: 1
  labels:
    controller-tools.k8s.io: "1.0"
  name: openshift-image-registry-gcs
  resourceVersion: "2582"
  selfLink: /apis/cloudcredential.openshift.io/v1/namespaces/openshift-cloud-credential-operator/credentialsrequests/openshift-image-registry-gcs
  uid: ff258b17-af21-11e9-a90b-42010a000005
spec:
  providerSpec:
    kind: GCPProviderSpec
    predefinedRoles:
    - roles/storage.admin
    - roles/iam.serviceAccountUser
  secretRef:
    name: installer-cloud-credentials
    namespace: openshift-image-registry
status:
  conditions:
  - lastProbeTime: "2019-07-25T21:21:32Z"
    lastTransitionTime: "2019-07-25T21:21:32Z"
    message: 'failed to grant creds: not all required service APIs are enabled'
    reason: CredentialsProvisionFailure
    status: "True"
    type: CredentialsProvisionFailure
  lastSyncGeneration: 0
  provisioned: false
```

```
gcloud iam roles describe roles/storage.admin
description: Full control of GCS resources.
etag: AA==
includedPermissions:
- firebase.projects.get
- resourcemanager.projects.get
- resourcemanager.projects.list
- storage.buckets.create
- storage.buckets.delete
- storage.buckets.get
- storage.buckets.getIamPolicy
- storage.buckets.list
- storage.buckets.setIamPolicy
- storage.buckets.update
- storage.objects.create
- storage.objects.delete
- storage.objects.get
- storage.objects.getIamPolicy
- storage.objects.list
- storage.objects.setIamPolicy
- storage.objects.update
name: roles/storage.admin
stage: GA
title: Storage Admin
```

So it's failing due to `firebase.projects.get` isn't enabled.. (I don't think that's fair for it to be required to be enabled)

bucket create

And since only `role/storage.admin` allows buckets to be created/delete there isn't much choice available. :(

csrwng:
> Hmm.. do we want each credential request to do this..?

Ahh, you mean that we should add it as a part of the credentialsrequest provider spec?

abhinavdahiya:
> > Hmm.. do we want each credential request to do this..?
> 
> Ahh, you mean that we should add it as a part of the credentialsrequest provider spec?

Yeah, the requestor is taking ownership of skipping as it knows it won't use the disabled ones..

csrwng:
> and I think we should log that service check failed but skipped as requested.

We do log a warning with the services that are not enabled (which would still print out in the log): https://github.com/openshift/cloud-credential-operator/blob/3b8eefaa24b77c26104fcfaa239a1201221415c7/pkg/controller/utils/gcp/utils.go#L204

csrwng:
updated

pkg/apis/cloudcredential/v1/gcp_types.go
```diff
@@ -29,6 +29,9 @@ type GCPProviderSpec struct {
 	// PredefinedRoles is the list of GCP pre-defined roles
 	// that the CredentialsRequest requires.
 	PredefinedRoles []string `json:"predefinedRoles"`
+	// SkipServiceCheck can be set to true to skip the check whether the requested roles
+	// have the necessary services enabled
+	SkipServiceCheck bool `json:"skipServiceCheck"`
```
abhinavdahiya: this is optional, right so omitempty/ and `// +optional` ??

pkg/gcp/actuator/actuator.go
```diff
@@ -318,7 +318,7 @@ func (a *Actuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequest
 	if err != nil {
 		return err
 	}
-	if !serviceAPIsEnabled {
+	if !gcpSpec.SkipServiceCheck && !serviceAPIsEnabled {
 		return fmt.Errorf("not all required service APIs are enabled")
 	}
```
abhinavdahiya: `serviceAPIsEnabled = true` should also be useful...

csrwng:
updated

PR #94, opened 2019-07-26
Split call to check permissions into chunks

csrwng:
The call to check available permissions only allows 100 permissions at a
time per check. This change splits the call into multiple calls when
there are more than 100 permissions to check.

### DIFF
--- a/pkg/controller/utils/gcp/utils.go
+++ b/pkg/controller/utils/gcp/utils.go
@@ -147,37 +147,40 @@ func filterOutPermissions(permList, filterOutList []string) []string {
 // gcpClient creds have sufficient permissions to perform the actions.
 // Will return true/false indicating whether the permissions are sufficient.
 func CheckPermissionsAgainstPermissionList(gcpClient ccgcp.Client, permList []string, logger log.FieldLogger) (bool, error) {
+
 	projectName := gcpClient.GetProjectName()
 
 	filteredPermList := filterOutPermissions(permList, invalidPermsForProjectScopedPermissionsTesting)
 
-	permRequest := &cloudresourcemanager.TestIamPermissionsRequest{
-		Permissions: filteredPermList,
-	}
-	permResponse, err := gcpClient.TestIamPermissions(projectName, permRequest)
-	if err != nil {
-		return false, fmt.Errorf("error testing permissions: %v", err)
+	if len(filteredPermList) == 0 {
+		return true, nil
 	}
 
-	// check that each perm in our list is actually available
-	permMap := map[string]bool{}
-	for _, perm := range permResponse.Permissions {
-		permMap[perm] = true
-	}
-
-	disallowedPerms := sets.NewString()
-	for _, perm := range filteredPermList {
-		if _, ok := permMap[perm]; !ok {
-			disallowedPerms.Insert(perm)
+	// Split list to only check 100 permissions at a time
+	allowedPerms := sets.NewString()
+	chunkSize := 100
+	permLen := len(filteredPermList)
+	for i := 0; i < permLen; i += chunkSize {
+		end := i + chunkSize
+		if end > permLen {
+			end = permLen
 		}
+		req := &cloudresourcemanager.TestIamPermissionsRequest{Permissions: filteredPermList[i:end]}
+		resp, err := gcpClient.TestIamPermissions(projectName, req)
+		if err != nil {
+			return false, fmt.Errorf("error testing permissions: %v", err)
+		}
+		allowedPerms.Insert(resp.Permissions...)
 	}
 
-	if len(disallowedPerms) > 0 {
-		logger.Warnf("Detected some unallowed permissions: %s", disallowedPerms.List())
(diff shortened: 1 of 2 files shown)


csrwng:
@dgoodwin ptal

dgoodwin:
Is it feasible to gets some unit tests around this? It looks like a good spot for an edge case  bug to slip in.

csrwng:
> Is it feasible to gets some unit tests around this? It looks like a good spot for an edge case bug to slip in.

Yup, let me look into adding a test

pkg/controller/utils/gcp/utils.go
```diff
@@ -151,18 +151,34 @@ func CheckPermissionsAgainstPermissionList(gcpClient ccgcp.Client, permList []st
...
+		permRequest := &cloudresourcemanager.TestIamPermissionsRequest{
+			Permissions: filteredPermList[index:upperLimit],
+		}
+		permResponse, err := gcpClient.TestIamPermissions(projectName, permRequest)
+		if err != nil {
+			return false, fmt.Errorf("error testing permissions: %v", err)
+		}
+		responses = append(responses, permResponse)
+		index += 100
 	}
```
ironcladlou: Just another possible approach for consideration since I found the problem interesting:

```go
var chunks [][]string
chunkSize := 100
permLen := len(filteredPermList)
for i := 0; i < permLen; i += chunkSize {
    end := i + chunkSize
    if end > permLen {
        end = permLen
    }
    chunks = append(chunks, filteredPermList[i:end])
}

for chunk := range chunks {
  // do request
}
```

csrwng: Thx! I used it for the test as an alternative way to come up with the chunks

csrwng:
@dgoodwin added a unit test

pkg/controller/utils/gcp/utils.go
```diff
@@ -147,22 +147,43 @@ func filterOutPermissions(permList, filterOutList []string) []string {
...
-	}
-	permResponse, err := gcpClient.TestIamPermissions(projectName, permRequest)
-	if err != nil {
-		return false, fmt.Errorf("error testing permissions: %v", err)
+	// Split list to only check 100 permissions at a time
+	responses := []*cloudresourcemanager.TestIamPermissionsResponse{}
+	done := false
+	index := 0
+	for !done {
+		upperLimit := len(filteredPermList)
```
wking: nit: I'd rather use @ironcladlou's [loop approach][1] here and just hard-code the expected chunking in the unit test.

```go
for _, testCase := range []struct{
  length int
  chunks []int
}{
  {length: 0},
  {length: 2, chunks: {2}},
  {length: 100, chunks: {100}},
  {length: 101, chunks: {100, 1}},
  {length: 230, chunks: {100, 100, 30}},
  {length: 300, chunks: {100, 100, 100}},
  {length: 379, chunks: {100, 100, 100, 79}},
} {
  t.Run(fmt.Sprintf("length=%d", testCase.length), ...)
}
```

[1]: https://github.com/openshift/cloud-credential-operator/pull/94#discussion_r307866955

pkg/controller/utils/gcp/utils.go
```diff
@@ -147,22 +147,43 @@ func filterOutPermissions(permList, filterOutList []string) []string {
...
+		}
+		responses = append(responses, permResponse)
+		index += 100
 	}
 
 	// check that each perm in our list is actually available
 	permMap := map[string]bool{}
-	for _, perm := range permResponse.Permissions {
-		permMap[perm] = true
+	for _, resp := range responses {
```
wking: nit: instead of collecting a slice of responses, I'd rather create `permMap` before the chunk loop and then iterate over each chunk's response inside the loop:

```go
permMap := map[string]bool{}
for i := 0; i < permLen; i += chunkSize {
  ...
  for _, perm := range resp.Permissions {
    permMap[perm] = true
  }
}
```

wking:
Left two minor nits inline, but this looks fine to me whether those get addressed or not.

csrwng:
> Left two minor nits inline, but this looks fine to me whether those get addressed or not.

Thx! addressed.

pkg/controller/utils/gcp/utils.go
```diff
@@ -147,37 +147,47 @@ func filterOutPermissions(permList, filterOutList []string) []string {
...
+	// Split list to only check 100 permissions at a time
+	var chunks [][]string
+	chunkSize := 100
+	permLen := len(filteredPermList)
+	for i := 0; i < permLen; i += chunkSize {
+		end := i + chunkSize
+		if end > permLen {
+			end = permLen
+		}
+		chunks = append(chunks, filteredPermList[i:end])
```
wking: You don't need to build a `chunks` slice; just make your request with `Permissions: filteredPermList[I:end]` in this same loop.

csrwng: Ack, fixed

PR #97, opened 2019-08-07
4.2 Upgrade: Don't assume platformStatus is present in Infra

csrwng:


pkg/controller/controller.go
```diff
@@ -80,18 +81,18 @@ func AddToManager(m manager.Manager) error {
 			}
 		case configv1.GCPPlatformType:
 			log.Info("initializing GCP actuator")
-			if plat.GCP == nil {
+			if infraStatus.PlatformStatus == nil || infraStatus.PlatformStatus.GCP == nil {
```
abhinavdahiya: for GCP platformstatus will always be set. since it's a 4.2 platform

csrwng: ack... which is why if it's not there we should just exit

abhinavdahiya: +1

PR #98, opened 2019-08-08
Bug 1707827: Enable leader election.

dgoodwin:
Requires granting RBAC for configmaps and events, otherwise quite
straightforward.

Also fixes an issue discovered where our RBAC generated files were not
being used, due to a mismatch in their names.

There have been intermittent reports of CCO reporting the old version after an upgrade. A separate bug exists where CCO has no level gate and thus failing to report the new version does not impact the upgrade, which is probably should.

However for this issue, we have never been able to reproduce the problem, but we know from logs that something was setting it back to the old value despite the new pod being running. We reasoned this likely to be caused by a lack of leader election, when the new pod comes up the old is still running and if the timing is right, the old can reset the value to the old. The situation does not break the upgrade due to the missing level gate, and the problem apparently corrects itself as the old pod is scaled down and the new pod eventually reconciles and updates the version to the new again.

Despite never being able to reproduce, both of these fixes have been present in 4.2 for some time. We believe the issue should be solved.

This PR only backports the leader election, I am unsure if we should also immediately take the level gate change as well at the risk of breaking upgrades if the leader election does not turn out to solve the problem.

CC @derekwaynecarr


dgoodwin:
/hold

Not sure if the lgtm will auto-merge as this is for 4.1 where z-patch manager will need to approve, Holding just in case until we hear from Derek.

wking:
> Not sure if the lgtm will auto-merge as this is for 4.1 where z-patch manager will need to approve...

It won't (although a hold doesn't hurt), because [you need `cherry-pick-approved`][1].  Basically, you can't accidentally merge into a release branch unless you're a patch manager ;).

[1]: https://github.com/openshift/release/blob/678c1092d62522db713640a8da6e383f315a86db/cluster/ci/config/prow/config.yaml#L234-L250

derekwaynecarr:
/hold cancel

there is no corresponding 4.2 BZ for this issue as it was hard to reproduce, but the linked code in 4.2 has been present for an extended period.  leader election on the operator is a good thing, and this is using the standard controller runtime library for it.  let's take this back.

PR #102, opened 2019-08-19
WIP: tag resource groups with created App Registration IDs

joelddiaz:
Azure Application Registrations exist outside of a Resource Group. In order to allow a mechanism for the OpenShift installer/uninstaller to more fully clean up a cluster during deprovisioning, track any created App Registrations by tagging the Resource Group containing the cluster.

The tag format is of the form key: [app reg UUID] with value: [display name of app reg].

Also, update test cases to more clearly indicate the expected Azure API calls for each test case.

NOTE: tagging resource groups requires that the credentials that cloud-cred-operator is running have the role Contributor.

### DIFF
--- a/pkg/azure/actuator.go
+++ b/pkg/azure/actuator.go
@@ -59,16 +59,16 @@ func NewActuator(c client.Client) (*Actuator, error) {
 
 	client := newClientWrapper(c)
 	return &Actuator{
-		client: client,
-		codec:  codec,
+		client:                  client,
+		codec:                   codec,
 		credentialMinterBuilder: NewAzureCredentialsMinter,
 	}, nil
 }
 
 func NewFakeActuator(c client.Client, codec *minterv1.ProviderCodec, credentialMinterBuilder credentialMinterBuilder) *Actuator {
 	return &Actuator{
-		client: newClientWrapper(c),
-		codec:  codec,
+		client:                  newClientWrapper(c),
+		codec:                   codec,
 		credentialMinterBuilder: credentialMinterBuilder,
 	}
 }
@@ -188,6 +188,15 @@ func (a *Actuator) Delete(ctx context.Context, cr *minterv1.CredentialsRequest) 
 	// Deleting AAD application results in deleting its service principal
 	// and all roles assigned
 	if err := azureCredentialsMinter.DeleteAADApplication(ctx, spName); err != nil {
+		return err
+	}
+
+	infraResourceGroups, err := loadAzureInfrastructureResourceGroups(a.client.Client, logger)
+	if err != nil {
+		return err
+	}
+
+	if err = azureCredentialsMinter.UntagResourceGroups(ctx, infraResourceGroups, azureStatus.AppID); err != nil {
 		return err
 	}
 
@@ -390,6 +399,11 @@ func (a *Actuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequest
 
 	azureStatus.AppID = *aadApp.AppID
 	err = a.updateProviderStatus(ctx, logger, cr, azureStatus)
+	if err != nil {
+		return err
+	}
+
+	err = azureCredentialsMinter.TagResourceGroups(ctx, infraResourceGroups, *aadApp.AppID, *aadApp.DisplayName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


joelddiaz:
@abhinavdahiya what do you think of this proposed tagging scheme to keep track of the Application Registrations the cloud-cred-operator creates outside of the cluster's Resource Group?

We would of course then need to plug into the installer's uninstall path to read these tags and delete the App Registrations. (Should the tagging scheme allow for objects besides App Registrations?)

joelddiaz:
@ingvagabund since you're most familiar with the azure part of the repo, PTAL

abhinavdahiya:
> In order to allow a mechanism for the OpenShift installer/uninstaller to more fully clean up a cluster during deprovisioning, track any created App Registrations by tagging the Resource Group containing the cluster.

In my opinion, this can be an optimization but not **the** solution. Also since we are tagging the resource group for application ID generated, we should make sure we clean it up when the application is deleted due to say delete credential request.. Make sure the lifecycle of the tag is maintained.

joelddiaz:
> In my opinion, this can be an optimization but not **the** solution. Also since we are tagging the resource group for application ID generated, we should make sure we clean it up when the application is deleted due to say delete credential request.. Make sure the lifecycle of the tag is maintained.

When someone deletes a CredentialsRequest on a running cluster, the cloud-cred-operator will clean up the Azure resources (and remove the tag on the Resource Group). But when someone runs 'openshift-install destroy cluster', the cloud-cred-operator doesn't get a chance to clean up, so we would leak the App Registration(s). So the tags are to give openshfit-(un)install a chance to clean up.
What do you have in mind for a full solution?

abhinavdahiya:
> > In my opinion, this can be an optimization but not **the** solution. Also since we are tagging the resource group for application ID generated, we should make sure we clean it up when the application is deleted due to say delete credential request.. Make sure the lifecycle of the tag is maintained.
> 
> When someone deletes a CredentialsRequest on a running cluster, the cloud-cred-operator will clean up the Azure resources (and remove the tag on the Resource Group). But when someone runs 'openshift-install destroy cluster', the cloud-cred-operator doesn't get a chance to clean up, so we would leak the App Registration(s). So the tags are to give openshfit-(un)install a chance to clean up.
> What do you have in mind for a full solution?

since app creation and tagging the resource group are 2 steps.. there will be cases where the destroy leaks an application.

What is the problem in listing applications for the cluster directly? based on tags or name prefix or even ownedBy relationships...

ingvagabund:
> What is the problem in listing applications for the cluster directly? based on tags or name prefix or even ownedBy relationships...

application resource can not be tagged, only service principals can be.

joelddiaz:
> since app creation and tagging the resource group are 2 steps.. there will be cases where the destroy leaks an application.
> 
Just like AWS today ;)

> What is the problem in listing applications for the cluster directly? based on tags or name prefix or even ownedBy relationships...

The naming of the App Registrations already captures the cluster name today:

```
[jdiaz@minigoomba bin (master $%=)]$ az ad app list --display-name "jdiaz-az-4nm6w" | jq -r '.[].displayName'
jdiaz-az-4nm6w-openshift-image-registry-azure
jdiaz-az-4nm6w-openshift-ingress-azure
jdiaz-az-4nm6w-openshift-machine-api-azure
```
@abhinavdahiya in your opinion, is it enough to just find App Registrations with that naming prefix during 'openshift-install destroy cluster' (forget this PR with the tagging)?

abhinavdahiya:
```console
az ad sp list --show-mine -otable
AccountEnabled    AppDisplayName     AppId                                 AppOwnerTenantId                      AppRoleAssignmentRequired    DisplayName        Homepage                   ObjectId                              ObjectType        Odata.type                                    PublisherName    ServicePrincipalType    SignInAudience
----------------  -----------------  ------------------------------------  ------------------------------------  ---------------------------  -----------------  -------------------------  ------------------------------------  ----------------  --------------------------------------------  ---------------  ----------------------  ----------------
```

listing service principal provide the AppID too.. why not delete service principal for the cluster and delete the corresponding application??

joelddiaz:
> listing service principal provide the AppID too.. why not delete service principal for the cluster and delete the corresponding application??

So tag the service principal then cluster destroy looks like:
1) find service principal by tag
2) delete app registration associated with SP
3) delete SP
?

joelddiaz:
closing as we went with #104 instead

PR #103, opened 2019-08-19
Bug 1742753: clear other conditions when ignoring a cred request

joelddiaz:
don't allow a credentialsRequest that is being ignored to put cloud-cred-operator in a degraded state.

clear other conditions when marking a credentialsRequest as ignored.


sdodson:
/retitle Bug 1742753: clear other conditions when ignoring a cred request

PR #104, opened 2019-08-20
tag Service Principals on create

joelddiaz:
Azure Application Registrations exist outside of a Resource Group and are not taggable. In order to allow a mechanism for the OpenShift installer/uninstaller to more fully clean up a cluster during deprovisioning, track any created App Registrations by tagging the App Registration's Service Principal (which is taggable). The App Registration ID (needed for full cleanup), can be found in the Service Principal's 'appId' field.

Service Principals tags are just strings (not key=value pairs), so tag the Service Principal with 'kubernetes.io_cluster.[cluster infra name]=owned' (to closely follow the pattern where key, values are supported).

Also, update test cases to more clearly indicate the expected Azure API calls for each test case.

### DIFF
--- a/pkg/azure/minter.go
+++ b/pkg/azure/minter.go
@@ -126,8 +126,7 @@ func (credMinter *AzureCredentialsMinter) CreateOrUpdateAADApplication(ctx conte
 
 // CreateOrGetServicePrincipal creates a new SP and returns it.
 // Service principal that already exist is returned.
-
-func (credMinter *AzureCredentialsMinter) CreateOrGetServicePrincipal(ctx context.Context, appID string) (*graphrbac.ServicePrincipal, error) {
+func (credMinter *AzureCredentialsMinter) CreateOrGetServicePrincipal(ctx context.Context, appID, infraName string) (*graphrbac.ServicePrincipal, error) {
 	spItems, err := credMinter.spClient.List(ctx, fmt.Sprintf("appId eq '%v'", appID))
 	if err != nil {
 		return nil, err
@@ -137,10 +136,13 @@ func (credMinter *AzureCredentialsMinter) CreateOrGetServicePrincipal(ctx contex
 	case 0:
 		credMinter.logger.Infof("Creating service principal for AAD application %q", appID)
 		var servicePrincipal *graphrbac.ServicePrincipal
+		ownedTag := fmt.Sprintf("kubernetes.io_cluster.%s=owned", infraName)
+
 		err := wait.PollImmediate(5*time.Second, 60*time.Second, func() (bool, error) {
 			sp, err := credMinter.spClient.Create(ctx, graphrbac.ServicePrincipalCreateParameters{
 				AppID:          to.StringPtr(appID),
 				AccountEnabled: to.BoolPtr(true),
+				Tags:           &[]string{ownedTag},
 			})
 			// ugh: Azure client library doesn't have the types registered to
 			// unmarshal all the way down to this error code natively :-(
--- a/pkg/controller/credentialsrequest/credentialsrequest_controller.go
+++ b/pkg/controller/credentialsrequest/credentialsrequest_controller.go
@@ -255,7 +255,7 @@ type ReconcileCredentialsRequest struct {
 // +kubebuilder:rbac:groups=core,resources=secrets;configmaps;events,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch
 // +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
-// +kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures,verbs=get;list;watch
+// +kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures;dnses,verbs=get;list;watch
 // +kubebuilder:rbac:groups=config.openshift.io,resources=clusteroperators;clusteroperators/status,verbs=create;get;update;list;watch
 func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	logger := log.WithFields(log.Fields{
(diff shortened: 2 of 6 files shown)


joelddiaz:
@abhinavdahiya this look more reasonable as a tagging strategy (as opposed to tagging the resource groups)?

This should allow the 'openshift-install destroy cluster' to just get all Service Principals with the appropriate tag, and delete the Application Registrations/Service Principals.

joelddiaz:
@ingvagabund here's an alternative way to handle tagging things

abhinavdahiya:
`'kubernetes.io_cluster.[cluster infra name]=owned` personally `kubernetes_io.cluster.<infra-id>` infra id can have `_-` but not `.`

ingvagabund:
@joelddiaz it will not discovered apps for which there is no SP. It might happen CCO will be able to create aad apps but fail to create SP. In which case you will still leak some apps.

Though, I like this approach more.

joelddiaz:
> @joelddiaz it will not discovered apps for which there is no SP. It might happen CCO will be able to create aad apps but fail to create SP. In which case you will still leak some apps.
> 
Agreed, there is still the window where we can leak, but to me this is no worse than in AWS where tagging is a separate step than creating an object. And I much prefer finding things by tags than by naming convention (especially since Azure allows two different objects to be named identically).

> Though, I like this approach more.

joelddiaz:
> `'kubernetes.io_cluster.[cluster infra name]=owned` personally `kubernetes_io.cluster.<infra-id>` infra id can have `_-` but not `.`

@abhinavdahiya not sure what you're indicating in this comment. Are you concerned that the characters allowed for infra-id might conflict with what is allowed in a Service Principal tag field?

pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -255,7 +255,7 @@ type ReconcileCredentialsRequest struct {
 // +kubebuilder:rbac:groups=core,resources=secrets;configmaps;events,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch
 // +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
-// +kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures,verbs=get;list;watch
+// +kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures;dnses,verbs=get;list;watch
```
csrwng: We should get rid of these comments, since they’re no longer being used to generate rbac and changing them has no effect on anything.

csrwng:
> @abhinavdahiya not sure what you're indicating in this comment. Are you concerned that the characters allowed for infra-id might conflict with what is allowed in a Service Principal tag field?

I think this is just a comment on the text of your PR description

pkg/controller/credentialsrequest/credentialsrequest_controller.go
joelddiaz: the CCO repo still generates them. i added 'dnses' b/c this PR missed adding it https://github.com/openshift/cloud-credential-operator/pull/105, and my local builds were backing out the manual addition of 'dnses' from that PR.

csrwng: > the CCO repo still generates them

Not sure what this means? :) But afaik the only thing that adds these is kubebuilder when you add a new type/controller, but it adds it to the generated file, not to existing files. My concern is that going forward we may get the idea that adding permissions here will be enough without realizing that we really need to add them [here](https://github.com/openshift/cloud-credential-operator/blob/master/manifests/01_deployment.yaml#L15)

joelddiaz: I mean that the Makefile in the CCO repo still auto-generates the manifests/01_deployment.yaml.

I think we're both worried about the same thing :). Right now in the master branch if you run 'make' you end up with this diff
```
[jdiaz@minigoomba cloud-credential-operator (master *$%=)]$ git diff manifests/01_deployment.yaml
diff --git a/manifests/01_deployment.yaml b/manifests/01_deployment.yaml
index e278a4e5..2c31d450 100644
--- a/manifests/01_deployment.yaml
+++ b/manifests/01_deployment.yaml
@@ -62,7 +62,6 @@ rules:
   - config.openshift.io
   resources:
   - infrastructures
-  - dnses
   verbs:
   - get
   - list
```

In order to not lose the 'dnses' entry that's already in the master branch of manifests/01_deployment.yaml, I added 'dnses' so that when 'make manifests' runs it keeps 'dnses' in the generated 01_deployment.yaml.

pkg/azure/minter.go
```diff
@@ -137,10 +136,13 @@ func (credMinter *AzureCredentialsMinter) CreateOrGetServicePrincipal(ctx contex
 	case 0:
 		credMinter.logger.Infof("Creating service principal for AAD application %q", appID)
 		var servicePrincipal *graphrbac.ServicePrincipal
+		ownedTag := fmt.Sprintf("kubernetes.io_cluster.%s=owned", infraName)
```
abhinavdahiya: the separation in this tag is wierd.. `kubernetes.io` and `cluster` are separate by `_` while the `cluster` and `infraname` are separated by `.` ??

pkg/controller/credentialsrequest/credentialsrequest_controller.go
csrwng: Sorry just looked at the Makefile, I see what you mean. For some reason I thought we’d abandoned the rbac generated from the code.

pkg/azure/minter.go
joelddiaz: @abhinavdahiya i was just trying to follow the existing tagging i see on Azure resources after install:
```
[jdiaz@minigoomba ~]$ az group show --name jdiaz-az-48vvg-rg | jq -r .tags
{
  "kubernetes.io_cluster.jdiaz-az-48vvg": "owned"
}
```

csrwng: I vote for consistency over correctness :) https://github.com/openshift/installer/blob/master/data/data/azure/main.tf#L4

PR #105, opened 2019-08-20
Bug 1743728: pkg/azure: also provide operators access to the resource group containing public DNS zone

abhinavdahiya:
cluster-ingress-operator updates the records in the public DNS zone that is used by the cluster, hence separate from the resource group created for the cluster.
The DNS or dnses.config.openshift.io/v1 object's `.spec.publicZone` is optional, and when set the, resource-group from the ID will be used.

The current api doesn't allow scoping permissions to a single resource, therefore, the permissions are provided to the entire resource group.
Also adds a resourceid helper to find the resource group for the public zone ID.

### DIFF
--- a/pkg/azure/actuator.go
+++ b/pkg/azure/actuator.go
@@ -273,6 +273,22 @@ func loadAzureInfrastructureResourceGroups(c client.Client, logger log.FieldLogg
 
 	resourceGroups := []string{infra.Status.PlatformStatus.Azure.ResourceGroupName}
 
+	dns := &configv1.DNS{}
+	err = c.Get(context.Background(), types.NamespacedName{Name: "cluster"}, dns)
+	if err != nil {
+		logger.Error("error loading DNS config 'cluster'")
+		return nil, err
+	}
+
+	if pzone := dns.Spec.PublicZone; pzone != nil {
+		id, err := parseAzureResourceID(pzone.ID)
+		if err != nil {
+			logger.Error("failed to parse ID for public zone")
+			return nil, err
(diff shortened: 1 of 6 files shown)


pkg/azure/actuator.go
```diff
@@ -273,6 +273,22 @@ func loadAzureInfrastructureResourceGroups(c client.Client, logger log.FieldLogg
...
 
+	dns := &configv1.DNS{}
+	err = c.Get(context.Background(), types.NamespacedName{Name: "cluster"}, dns)
+	if err != nil {
+		logger.Error("error loading DNS config 'cluster'")
+		return nil, err
+	}
+
+	if pzone := dns.Spec.PublicZone; pzone != nil {
+		id, err := parseAzureResourceID(pzone.ID)
```
ingvagabund: +1 for implementing `parseAzureResourceID`.

pkg/azure/actuator.go
```diff
@@ -273,6 +273,22 @@ func loadAzureInfrastructureResourceGroups(c client.Client, logger log.FieldLogg
...
+		if err != nil {
+			logger.Error("failed to parse ID for public zone")
+			return nil, err
+		}
+		resourceGroups = append(resourceGroups, id.ResourceGroup)
```
ingvagabund: Only `ResourceGroup` field of `resourceID` is actually accessed. Though, there are few other places in the code where `parseAzureResourceID` might be used as well. Not just to access high level keys. E.g. to parse `RoleDefinitionID`.

PR #106, opened 2019-08-21
status: Add CredentialsRequests to relatedObjects

Miciah:
Add all CredentialRequests resources to `relatedObjects` so that [must-gather](https://github.com/openshift/must-gather) will gather them.

* `pkg/controller/credentialsrequest/status.go` (`syncOperatorStatus`): Pass the list of CredentialRequests to `buildExpectedRelatedObjects`.
(`buildExpectedRelatedObjects`): Add a parameter for a slice of CredentialRequests.  Add the provided CredentialRequests resources the list of related objects.

### DIFF
--- a/pkg/controller/credentialsrequest/status.go
+++ b/pkg/controller/credentialsrequest/status.go
@@ -64,7 +64,7 @@ func (r *ReconcileCredentialsRequest) syncOperatorStatus() error {
 	oldRelatedObjects := co.Status.RelatedObjects
 	co.Status.Conditions = computeStatusConditions(oldConditions, credRequests, r.platformType)
 	co.Status.Versions = computeClusterOperatorVersions()
-	co.Status.RelatedObjects = buildExpectedRelatedObjects()
+	co.Status.RelatedObjects = buildExpectedRelatedObjects(credRequests)
 
 	// ClusterOperator should already exist (from the manifests payload), but recreate it if needed
 	if isNotFound {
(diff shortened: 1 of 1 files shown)


pkg/controller/credentialsrequest/status.go
```diff
@@ -276,11 +276,20 @@ func findClusterOperatorCondition(conditions []configv1.ClusterOperatorStatusCon
...
+	related := []configv1.ObjectReference{
 		{
 			Resource: "namespaces",
 			Name:     cloudCredOperatorNamespace,
 		},
 	}
+	for _, cr := range credRequests {
+		related = append(related, configv1.ObjectReference{
+			Group:     minterv1.SchemeGroupVersion.Group,
+			Resource:  "CredentialsRequest",
```
abhinavdahiya: the resource is wrong, this should be `credentialsrequests`

abhinavdahiya: https://github.com/openshift/cloud-credential-operator/blob/77eaf88b08645f1953b657bf0109aab15054c50b/manifests/00_v1_crd.yaml#L12

Miciah: How is the `plural` field more correct to use here than the `kind` field?  Is there an API that using kind violates, or a tool that it breaks?  For what it's worth, I tested this change and verified that must-gather gathered the CredentialsRequest resources.  Morever, use the kind (upper camel case) in the ingress operator too have not seen or received reports of any problems.

PR #107, opened 2019-08-21
[gcp] set timeout on cloud API calls

csrwng:
Adds a timeout to calls to the GCP cloud API so that network errors do not freeze the operator.

### DIFF
--- a/pkg/gcp/client.go
+++ b/pkg/gcp/client.go
@@ -19,6 +19,7 @@ package gcp
 import (
 	"context"
 	"fmt"
+	"time"
 
 	// GCP auth
 	"golang.org/x/oauth2/google"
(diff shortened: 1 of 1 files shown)


pkg/gcp/client.go
```diff
@@ -64,50 +65,77 @@ type gcpClient struct {
...
 
 func (c *gcpClient) ListServiceAccountKeys(ctx context.Context, request *iamadminpb.ListServiceAccountKeysRequest) (*iamadminpb.ListServiceAccountKeysResponse, error) {
+	ctx, cancel := contextWithTimeout(ctx)
+	defer cancel()
 	return c.iamClient.ListServiceAccountKeys(ctx, request)
 }
 
 func (c *gcpClient) GetProjectIamPolicy(projectName string, request *cloudresourcemanager.GetIamPolicyRequest) (*cloudresourcemanager.Policy, error) {
-	return c.cloudResourceManagerClient.Projects.GetIamPolicy(projectName, request).Do()
+	ctx, cancel := contextWithTimeout(context.Background())
```
abhinavdahiya: Background should only be called in main.. everything else should either get the context or use TODO.
see https://golang.org/pkg/context/#Background and https://golang.org/pkg/context/#TODO

csrwng:
Switched calls from Background to TODO(). Eventually there should just be one context that is created by the controller and passed around to all code that needs to make API calls. We can do that post 4.2

joelddiaz:
trying to get gcp e2e
/close

csrwng:
This is weird, the gcp-e2e says it passed (which is awesome), but if you look at the detail, 3 tests failed. Hmm

PR #108, opened 2019-08-27
allow 32 char infraID in generated Service Principal names

joelddiaz:
Azure allows a 93 character service principal name. Preserve the entire 32 character infraID (max length from the installer) and leave 54 characters for the credentialsRequest name, and a final 5 chars for randomness (all fields separated by dashes).

### DIFF
--- a/pkg/azure/actuator.go
+++ b/pkg/azure/actuator.go
@@ -43,11 +43,16 @@ import (
 
 var _ actuator.Actuator = (*Actuator)(nil)
 
+type servicePrincipalNameBuilder func(string, string) (string, error)
+
 // Actuator implements the CredentialsRequest Actuator interface to create credentials for Azure.
 type Actuator struct {
 	client                  *clientWrapper
 	codec                   *minterv1.ProviderCodec
 	credentialMinterBuilder credentialMinterBuilder
+
+	// Allow mocking the random string generation
+	generateServicePrincipalName servicePrincipalNameBuilder
 }
 
 func NewActuator(c client.Client) (*Actuator, error) {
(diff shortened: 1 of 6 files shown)


pkg/controller/utils/utils.go
```diff
@@ -76,3 +76,25 @@ func GetCredentialsRequestCloudType(providerSpec *runtime.RawExtension) (string,
...
+
+	infraPrefix := ""
+	if infraName != "" {
+		if len(infraName) > infraNameMaxLen {
+			infraName = infraName[0:infraNameMaxLen]
+		}
+		infraPrefix = infraName + "-"
+	}
+	if len(crName) > crNameLen {
+		crName = crName[0:crNameLen]
```
ingvagabund: I would still error when the entire generated name is longer than 93 instead of truncating the string. One might create two requests where both request name has first 60 characters in common. In which case both requests will get the same service principal. And if both requests have different list of roles, one request will override roles of another request.

ingvagabund: In that case we will first need to update CCR status [1] and set the generated service principal name there [2] before the AAD application is created.

[1] https://github.com/openshift/cloud-credential-operator/blob/master/pkg/azure/actuator.go#L426-L429
[2] https://github.com/openshift/cloud-credential-operator/blob/master/pkg/azure/actuator.go#L372

ingvagabund: > rather than save the generated names of things, couldn't we just find the App Registration and the Service Principal by their appID fields? we already save the appID in the CCR status.

There is a risk you will be able to create AAD application but fail to update CCR status and set the appID there. In which case you will generate different app name and will have orphaned app application resource. The resource will have no roles assigned, no secret generated so it will cause no harm. As long as that's fine, even this makes sense.

pkg/azure/actuator.go
```diff
@@ -363,11 +372,19 @@ func (a *Actuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequest
...
+		spName, err = a.generateServicePrincipalName(infraName, cr.Name)
 		if err != nil {
 			return err
 		}
 		azureStatus.ServicePrincipalName = spName
+
+		// Save status immediately in case we successfully create the AppRegistration and
+		// then fail to update credReq status causing us to leak the created AppRegistration
+		// because we can't find it by name anymore.
+		err = a.updateProviderStatus(ctx, logger, cr, azureStatus)
```
ingvagabund: Even `updateProviderStatus` can error in which case the created AAD application will be orphaned and new AAD application will be created.

PR #109, opened 2019-09-04
handle the posibilitiy of paginated results from SimulatePrincipalPolicy

joelddiaz:
- [x] add and use SimulatePrincipalPolicyPages() to the AWS client so that we can handle a multi-paged response from AWS
- [x] update test cases to mock up the SimulatePrincipalPolicyPages() responses

### DIFF
--- a/pkg/controller/secretannotator/aws/reconciler_test.go
+++ b/pkg/controller/secretannotator/aws/reconciler_test.go
@@ -57,6 +57,17 @@ const (
 	testAWSSecretAccessKey = "KEEPITSECRET"
 )
 
+var (
+	failedSimulationResponse = &iam.SimulatePolicyResponse{
+		EvaluationResults: []*iam.EvaluationResult{
+			{
+				EvalDecision:   aws.String("notallowed"),
+				EvalActionName: aws.String("SomeAWSAction"),
+			},
+		},
+	}
+)
+
 func init() {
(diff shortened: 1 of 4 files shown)


pkg/controller/secretannotator/aws/reconciler_test.go
```diff
@@ -232,41 +243,27 @@ func mockGetUser(mockAWSClient *mockaws.MockClient) {
...
-	mockAWSClient.EXPECT().SimulatePrincipalPolicy(gomock.Any()).Return(&iam.SimulatePolicyResponse{
-		EvaluationResults: []*iam.EvaluationResult{
-			{
-				EvalDecision:   aws.String("notallowed"),
-				EvalActionName: aws.String("SomeAWSAction"),
-			},
-		},
-	}, nil)
+	mockAWSClient.EXPECT().SimulatePrincipalPolicyPages(gomock.Any(), gomock.Any()).Return(nil).
+		// Now in the Do() receive the lambda function f() so we can send it the failed result
```
staebler: We should have a test case that verifies the behavior when the lambda function is supplied passing input as well.

joelddiaz: yes, right now it relies on the fact that an empty response means that the code assumes everything worked :/

PR #110, opened 2019-09-05
use a more realistic mock setup for the success cases

joelddiaz:


joelddiaz:
@staebler here's a more realistic looking happy path test setup

PR #111, opened 2019-09-09
Bug 1747366: Allow use of global proxy

joelddiaz:
create a configmap that will be populated with the cluster's certificate bundle.
mark that configmap as an optional volume for the CCO pod.

TODO: watch the configmap for changes and restart COO as necessary.

### DIFF
--- a/manifests/01_deployment.yaml
+++ b/manifests/01_deployment.yaml
@@ -112,6 +112,14 @@ subjects:
   namespace: openshift-cloud-credential-operator
 ---
 apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    config.openshift.io/inject-trusted-cabundle: "true"
+  name: cco-trusted-ca
+  namespace: openshift-cloud-credential-operator
+---
+apiVersion: v1
 kind: Service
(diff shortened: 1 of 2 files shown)


manifests/01_deployment.yaml
```diff
@@ -167,6 +169,9 @@ spec:
             cpu: 10m
             memory: 150Mi
         terminationMessagePolicy: FallbackToLogsOnError
+        volumeMounts:
+        - name: cco-trusted-ca
+          mountPath: /etc/pki/ca-trust/extracted/pem/
```
dgoodwin: Does this behave ok when there's no proxy in play?

joelddiaz: supposedly. copied this from https://github.com/operator-framework/operator-marketplace/pull/245/files#diff-5e5f804a8ee2fb8c81711b728eb5a21dR76

joelddiaz: also why i pushed this PR. to see how CI handle this

dgoodwin:
@joelddiaz will remove hold pending one more confirmation this is working.

PR #113, opened 2019-09-13
watch for changes in trusted CA configmap

joelddiaz:
in order to have the pod run with an up-to-date set of trusted CAs, watch the configmap containing the pods's CAs, and if the contents change, exit the container, so that we get restarted with the new CA content.

### DIFF
--- a/pkg/controller/configmap/configmap_controller.go
+++ b/pkg/controller/configmap/configmap_controller.go
@@ -1,0 +1,136 @@
+/*
+Copyright 2019 The OpenShift Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package configmap
+
+import (
+	"context"
+	"crypto/md5"
+	"fmt"
+	"os"
+
+	log "github.com/sirupsen/logrus"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	controllerName   = "configmap"
+	configMapName    = "cco-trusted-ca"
+	configMapKeyName = "ca-bundle.crt"
+	ccoNamespace     = "openshift-cloud-credential-operator"
+)
+
+// Add creates a new ConfigMap controller and adds it to the manager.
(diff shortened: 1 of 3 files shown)


pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,129 @@
...
+	return cm.GetName() == configMapName && cm.GetNamespace() == ccoNamespace
+}
+
+var _ reconcile.Reconciler = &ReconcileConfigMap{}
+
+// ReconcileConfigMap reconciles a ConfigMap against the on-disk contents.
+type ReconcileConfigMap struct {
+	client.Client
+	logger            log.FieldLogger
+	configMapDataHash string
```
dgoodwin: This is persistent state across reconcile loops isn't it. That is interesting, we don't have a lot of examples of that. It looks risky for concurrency, but then we're only reconciling on one single config map. Looks like it should be ok to me.

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,129 @@
...
+	cm := &corev1.ConfigMap{}
+	if err := r.Get(context.TODO(), request.NamespacedName, cm); err != nil {
+		r.logger.WithError(err).Error("failed to read in configmap")
+		return reconcile.Result{}, err
+	}
+
+	cmHash := fmt.Sprintf("%x", md5.Sum([]byte(cm.Data[configMapKeyName])))
+
+	if r.configMapDataHash == "" {
+		r.logger.Info("Saving hash of configmap containing CA data")
```
dgoodwin: When initializing r.logger lets add the configmap namespace/name, and the key as context fields.

Maybe "saving hash of proxy CA configmap".

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,129 @@
...
+		r.configMapDataHash = cmHash
+	} else if r.configMapDataHash == cmHash {
+		r.logger.Debug("no change in configmap detected")
```
dgoodwin: "no change in proxy CA configmap detected"

pkg/controller/configmap/configmap_controller.go
```diff
@@ -0,0 +1,129 @@
...
+	} else {
+		r.logger.Info("Configmap change detected, restarting pod")
```
dgoodwin: "proxy CA configmap change detected, restarting pod"

pkg/controller/configmap/configmap_controller.go
joelddiaz: Yes, persistent across reconcile calls. And as you say, since we're really only watching one single configmap (openshift-cloud-credential-operator/cco-trusted-ca), this should be safe.

wking:
> in order to have the pod run with an up-to-date set of trusted CAs, watch the configmap containing the pods's CAs, and if the contents change, exit the container, so that we get restarted with the new CA content.

Do we need this?  From [here][1]:

> When a ConfigMap already being consumed in a volume is updated, projected keys are eventually updated as well. Kubelet is checking whether the mounted ConfigMap is fresh on every periodic sync. However, it is using its local ttl-based cache for getting the current value of the ConfigMap. As a result, the total delay from the moment when the ConfigMap is updated to the moment when new keys are projected to the pod can be as long as kubelet sync period (1 minute by default) + ttl of ConfigMaps cache (1 minute by default) in kubelet.

[1]: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically

joelddiaz:
> Do we need this? 

@wking If memory serves me correctly, the way I understand this is that the configmap changes eventually make it into the container, but the database of trusted CAs wouldn't be reflected in the controller without rebuilding the trusted CA database.
If there's a better way, I'm happy to hear it.

wking:
> ...but the database of trusted CAs wouldn't be reflected in the controller...

Ah, yeah, turns out that Go caches the system store [on first load][1] and then never reloads the system pool.  I'll see if I can get some motion on that upstream.

[1]: https://github.com/golang/go/blob/f2e51f00158c2dcdff37c573c24f798d1e63db31/src/crypto/x509/cert_pool.go#L53-L54

PR #115, opened 2019-09-30
render manifests for use as a bootstrap pod

joelddiaz:
some cluster components would require early minting of credentials. this requires CCO to run as a bootstrap static pod to satisfy the early credentialsRequests.
    
load the various objects that go into the manifests directory of this repo into binary assets.
    
break out some of the objects CCO creates in the manifests dir of this repo so that we can selectively fetch them at runtime.
    
add a 'render' subcommand to the CCO binary that can be called from the installer's bootstrap node.
it will create the various etcd objects needed for the CCO to be able to run early during cluster bringup (namespace, service, etc).
it will also create a pod object that will be brought up as a static pod on the bootstrap node. this will look somewhat different than the pod spec in the regular Deployment for CCO as we need to load the bootstrap node's kubeconfig into the CCO pod.
    
update the controller initializations for secretannotator and credentialsrequest so that they can use/respect the '--kubeconfig' command line parameter passed to the controller manager.


joelddiaz:
the work in this PR comes out of the request/discussion in https://github.com/openshift/cluster-network-operator/pull/211

pkg/controller/platform/platform.go
```diff
@@ -35,8 +36,9 @@ func GetType(infraStatus *configv1.InfrastructureStatus) configv1.PlatformType {
 	return infraStatus.Platform
 }
 
-func getClient() (client.Client, error) {
+func getClient(explicitKubeconfig string) (client.Client, error) {
```
dgoodwin: Should this handle empty string? It seems like this flag won't always be set.

cmd/manager/main.go
```diff
@@ -141,6 +142,7 @@ func (writer glogWriter) Write(data []byte) (n int, err error) {
 func main() {
 	defer glog.Flush()
 	cmd := NewRootCommand()
+	cmd.AddCommand(newRenderCommand())
```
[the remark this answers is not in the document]
joelddiaz: yep, fixed with latest push to have it 'go run ./cmd/manager --log-level=debug'

pkg/controller/platform/platform.go
joelddiaz: a few lines down where we
```rules.ExplicitPath = explicitKubeconfig```
will take the empty string and reassign the empty string to rules.ExplicitPath (which is already the default setting). so in that case passing the empty string to getClient() gets us back the old behavior of not setting ExplicitPath

dgoodwin: Sounds good. thanks.

PR #117, opened 2019-10-04
Bug 1758666: copy trust bundle when actually defined

joelddiaz:
Dependent on the speed at which other components are restarted during a 4.1 to 4.2 upgrade, the CCO could be upgraded to a version that expects a trust bundle before it's fulfilled all credential requests thus causing the upgrade to deadlock because the CNO has not yet populated the trust bundle. Therefore, only utilize the trust bundle when it exists.

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -76,14 +76,18 @@ spec:
         - containerPort: 9876
           name: webhook-server
           protocol: TCP
-        command:
-          - /root/manager
-          - --log-level
-          - debug
+        command: ["/bin/bash", "-ec"]
+        args:
+        - |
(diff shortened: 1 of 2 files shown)


dgoodwin:
TLS errors...

/retest

sdodson:
/retitle Bug 1758666: copy trust bundle when actually defined

sdodson:
After reviewing @wking deep dive into the problem starting at comment 2 here https://bugzilla.redhat.com/show_bug.cgi?id=1758666#c2 this is certainly root cause and the raciness has been explained. Lets get the description of the PR updated with a short summary and this should be good to go. Just the normal PR title explaining what is done (this is fine already) and description explaining why it's being done.

PR #121, opened 2019-10-10
pkg/azure: add the network resource group to the list of infra groups

abhinavdahiya:
Currently the list of infra groups is `cluster` and `public dns zone`. And because the network resources like Virtual Network and Subnets are created for each cluster today this list is enough.
but to support Openshift clusters that use pre-existing Virtual Network and Subnets, we need the cluster to have access to this resource group too.

the network and cluster resource group split means all the cluster resources stay in the same group, but some network resources are used from a separate group. An example of operator that needs this access
is the machine-api-controllers, as they need to attach the nics for the virtual machine to subnets from network group.

The infratructure object provides the field for CloudConfig [1], which provides reference to the configmap and key that stores the k8s cloud provider config for Azure.
The cloud provider config for Azure contains the `vnetresourcegroupname` that is the network group for the cluster.

Using the cloud provider config to fetch the network resourcegroup name is as it is a well-known api.

Existing clusters have the cluster resource group as the network resource group, therefore no change in credentials and permissions is required. And only new clusters that are created with shared Virtual Network operator
will have the network group different from cluster group as install-time.

So even if the existing clusters have their permission updating capacity taken away, this will not create any failure during upgrades.

This shared Virtual Network work is being done in https://github.com/openshift/installer/pull/2441

[1]: https://github.com/openshift/api/blob/9c3d92fbdc669baf068d5503f2f1b1e5076d40d3/config/v1/types_infrastructure.go#L31

### DIFF
--- a/pkg/azure/actuator.go
+++ b/pkg/azure/actuator.go
@@ -18,6 +18,7 @@ package azure
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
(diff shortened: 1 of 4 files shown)


abhinavdahiya:
https://github.com/openshift/cloud-credential-operator/pull/122 is an alternate approach.

PR #122, opened 2019-10-10
 pkg/azure: add the network resource group to infra groups 

abhinavdahiya:
Currently the list of infra groups is `cluster` and `public dns zone`. And because the network resources like Virtual Network and Subnets are created for each cluster today this list is enough.
but to support Openshift clusters that use pre-existing Virtual Network and Subnets, we need the cluster to have access to this resource group too.

the network and cluster resource group split means all the cluster resources stay in the same group, but some network resources are used from a separate group. An example of operator that needs this access
is the machine-api-controllers, as they need to attach the nics for the virtual machine to subnets from network group.

The infrastructure object has the field NetworkResourceGroupName [1] which when set and different from the cluster group needs to be added to the list of infra groups to all cluster operators access to it.

Existing clusters have the cluster resource group as the network resource group, therefore no change in credentials and permissions is required. And only new clusters that are created with shared Virtual Network operator
will have the network group different from cluster group as install-time.

So even if the existing clusters have their permission updating capacity taken away, this will not create any failure during upgrades.

this depends https://github.com/openshift/api/pull/463 and is an alternative to https://github.com/openshift/cloud-credential-operator/pull/121

[1]: https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L142


joelddiaz:
i think i prefer this version over #121

wking:
If I'm reading this right, we're attempting to grant whatever creds on all resource groups we've assembled from the various locations, but we may only need some creds in some resource groups and other creds in other resource groups.  Does it make sense to allow for resource groups to be submitted on the cred request itself, so the machine-API operator can ask for things like "I want create-a-machine in machine-resource-group and attach-interface-to-subnet in network-resource-group"?

abhinavdahiya:
> If I'm reading this right, we're attempting to grant whatever creds on all resource groups we've assembled from the various locations, but we may only need some creds in some resource groups and other creds in other resource groups. 

> Does it make sense to allow for resource groups to be submitted on the cred request itself, so the machine-API operator can ask for things like "I want create-a-machine in machine-resource-group and attach-interface-to-subnet in network-resource-group"?

your change would require, the credential requests to move from the release-image where they are easily auditable to resources created and manged by individual operator.
Also each operator would need access to create these resources in the cred-minter namespace, compared to today CVO manages these.
Also operator that themselves require the creds to talk to APIs, would either need another high level object that can mange it for it, or move to secret mount volume to fetching and watching the secret from the credential request.
Another benefit from keeping those in credential requests is that those have capavity to change only on upgrade and not in a running cluster, allowing users to keep the cred-minting powers away unless required during upgrades compared to in cluster at all times as operators would then need to be vetted to make sure the requests are not dynamic.

And do we really need fine-grained to each individual resource permissions, although nice to achieve, or the coarse-grained permissions are enough?

So I kind-of like the trade-off we have currently made.

abhinavdahiya:
This is ready for review.

@joelddiaz @twiest

PR #124, opened 2019-10-14
Bug 1750338: improve permissions simulation by adding region info

joelddiaz:
Allow providing an aws region to be used when performing the simulations.

IAM permissions can be defined in a way where permissions are granted as long as the region is set to some specific region(s). Add region info when simulating permissions testing so that we avoid false failures (ie we claim permission denied when it would actually work with an acceptable region set).

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -149,7 +149,7 @@ func (a *AWSActuator) Exists(ctx context.Context, cr *minterv1.CredentialsReques
 
 // needsUpdate will return whether the current credentials satisfy what's being requested
 // in the CredentialsRequest
-func (a *AWSActuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequest, infraName string) (bool, error) {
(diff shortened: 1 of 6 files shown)


dgoodwin:
Looks fine I would just expand on the commit message to clarify what it helps and what it doesn't. I assume adding this information is safe and doesn't require us to do both a simulate call with region, and a call without?

joelddiaz:
> Looks fine I would just expand on the commit message to clarify what it helps and what it doesn't. I assume adding this information is safe and doesn't require us to do both a simulate call with region, and a call without?

Right. We're just making the simulation more accurate when compared with the actual calls that we expect will be performed.

joelddiaz:
@dgoodwin i defined some test cases to ensure we are using the region as part of our permissions simulations (let me know if you want the commit squashed)

dgoodwin:
Commit title is a little generic, lets squash. Thanks!

sdodson:
/retitle Bug 1750338: improve permissions simulation by adding region info

PR #126, opened 2019-10-16
add metrics reporting the count of credReqs with conditions

joelddiaz:
should start to see these kinds of metrics being reported

```
cco_credentials_requests_conditions{condition="Ignored"} 8
cco_credentials_requests_conditions{condition="MissingTargetNamespace"} 1
```


joelddiaz:
@dgoodwin PTAL

PR #127, opened 2019-10-17
document metrics

joelddiaz:


Documentation/metrics.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Server
+
+These metrics describe the status of CCO. 
+
+All these metrics are prefixed with `cco_`
+
+| Name                            | Description                                        | Type      |
+|---------------------------------|----------------------------------------------------|-----------|
+| credentials_requests            | Count of credentials requests                      | Gauge     |
```
dgoodwin: Would it be ok to put the cco_ prefix on each name just for clarity and copying?

Documentation/metrics.md
```diff
@@ -0,0 +1,26 @@
...
+| controller_reconcile_seconds    | Controller execution time in reconcile loop.       | Histogram |
+| credentials_requests_conditions | Credentials requests with active conditions.       | Gauge     |
+
+`credentials_requests` counts the total number of CredentialRequest objects and labels them by their cloud/infrastructure.
+
+`controller_reconcile_seconds` measures each controller's (ie secretannotator, credentialsrequests, configmap, etc) execution time through it's respective reconcile loop.
+
+`credentials_requests_conditions` reports the number of CredentialsRequests with active conditions in their status (ie. Ignored, ProvisionFailed, MissingTargetNamespace, etc).
```
dgoodwin: Feels like these belong in the description field. Does the table render ok if we do that?

Documentation/metrics.md
```diff
@@ -0,0 +1,26 @@
...
+These metrics describe the status of CCO. 
+
+All these metrics are prefixed with `cco_`
+
+| Name                            | Description                                        | Type      |
+|---------------------------------|----------------------------------------------------|-----------|
+| credentials_requests            | Count of credentials requests                      | Gauge     |
+| controller_reconcile_seconds    | Controller execution time in reconcile loop.       | Histogram |
+| credentials_requests_conditions | Credentials requests with active conditions.       | Gauge     |
+
```
dgoodwin: It might also be good to call out the controller runtime metrics as well, and show some examples of what we get for free. (controller_runtime_reconcile_total most notably)

joelddiaz:
@dgoodwin i was working off the docs out in the etcd repo, but i agree throwing everything in the table cleans things up.
also added the info about the extra metrics with some examples.

joelddiaz:
/retest
@dgoodwin i think this is good for a re-review as i believe i addressed the feedback

PR #128, opened 2019-10-18
enable prometheus to scrape metrics

joelddiaz:
this involves:
- [x] labeling the namespace with openshift.io/cluster-monitoring: "true"
- [x] creating a Service to point to the pods metrics port (2112)
- [x] creating a ServiceMonitor to tell prometheus where to scrape
- [x] creating a Role that allows get/list/watch on endpoints/service/pods (for prometheus to use)
- [x] a RoleBinding to grant the prometheus ServiceAccount the permissions defined in the above Role

update the kustomize pieces to use these new files when generating manifests.


dgoodwin:
/lgtm

Nice work.

PR #129, opened 2019-10-22
add initial prometheus alerts

joelddiaz:
define two alerts in the PromtheusAlert object to fire when a CredentialsRequest has any of the following conditions:
MissingTargetNamespace for more than 5 minutes
ProvisioningFailed for more than 5 minutes


joelddiaz:
@dgoodwin FYI, all i saw available for the severity levels for alerts (by looking a running cluster) were either `warning` or `critical`, so I marked both these initial alerts as `warning`.

PR #131, opened 2019-10-24
in the event that the secret for a CR is missing, re-reconcile

joelddiaz:
fix the case where the credRequest has previously been provisioned, but then the secret holding the creds (cr.Spec.SecretRef) has been deleted.

detect this case, and cause a re-reconcile even if the credRequest has recently been synced

add/update test cases

### DIFF
--- a/pkg/controller/credentialsrequest/credentialsrequest_controller.go
+++ b/pkg/controller/credentialsrequest/credentialsrequest_controller.go
@@ -412,12 +412,28 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
 	} else {
 		logger.Debug("found secret namespace")
 		setMissingTargetNamespaceCondition(cr, false)
+
+	}
+
+	// Check if the secret the credRequest wants created already exists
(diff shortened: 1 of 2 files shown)


pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -412,12 +412,28 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
...
+		}
+	} else {
+		crSecretExists = true
 	}
 
 	isStale := cr.Generation != cr.Status.LastSyncGeneration
 	hasRecentlySynced := cr.Status.LastSyncTimestamp != nil && cr.Status.LastSyncTimestamp.Add(time.Hour*1).After(time.Now())
 
-	if !isStale && hasRecentlySynced {
+	if !isStale && hasRecentlySynced && crSecretExists {
```
dgoodwin: Aha so we were watching secret deletions and this probably worked at one point, but the logic changed and it was getting skipped if recently synced.

PR #132, opened 2019-10-28
Bug 1717255: add alert for when operator is down

joelddiaz:
use the 'up' metric to detect when the CCO pod is down (ie Promtheus cannot scrape metrics).

### DIFF
--- a/config/manager/namespace.yaml
+++ b/config/manager/namespace.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/node-selector: ""
+  labels:
+    controller-tools.k8s.io: "1.0"
+    openshift.io/run-level: "1"
+    openshift.io/cluster-monitoring: "true"
+  name: openshift-cloud-credential-operator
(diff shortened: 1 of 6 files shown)


sdodson:
Can we get this rebased and merged?

joelddiaz:
@dgoodwin PTAL (specifically restoring the namespace object)

config/manager/namespace.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/node-selector: ""
+  labels:
+    controller-tools.k8s.io: "1.0"
+    openshift.io/run-level: "1"
+    openshift.io/cluster-monitoring: "true"
+  name: openshift-cloud-credential-operator
```
dgoodwin: Duped with https://github.com/openshift/cloud-credential-operator/blob/master/manifests/00_namespace.yaml, can the manager render command use that?

joelddiaz: the render command has access to generated files out of config/*, but anyway, i thought it best to address this in its own PR #136 as there are subtle things going on here

joelddiaz:
@dgoodwin with #136 merged, this should be good to go now

dgoodwin:
Commit still appears in this PR and diff, probably due to a different hash or something, but it should merge clean.

/lgtm

joelddiaz:
> Commit still appears in this PR and diff, probably due to a different hash or something, but it should merge clean.

yeah, i had rebased on that branch for #136

PR #133, opened 2019-10-28
Rough documentation for a disabled cred minter install

dgoodwin:
Not working yet, looking for feedback.


joelddiaz:
this is pretty much what i expected the instructions to look like. i'm surprised to hear you say that it's not working yet

dgoodwin:
Yeah I talked to Trevor last night, he found that basically this doesn't work for making a release image, it comes out missing all the image refs.

I'm trying this morning to convert to using "buildah from" to layer in a change to one file, that too isn't working yet, the cluster just doesn't work.

docs/disabled-operator.md
```diff
@@ -0,0 +1,43 @@
...
+
+  * Extract a release image: `oc adm release extract registry.svc.ci.openshift.org/ocp/release:4.3.0-0.ci-2019-10-23-103858`
+  * Edit 0000_50_cloud-credential-operator_01_operator_configmap.yaml and set disabled to "true".
+  * Examine all CredentialsRequests in the directory for your cloud platform. You will need to manually create secrets for each of these for the location and permissions specified in the request. However we will inject these secrets in the installer manifests rather than the release image, as we do not want to embed powerful cloud credentials in a published release image.
+  * Build a custom release image from what you extracted and modified: `oc adm release new --from-dir="." --to-image quay.io/dgoodwin/origin-release:latest`
+  * Generate installer manifests: `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=quay.io/dgoodwin/origin-release:latest ./bin/openshift-install create manifests --dir=clusters/dgoodwin-aws`
+  * Add yaml files to the clusters manifests dir for each secret matching the CredentialsRequest in the release image for your cloud platform, as identified in the step above:
```
joelddiaz: I had to read this a couple of times before the step sunk in. This might make things a little clearer:
Add the yaml-formatted Secrets files to the cluster's manifests dir (eg clusters/dgoodwin-aws/manifests from the generate command above) for each secret...

joelddiaz:
if you're saying these instructions work then it LGTM

dgoodwin:
Not yet but I'm getting closer :)

dgoodwin:
I was able to get this to work but it required some shuffling of manifests, I made the namespace a separate yaml file, and reordered so the operator configmap is created before the deployment (to ensure we never start minting).

manifests/01_operator_configmap.yaml
```diff
@@ -6,4 +6,4 @@ metadata:
   annotations:
     release.openshift.io/create-only: "true"
 data:
-  disabled: "false"
+  disabled: "true"
```
joelddiaz: we don't want this disabled by default, do we?

dgoodwin: Uhhh no. :) Good catch.

dgoodwin:
Updated.

PR #135, opened 2019-11-13
Add manifest annotations for hosted deployment exclusions

csrwng:
Enables the CVO to exclude manifests in an externally hosted control plane deployment
See https://github.com/openshift/cluster-version-operator/pull/252

### DIFF
--- a/manifests/05_deployment.yaml
+++ b/manifests/05_deployment.yaml
@@ -3,6 +3,8 @@ kind: Role
 metadata:
   name: prometheus-k8s
   namespace: openshift-cloud-credential-operator
+  annotations:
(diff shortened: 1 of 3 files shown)


joelddiaz:
@csrwng so how do in-cluster components get cloud credentials for the hosted control-plane environment?

csrwng:
For now they don’t. IBM public cloud uses its own kube cloud provider for storage and lb’s. And they do their own machine management.

joelddiaz:
it looks like we're good to merge after the format of the annotation is finalized

csrwng:
@joelddiaz the CVO PR has merged, this should be good to go

PR #137, opened 2019-11-25
Bug 1776079: Azure: check existing role assignments before creating a new one

joelddiaz:
Currently when re-reconciling an already-provisioned CredentialsRequest in Azure, the actuator will just always attempt to create a role assignment even if it already exists. There is error handling to catch the "RoleAssignmentExists" error, and it just moves along to the next task.

This results in the Resource Group where the cluster is installed having periodic entries in the Azure console Activity Log for the Resource Group recording these (non-critical) errors:

```
{
    "authorization": {
        "action": "Microsoft.Authorization/roleAssignments/write",
        "scope": "/subscriptions/SUBSCRIPTION_ID/resourceGroups/jdiaz-az-gpqgx-rg/providers/Microsoft.Authorization/roleAssignments/94025186-5e7b-4e18-88de-4625cac3ed19"
    },
    "level": "Error",
    "operationName": {
        "value": "Microsoft.Authorization/roleAssignments/write",
        "localizedValue": "Create role assignment"
    },
    "resourceGroupName": "jdiaz-az-gpqgx-rg",
    "subStatus": {
        "value": "Conflict",
        "localizedValue": "Conflict (HTTP Status Code: 409)"
    },
    "properties": {
        "statusCode": "Conflict",
        "serviceRequestId": "931445cc-fbc6-469a-a6f4-2f7750788255",
        "statusMessage": "{\"error\":{\"code\":\"RoleAssignmentExists\",\"message\":\"The role assignment already exists.\"}}"
    },
}
```

Change the logic to pull the current list of Role Assignments so that we can avoid making unnecessary CreateRoleAssignment calls.

### DIFF
--- a/pkg/azure/minter.go
+++ b/pkg/azure/minter.go
@@ -236,11 +236,34 @@ func (credMinter *AzureCredentialsMinter) AssignResourceScopedRole(ctx context.C
 		return fmt.Errorf("more than one role %q found", targetRole)
 	}
 
+	credMinter.logger.Debugf("getting current role assignments for service principal %s", principalName)
+	filter := fmt.Sprintf("principalId eq '%s'", principalID)
+	currentRoleAssignments, err := credMinter.roleAssignmentsClient.List(ctx, filter)
+	if err != nil {
+		credMinter.logger.WithError(err).Error("failed to list role assignments")
+		return err
+	}
+
 	for _, resourceGroup := range resourceGroups {
-		scope := "subscriptions/" + credMinter.subscriptionID + "/resourceGroups/" + resourceGroup
+		scope := "/subscriptions/" + credMinter.subscriptionID + "/resourceGroups/" + resourceGroup
+
+		// check whether assignment already exists
+		alreadyExists := false
+		for _, r := range currentRoleAssignments {
+			if *r.Properties.Scope == scope {
+				credMinter.logger.Debugf("role %s already assigned in resource group %s for service principal %s", targetRole, resourceGroup, principalName)
+				alreadyExists = true
+				break
+			}
+		}
+		if alreadyExists {
+			continue
+		}
+
 		raName := uuid.NewV4().String()
 
 		err = wait.PollImmediate(5*time.Second, 60*time.Second, func() (bool, error) {
+			credMinter.logger.Debugf("assigning role %s for resource group %s", *roleDefinition.Name, resourceGroup)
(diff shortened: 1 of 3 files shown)


pkg/azure/minter.go
```diff
@@ -236,11 +236,34 @@ func (credMinter *AzureCredentialsMinter) AssignResourceScopedRole(ctx context.C
...
+		return err
+	}
+
 	for _, resourceGroup := range resourceGroups {
 		scope := "subscriptions/" + credMinter.subscriptionID + "/resourceGroups/" + resourceGroup
+
+		// check whether assignment already exists
+		alreadyExists := false
+		for _, r := range currentRoleAssignments {
+			if strings.Contains(*r.Properties.Scope, scope) {
```
dgoodwin: Can you add a comment here clarifying why we can't do an eq and what the data strings look like? I see scope above but what does r.Properties.Scope come out looking like? The test data looks awfully similar.

joelddiaz:
> Were you able to reproduce what the customer found?

Yes. I stood up an Azure cluster and by forcing CCO to re-reconcile a CredentialsRequest, you could then see the log messages on the Azure console (that's how I got that JSON example ;) )

pkg/azure/minter.go
joelddiaz: the Contains() comparison is because our generated scope var string doesn't have a leading /, and the returned r.Properties.Scope does have the leading /.

i reworked the generated string a few lines above to just put the leading / in the string so we can now do a simple string comparison. putting the leading / still allows all the minting to work.

pearj:
> > Were you able to reproduce what the customer found?
> 
> Yes. I stood up an Azure cluster and by forcing CCO to re-reconcile a CredentialsRequest, you could then see the log messages on the Azure console (that's how I got that JSON example ;) )

Thanks for fixing this so quickly!

Out of curiosity, how do you force CCO to re-reconcile? Just delete the container and then it runs as soon as it starts?

dgoodwin:
That will work, as will just adding a meaningless annotation to a credentials request.

joelddiaz:
> Out of curiosity, how do you force CCO to re-reconcile? Just delete the container and then it runs as soon as it starts?

As long as you can get the code to pass through this throttling/sanity check https://github.com/openshift/cloud-credential-operator/blob/master/pkg/controller/credentialsrequest/credentialsrequest_controller.go#L433-L439 , then that will cause the code to go through a full re-reconcile.

PR #138, opened 2019-11-25
preserve azure cred-minter mode

mjudeikis:
make minter mode persistent for azure

This is requirements for Azure RedHat OpenShift service
/cc @jim-minter

### DIFF
--- a/pkg/controller/secretannotator/azure/reconciler.go
+++ b/pkg/controller/secretannotator/azure/reconciler.go
@@ -135,7 +135,6 @@ func (r *ReconcileCloudCredSecret) validateCloudCredsSecret(secret *corev1.Secre
 
 	// TODO(jchaloup): find a way to dynamically check whether these creds really
 	// can be used for minting (or passthrough) or whether they are useless (i.e. InsufficientAnnotation)
-	r.Logger.Info("Platform is azure: allowing to mint new credentials")
 	if err := r.updateSecretAnnotations(secret, constants.MintAnnotation); err != nil {
 		r.Logger.Errorf("error while validating cloud credentials: %v", err)
 		return err
@@ -150,7 +149,14 @@ func (r *ReconcileCloudCredSecret) updateSecretAnnotations(secret *corev1.Secret
 		secretAnnotations = map[string]string{}
 	}
 
-	secretAnnotations[constants.AnnotationKey] = value
+	// ARO relies on being CCO being in passthrough mode -
+	// should not automatically switch to mint mode
+	if secretAnnotations[constants.AnnotationKey] != constants.PassthroughAnnotation {
+		r.Logger.Info("Platform is azure: allowing to mint new credentials")
+		secretAnnotations[constants.AnnotationKey] = value
+	} else {
+		r.Logger.Info("Platform is azure: passthrough credentials")
+	}
 	secret.SetAnnotations(secretAnnotations)
 
 	return r.Update(context.Background(), secret)
--- a/pkg/controller/secretannotator/azure/reconciler_test.go
+++ b/pkg/controller/secretannotator/azure/reconciler_test.go
@@ -2,14 +2,13 @@ package azure_test
 
 import (
 	"context"
+	"reflect"
 	"testing"
 
 	log "github.com/sirupsen/logrus"
-	"github.com/stretchr/testify/assert"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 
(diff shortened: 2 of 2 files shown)


pkg/controller/secretannotator/azure/reconciler.go
```diff
@@ -150,7 +150,11 @@ func (r *ReconcileCloudCredSecret) updateSecretAnnotations(secret *corev1.Secret
 		secretAnnotations = map[string]string{}
 	}
 
-	secretAnnotations[constants.AnnotationKey] = value
+	// ARO relies on being CCO being in passthrough mode -
+	// should not automatically switch to mint mode
+	if secretAnnotations[constants.AnnotationKey] != constants.PassthroughAnnotation {
+		secretAnnotations[constants.AnnotationKey] = value
+	}
```
dgoodwin: I guess if a non-ARO OCP user wants to switch from passthrough to mint, they could get their creds in place and then delete the annotation but this change does make it a little less magical. What do you think @joelddiaz

dgoodwin: Alternative would be some kind of a config option that just says "assume passthrough". We did just recently add a configmap for another feature that could be used for this purpose.

mjudeikis: If to follow code behavior - No annotation meant always use Mint mode. 
This change just to preserves passthrough credentials mode. So this is "assume passthrough" configuration? isn't?

At the moment we are missing credentials checking to define if we can mint credentials. In ARO case this would not be possible so the flow would set to passthrough. So this is stopgap until this feature lands into CCO.

mjudeikis: > Alternative would be some kind of a config option that just says "assume passthrough". We did just recently add a configmap for another feature that could be used for this purpose.

This would cause more confusion. As we get a situation where "my annotations says Mint but behaves like Passthrough"

dgoodwin:
As discussed on slack this is desired for 4.3, will need a bugzilla added for 4.4, we can handle the cloning for 4.3 from there.

mjudeikis:
4.4.0 https://bugzilla.redhat.com/show_bug.cgi?id=1776980
4.3.0 https://bugzilla.redhat.com/show_bug.cgi?id=1776990

pkg/controller/secretannotator/azure/reconciler.go
joelddiaz: If I'm reading these changes correctly, for ARO something outside of the cluster will come along an annotate the secret to force it into passthrough mode.

@mjudeikis why would it not be possible to do credentials checking in ARO? Are there Azure API calls available to allow us to perform the whole passthrough vs mint determination dynamically? Or is this a limitation of the kind of introspection the Azure API allows?

@dgoodwin I think I'd prefer the config option that just lets someone set something like 'forceMode: passthrough'. especially if we already have some external code setting the annotation. I think it makes is more clear if the change is driven through our configmap.

@mjudeikis I don't think Devan is saying having the annotation mismatch the runtime behavior. The annotation would be driven/overridden by the config option.

mjudeikis: @joelddiaz ARO uses the flavor of UPI, and the secret will be created with passthrough annotation in the first place. The plan is to make this option available in the Installer codebase but not used for IPI (https://github.com/jim-minter/installer/commit/588ea7de41723bcdb0ba1b30db00f79dd9a20388). 

By default, the existing code path will not change, and secrets will be created without annotation. Because ARO uses the installer as a library code we will set this during install and we need it to be persistent during install and cluster lifecycle. ConfigMap option would mean day2 operation as it is not available as an asset in the installer.

I think the configMap option is not an option here.

Happy to go via full technical details offline if this helps.

dgoodwin: I can see now that we force mint mode for Azure, previously I think we assumed passthrough? Did this change in 4.2 or 4.3? I assume that is where your breakage originates?

Looking at this again this morning but I still do not feel this is the right approach. We're calling a function to updateSecretAnnotations(secret *corev1.Secret, value string) and then ignoring the value if the current value happens to be passthrough.

That annotation was a programatic implementation detail, it was not meant to be a mode for an external actor to configure how the operator functions. That is a much more logical fit for some kind of configuration, not an annotation on a secret. If we implement the concept of it being something that must be preserved, it's lost it's value and we can't do what we intended with it, which was to indicate if the operator determined you can mint or not. That will be a problem when we get to the TODO on line 136 and now we can't update our annotation.

I'll be on the call today, we may be able to work out a way to get the configuration in at install time via the installer manifests directory support.

joelddiaz: FYI, we had identified the 'mint assumption' when the mint code originally landed https://github.com/openshift/cloud-credential-operator/pull/95#discussion_r311717042 . I can't find the approval, but I remember being told it was okay and that we'd come back and find a way to determine mint vs passthrough dynamically.

jim-minter: I've worked out how to check whether minting can be enabled dynamically - see https://gist.github.com/jim-minter/de0b581fb74435eb5a602bf5b02668ae.  I think that by using that code, you can decide whether or not to set to mint mode, and that will solve the issues both for ARO and for OCP on Azure.

jim-minter: (tl;dr: it turns out that all you need to do is get the oauth token, parse the JWT and inspect the roles claim)

jim-minter:
closing for #141

PR #141, opened 2019-11-27
enable azure passthrough

mjudeikis:
Enables azure checking for mint and passthrough modes

/cc @jim-minter @dgoodwin @joelddiaz


pkg/controller/secretannotator/azure/validate.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/Azure/go-autorest/autorest/azure"
+)
+
+//go:generate mockgen -source=./validate.go -destination=./mock/validate_generated.go -package=mock
+
+// AdalClient is interface for github.com/Azure/go-autorest/autorest/adal for easier testing
+type AdalClient interface {
```
mjudeikis: This is needed for mocking for testing. But I think this pkg might not be the best place. And name too. Suggestions are welcome.

pkg/controller/secretannotator/azure/validate.go
```diff
@@ -0,0 +1,43 @@
...
+	NewOAuthConfig(activeDirectoryEndpoint, tenantID string) (*adal.OAuthConfig, error)
+	NewServicePrincipalToken(oauthConfig adal.OAuthConfig, clientID string, secret string, resource string, callbacks ...adal.TokenRefreshCallback) (*adal.ServicePrincipalToken, error)
+}
+
+type adalClient struct{}
```
mjudeikis: alternative name?

pkg/controller/secretannotator/azure/reconciler_test.go
```diff
@@ -23,20 +26,99 @@ import (
...
+				s.Annotations[constants.AnnotationKey] = constants.InsufficientAnnotation
+			},
+			roles: nil,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			base := getInputSecret()
+			fakeClient := fake.NewFakeClient(base)
```
mjudeikis: This test is quite complicated now. But I think it does what it needs to do :man_shrugging:

pkg/controller/secretannotator/azure/reconciler_test.go
```diff
@@ -47,37 +129,46 @@ func TestAzureSecretAnnotatorReconcile(t *testing.T) {
...
 
-	rcc := &ReconcileCloudCredSecret{
-		Client: fakeClient,
-		Logger: log.WithField("controller", "testController"),
-	}
+func setupAdalMock(r *mock.MockAdalClientMockRecorder, roles []string) {
+	gomock.InOrder(
+		// these methdods are extensivly tested in azure codebase
+		r.NewOAuthConfig(gomock.Any(), gomock.Any()).Return(&TestOAuthConfig, nil),
+		r.NewServicePrincipalToken(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(newServicePrincipalTokenManual(roles), nil),
```
joelddiaz: similar comment here. `gomock.Eq(TestOAuthConfig), gomock.Eq(TestClientID), gomock.Eq(TestClientSecret), gomock.Eq(azure.PublicCloud.GraphEndpoint)`
(with the necessary updates to the generated secret in getInputSecret() )

pkg/controller/secretannotator/azure/reconciler_test.go
```diff
@@ -23,20 +26,99 @@ import (
...
+			rcc := &ReconcileCloudCredSecret{
+				Client: fakeClient,
+				Logger: log.WithField("controller", "testController"),
+				Adal:   mockAdalClient,
+			}
+
+			// error will end-up in InsufficientAnnotation on the secret
+			rcc.Reconcile(reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Name:      "azure-credentials",
```
joelddiaz: can we constify the "azure-credentials" string into something like testSecretName so we can use the var in the three places it appears in this file?

pkg/controller/secretannotator/azure/reconciler_test.go
```diff
@@ -23,20 +26,99 @@ import (
...
+				},
+			})
+
+			secret := &corev1.Secret{}
+			fakeClient.Get(context.TODO(), client.ObjectKey{Name: "azure-credentials", Namespace: testNamespace}, secret)
+			expected := getInputSecret()
+			if test.wants != nil {
+				test.wants(expected)
+			}
+			if !reflect.DeepEqual(secret, expected) {
```
joelddiaz: why not just `assert.Equal(t, expected, secret)`?

pkg/controller/secretannotator/azure/reconciler_test.go
```diff
@@ -47,37 +129,46 @@ func TestAzureSecretAnnotatorReconcile(t *testing.T) {
...
+	)
+}
 
-	_, err := rcc.Reconcile(reconcile.Request{
-		NamespacedName: types.NamespacedName{
-			Name:      "azure-credentials",
-			Namespace: testNamespace,
-		},
-	})
+func newTestServicePrincialToken() (*adal.ServicePrincipalToken, error) {
```
joelddiaz: this function doesn't appear to be called.

pkg/controller/secretannotator/azure/validate.go
joelddiaz: I think this is okay. This is very azure-specific, and it's already in an azure package. I can't think of an alternative.

Maybe the filename of validate.go could be something like adalclient.go. WDYT?

mjudeikis: `AdalService`, as its not a client, It set of libraries and methods for AD

mjudeikis:
/hold
Running some test on kur code base. Will remove hold when done

dgoodwin:
/retitle Bug: 1776980 Dynamically detect if Azure credentials can mint or fallback to passthrough

PR #146, opened 2019-12-09
Bug 1781109: pre-populate conditions with count of zero

joelddiaz:
If a condition has previously been firing and later clears, the metrics calucation logic would not have any CredentialsRequests with the condition firing so there would be no updates to the metrics capturing that condition. This means the condition would never clear.

When initializtion a run for calculating current metrics, pre-populate all the failure condition types to '0' so that we ensure metrics are sent/cleared.

### DIFF
--- a/pkg/controller/credentialsrequest/constants/constants.go
+++ b/pkg/controller/credentialsrequest/constants/constants.go
@@ -1,0 +1,16 @@
+package constants
+
+import (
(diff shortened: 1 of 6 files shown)


sdodson:
/retitle Bug 1781109: pre-populate conditions with count of zero

sdodson:
This is a master branch PR and we're linking it to a bug with target release of 4.4.0, this should be fine.

PR #150, opened 2020-01-09
vSphere passthrough mode for CredentialsRequests

joelddiaz:
Implement a simple passthrough-mode-only actuator for vSphere. Assumes the credentials used during installation are good enough to be passed along to any in-cluster components needing vSphere credentials.

Add the VSphereProviderSpec definitions which allows listing permissions being requested (even though they are presently ignored as passthrough-mode is assumed).


pkg/apis/cloudcredential/v1/vsphere_types.go
```diff
@@ -0,0 +1,61 @@
...
+// to a separate repo.
+
+// VSphereProviderSpec contains the required information to create RBAC role
+// bindings for VSphere.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type VSphereProviderSpec struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Permissions contains a list of groups of privileges that are being requested.
+	Permissions []VSpherePermission `json:"permissions"`
```
dgoodwin: It looks like you've given thought to this and the entities can be tied to specific privileged in the future.

joelddiaz: that's the idea. that we can hopefully extend this in a non-breaking way if/when we get to mint-mode

joelddiaz:
@dgoodwin i was looking at adding vsphere e2e tests, but it doesn't look like vsphere IPI e2e is ready yet, so this should be good to go

pkg/controller/secretannotator/vsphere/reconciler.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/openshift/cloud-credential-operator/pkg/controller/utils"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	controllerName = "secretannotator"
+
+	// VSphereCloudCredSecretName is the name of the secret where credentials
+	// for vSphere are stored.
+	VSphereCloudCredSecretName = "vsphere-creds"
```
staebler: Is it safe to assume this hard-coded value instead of getting the secret name from the cloud provider config? What if there are multiple vCenters with different cred secrets?

joelddiaz: this is an excellent question. @jcpowermac do you know if it is safe to assume that the vsphere secret will always be the secret kube-system/vsphere-creds ?

jcpowermac: I haven't tested it but I wonder if you could continue to append the data section:
```
vcenterA.username: A
vcenterA.password: A
vcenterB.username: B
vcenterB.password: B
```
https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/k8s-secret.html

jcpowermac: @dav1x has done a lot with it maybe he has additional insight.

joelddiaz: the secret does have a format that would seem to allow this
```
data:
  vcsa-ci.vmware.devcluster.openshift.com.password: some_base64_password
  vcsa-ci.vmware.devcluster.openshift.com.username: some_base64_username
```

dav1x: It was only recently that VCP started to support multiple credentials or secrets via separate vCenter servers. 

https://github.com/kubernetes/cloud-provider-vsphere/pull/241

The examples in the PR above do a good job illustrating how this should look.

PR #151, opened 2020-01-10
Fix exclusion annotation for hosted use case

csrwng:
The annotation was in the wrong resource. It needs to be applied to both the deployment and the prometheus rule.

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -8,6 +8,7 @@ metadata:
     controller-tools.k8s.io: "1.0"
   annotations:
     config.openshift.io/inject-proxy: "manager"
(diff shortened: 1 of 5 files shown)


csrwng:
@joelddiaz @dgoodwin ptal

csrwng:
> @csrwng you want to edit config/manager/deployment.yaml and config/manager/prometheusrule.yaml, then run 'make' to have the generated manifests/05_deployment.yaml look the way you want

ahh, thx, will do

csrwng:
@joelddiaz thx, it should be good now

csrwng:
@joelddiaz thx, I think it should be good now

PR #152, opened 2020-01-10
Bug 1779390: force reconcile if any error conditions are set or !Provisioned

joelddiaz:
in the event that a CredentialsRequest object has been recently synced, but has an error condition set, ensure that the CredentialsRequest is re-reconciled to attemp detecting/handling the error condition (so that it can be cleared).

also check the CredentialsRequest.Status.Provisioned flag so that we always re-reconcile if Provisioned != True.

### DIFF
--- a/pkg/controller/credentialsrequest/credentialsrequest_controller.go
+++ b/pkg/controller/credentialsrequest/credentialsrequest_controller.go
@@ -433,8 +433,9 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
 
 	isStale := cr.Generation != cr.Status.LastSyncGeneration
 	hasRecentlySynced := cr.Status.LastSyncTimestamp != nil && cr.Status.LastSyncTimestamp.Add(time.Hour*1).After(time.Now())
+	hasActiveFailureConditions := checkForFailureConditions(cr)
 
-	if !isStale && hasRecentlySynced && crSecretExists {
+	if !isStale && hasRecentlySynced && crSecretExists && !hasActiveFailureConditions && cr.Status.Provisioned {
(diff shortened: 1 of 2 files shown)


pkg/controller/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -433,8 +433,9 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
 
 	isStale := cr.Generation != cr.Status.LastSyncGeneration
 	hasRecentlySynced := cr.Status.LastSyncTimestamp != nil && cr.Status.LastSyncTimestamp.Add(time.Hour*1).After(time.Now())
+	hasActiveFailureConditions := checkForFailureConditions(cr)
 
-	if !isStale && hasRecentlySynced && crSecretExists {
+	if !isStale && hasRecentlySynced && crSecretExists && !hasActiveFailureConditions && cr.Status.Provisioned {
```
dgoodwin: Is it technically feasible to *not* set lastSyncGen and lastSyncTime, if we did not fully reconcile the cred request without error?

I feel like the intent of these fields is to indicate the last successfully synced generation and they should not be set if we hit errors.

dgoodwin: From slack, we don't, normally, and we're not sure how this is possible that we got into this state.

wking:
> in the event that a CredentialsRequest object has been recently synced, but has an error condition set, ensure that the CredentialsRequest is re-reconciled...

When I read this, I thought "but we want to resync if there is an error, even if we haven't been recently synced".  Looking at the code, I see that is the case, where we skip resyncing if everything looks good, but an old sync, or an error condition, or a provision failure are all sufficient on their own to trigger a resync.  Just thought I'd translate that into English to make sure we're all on the same page and head off anyone else making my same mistaken first impression ;).

PR #154, opened 2020-02-05
Bug 1794536: don't report conditions when CCO is disabled

joelddiaz:
If CCO is disabled it should not be triggering Prometheus alerts. Skip reporting condition metrics if CCO has been disabled.

### DIFF
--- a/pkg/controller/metrics/metrics_test.go
+++ b/pkg/controller/metrics/metrics_test.go
@@ -46,47 +46,86 @@ func TestCredentialsRequests(t *testing.T) {
 
 	logger := log.WithField("controller", "metricscontrollertest")
 
-	awsCredRequests := []credreqv1.CredentialsRequest{
-		// just a regular cred request
-		testAWSCredRequest("aregular"),
-		// missing namespace condition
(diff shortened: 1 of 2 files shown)


pkg/controller/metrics/metrics_test.go
```diff
@@ -46,47 +46,87 @@ func TestCredentialsRequests(t *testing.T) {
...
+				assert.Equal(t, 3, accumulator.crTotals["aws"])
+				assert.Equal(t, 1, accumulator.crTotals["gcp"])
+
+				// conditions should all be zero as CCO is disabled
+				assert.Equal(t, 0, accumulator.crConditions[credreqv1.MissingTargetNamespace])
+				assert.Equal(t, 0, accumulator.crConditions[credreqv1.CredentialsProvisionFailure])
+				assert.Equal(t, 0, accumulator.crConditions[credreqv1.Ignored])
+				assert.Equal(t, 0, accumulator.crConditions[credreqv1.InsufficientCloudCredentials])
+			},
+		},
```
dgoodwin: Just a minor style comment, sub-test here probably isn't buying you much, seems there's almost no shared code and the validation functions are separate. Fine as it is though!

joelddiaz:
@dgoodwin i just updated the test case so that it loops through all known failure condition types rather than a static list of current failure conditions. other than that it's the same PR content

PR #158, opened 2020-02-26
provide region when setting up an AWS client

joelddiaz:
this allows the client setup code to insert the region info, which in turn allows the AWS SDK to use the alternative API endpoints when connecting to the AWS China regions.

also reorder parameters so we are consistently using region before infraName.


joelddiaz:
@dgoodwin this is good for review now

PR #159, opened 2020-03-02
Bug 1806892: no need to run CCO as privileged pod

joelddiaz:
Remove the label that was running CCO as a privileged pod. It is unnecessary for the kinds of operations that CCO needs to perform in cluster.
    
Removing the label makes the file copy of any provided certificate bundles fail (used for things like global proxy settings). Fix up the permissions at container build time to allow the Pod startup to continue to be able to copy over any certificate bundle mounted into the pod.
    
The label appears to have no effect when CCO is running on the bootstrap node, so there is no need to maintain a Namespace definition for use only during bootstrap (like we must do for the bootstrap-only Pod definition).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,8 @@ FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 WORKDIR /root/
 COPY --from=builder /go/src/github.com/openshift/cloud-credential-operator/manager .
 ADD manifests/ /manifests
+# Update perms so we can copy updated CA if needed
+RUN chmod -R g+w /etc/pki/ca-trust/extracted/pem/
 
 LABEL io.openshift.release.operator=true
 ENTRYPOINT ["./manager"]
--- a/config/manager/namespace.yaml
+++ b/config/manager/namespace.yaml
@@ -5,6 +5,5 @@ metadata:
     openshift.io/node-selector: ""
   labels:
     controller-tools.k8s.io: "1.0"
-    openshift.io/run-level: "1"
     openshift.io/cluster-monitoring: "true"
   name: openshift-cloud-credential-operator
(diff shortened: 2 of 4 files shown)


dgoodwin:
Can you elaborate on the history and how you reached this conclusion? (preferably in the git commit message) IIRC this developed as a result of that work to move to the bootstrap node but I can't remember the specifics.

joelddiaz:
The history of this privileged label seems unclear. It first appears at https://github.com/openshift/cloud-credential-operator/pull/13/commits/fbf7b74f0a9b66ad4d96910c381923e49c310489

It has since just been carried around.

Removing the label means that the file copy operation on pod startup fails at https://github.com/openshift/cloud-credential-operator/blob/master/manifests/05_deployment.yaml#L202-L205 , so the change in this PR to the Docker file addresses that.

I'll see what I can do to fill out the commit message.

joelddiaz:
/retitle BZ 1806892: no need to run CCO as privileged pod

joelddiaz:
/retitle Bug 1806892: no need to run CCO as privileged pod

Dockerfile
```diff
@@ -20,6 +20,8 @@ FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 WORKDIR /root/
 COPY --from=builder /go/src/github.com/openshift/cloud-credential-operator/manager .
 ADD manifests/ /manifests
+# Update perms so we can copy updated CA if needed
+RUN chmod -R g+w /etc/pki/ca-trust/extracted/pem/
```
lwan-wanglin: @joelddiaz Do we need to give write permission to files under /etc/pki/ca-trust/extracted/pem/ directory? If we just need copy updated CA, is it enough to only give write permission to directory?
I just have a doubt. If it doesn't influences, please ignore my question.

joelddiaz: If there is an existing file with RO permissions, then only changing the directory will still cause a failure.

PR #161, opened 2020-03-04
Bug 1838718: do not special-case handling read-only credentialsRequest

joelddiaz:
In the event that the install-time creds secret is removed (kube-system/aws-creds), reconciling the read-only CredentialsRequest would fail.

It is okay to just use the read-only credentials to make sure that no updates are needed to the read-only CredentialsRequest.

Also, update the function for handling whether a CredentialsRequest already exists so that it covers the passthrough credentials case where the AWS user fields are empty.

### DIFF
--- a/pkg/aws/actuator/actuator_test.go
+++ b/pkg/aws/actuator/actuator_test.go
@@ -17,11 +17,197 @@ limitations under the License.
 package actuator
 
 import (
-	"github.com/stretchr/testify/assert"
+	"fmt"
 	"strings"
 	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/aws/actuator/actuator_test.go
```diff
@@ -17,11 +17,219 @@ limitations under the License.
...
+
+				return r
+			},
+			validate: func(t *testing.T, clientRecorder *awsClientBuilderRecorder) {
+				assert.Equal(t, testRootAccessKeyID, string(clientRecorder.accessKeyID))
+				assert.Equal(t, testRootSecretAccessKey, string(clientRecorder.secretAccessKey))
+			},
+		},
+		{
+			name: "read only creds not ready",
```
dgoodwin: Identical test name to the one above.

joelddiaz:
Yeah, it seems to run fine without the special-case. Do the e2e cases cover a situation where we remove the root creds?

dgoodwin:
Pretty sure they do not cover that.

PR #162, opened 2020-03-05
WIP: change AWS behavior to dynamic determination of passthrough

joelddiaz:
Previous behavior was to store a static list of all permissions in an arbritrary OpenShift release to determine whether the cloud credentials were good enough for passthrough mode. This meant that changes to future OpenShift releases where a component needed updated permissions would require updates to the static list.

Going forward, mark the cloud credentials as 'passthrough' to indicate 'passthrough capable' if the creds can at least perform iam:GetUser and iam:SimulatePrincipalPolicy. Those two permissions are enough to then validate all CredentialsRequests being processed dynamically.

Remove the static list of permissions that have previously been used to determine passthrough mode.

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"reflect"
 
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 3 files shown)


joelddiaz:
@jharrington22 here's the PR that would change the cloud-cred-operator behavior to do dynamic determination of creds when in passthrough mode (similar to how it works in GCP today)

PR #163, opened 2020-03-09
search through source manifests to determine operator disabled

joelddiaz:
When the render process is run during bootstrap, it will check all files in the destination directory (which also contains the install-time manifests) for the CCO ConfigMap that enables/disables the operator.

This should allow an install process where a user can provide their own install-time manifests with the CCO ConfigMap defined to keep CCO from ever running in the cluster. The user-provided ConfigMap will make it into the cluster during the bootstrap stage, so when the in-cluster CCO is installed, it will see the ConfigMap and never attempt to process CredentialsRequests.

### DIFF
--- a/cmd/manager/render.go
+++ b/cmd/manager/render.go
@@ -25,7 +25,12 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	corev1 "k8s.io/api/core/v1"
+	yaml "k8s.io/apimachinery/pkg/util/yaml"
+
+	minterv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
 	"github.com/openshift/cloud-credential-operator/pkg/assets"
+	"github.com/openshift/cloud-credential-operator/pkg/controller/utils"
 )
 
 const (
@@ -99,6 +104,8 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
 	log.SetLevel(level)
 	log.Debug("debug logging enabled")
 
+	operatorDisabled := isDisabled()
+
 	log.Infof("Rendering files to %s", renderOpts.destinationDir)
 
 	ccoRenderDir := renderOpts.destinationDir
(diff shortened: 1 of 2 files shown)


joelddiaz:
@dgoodwin by making this new parameter optional, it means we need changes into the installer (simple changes).

In reality, the --dest-dir location mounted into the pod is where the manifests already are so we could accomplish this with the existing mountpoints loaded into the pod. But in the future if the installer is refactored so that this isn't the case, having the explicit parameter will make it more clear that additional consideration is needed. WDYT?

cmd/manager/render.go
```diff
@@ -115,22 +124,76 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+		decoder := yaml.NewYAMLOrJSONDecoder(file, 4096)
+		configMap := &corev1.ConfigMap{}
+		if err := decoder.Decode(configMap); err != nil {
+			log.WithError(err).Debug("failed to decode into configmap")
+			continue
+		}
+
+		if configMap.Namespace == minterv1.CloudCredOperatorNamespace && configMap.Name == minterv1.CloudCredOperatorConfigMap {
+
+			disabled, err := utils.CCODisabledCheck(configMap, &log.Logger{Level: log.GetLevel()})
```
staebler: This does not set up the logger sufficiently and will cause a panic if the logger is ever used. Use `log.New` instead.

dgoodwin:
So the manifest render command is called by the installer, and --dest-dir is where we're supposed to drop our manifests, but if the user injected custom manifests, these are also present there?

If so I think that's enough API stability, I can't envision they would start telling us to write somewhere manifests would not exist anymore. 

It would be simpler to avoid the installer change, I'd say lets assume dest-dir points to the manifests dir.

This is a nice clever solution to the problem.

joelddiaz:
@dgoodwin ok, removed the --manifests-dir parameter and will just search through the destination dir for the CCO configmap. No installer changes needed with this approach.

PR #166, opened 2020-03-13
Bug 1813343: handle old Infrastructure objects without PlatformStatus

joelddiaz:
Older clusters that have been upgraded over time will have a correspondingly old version of the Infrastructure CR where Status.PlatformStatus did not exist.

Detect that case, and return an empty region string. The AWS client code will just use the defaults, and the permissions simulations will simply not include the region as part of the simulation.

### DIFF
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -70,6 +70,12 @@ func LoadInfrastructureRegion(c client.Client, logger log.FieldLogger) (string, 
 		logger.WithError(err).Error("error loading Infrastructure region")
 		return "", err
 	}
+	if infra.Status.PlatformStatus == nil {
+		// Older clusters may have an Infrastructure object without the PlatformStatus fields.
+		// Send back an empty region and the AWS client will use default settings.
+		// The permissions simulation will also simply not fill out the region for simulations.
+		return "", nil
+	}
 	return infra.Status.PlatformStatus.AWS.Region, nil
 }
 


pkg/controller/utils/utils.go
```diff
@@ -24,6 +24,8 @@ const (
 	awsCredsSecretIDKey          = "aws_access_key_id"
 	awsCredsSecretAccessKey      = "aws_secret_access_key"
 	operatorConfigMapDisabledKey = "disabled"
+
+	defaultAWSRegion = "us-east-1"
```
wking: This is a risky assumption.  In the absence of `PlatformStatus` you should fall back to the install-config.  I think.  Looks like the registry used to do that, but has since stopped?  openshift/cluster-image-registry-operator#470  Hrm, we might need to revert that too...

joelddiaz: My thinking here is that IAM is region-less and the default endpoints seem to do exactly this https://github.com/openshift/cloud-credential-operator/blob/master/vendor/github.com/aws/aws-sdk-go/aws/endpoints/defaults.go#L1427-L1439

wking: > My thinking here is that IAM is region-less...

So why do we need to look at `PlatformStatus` at all, vs. just hard-coding `us-east-1` in #124 for [rhbz#1750338][1]?

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1750338

joelddiaz: AWS China is effectively a separate cloud that happens to have the same API as regular AWS. So setting the region in the client will then cause the aws-go-sdk to talk to the AWS China endpoints.

For #124, a customer can define the specifics in their permissions policies, so we just try to do the best simulation possible for how the actual in cluster components will behave with AWS.

wking: > So setting the region in the client will then cause the aws-go-sdk to talk to the AWS China endpoints.

So I'm not clear.  Do we care which region gets used or not?  If we do, we want to use `PlatformStatus`, falling back to the install-config ConfigMap.  If we don't, we can hard-code `us-east-1`.  Why would we only care for clusters with `PlatformStatus` but not for clusters without it?

joelddiaz: Yes we care about the region, b/c the default (from aws-go-sdk) of 'us-east-1' does not work for AWS China. Falling back to the ConfigMap would also work, but falling back to 'us-east-1' seems to accomplish the same result b/c the Infrastructure object has started to carry the necessary region information for clusters installed when that information matters.

wking: > aFalling back to the ConfigMap would also work, but falling back to 'us-east-1' seems to accomplish the same result...

Except for China?

> ...b/c the Infrastructure object has started to carry the necessary region information for clusters installed when that information matters.

But we can still have 4.1 -> 4.2 -> 4.3 -> 4.4 clusters that don't have PlatformStatus set, because there is still no migration code that fills in PlatformStatus.  Clusters born into 4.2+ will have PlatformStatus, but that's not all clusters.

joelddiaz: Since there are no 4.1 AWS China clusters, falling back to 'us-east-1' for all 4.1-era clusters (where 'us-east-1' was what was already being used) should be fine.

joelddiaz: as @csrwng pointed out, we can just send back an empty region string, and the existing code already handles that gracefully. so no 'us-east-1' default :)

PR #170, opened 2020-03-26
add aws-pod-identity-webhook to image refs and deployment

sjenning:
This gets the payload image for `aws-pod-identity-webhook` into the operator.  After this, we can start adding code to this operator to manage this admission webhook and other AWS pod identity bits.

fyi @derekwaynecarr

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -42,6 +42,8 @@ spec:
       - env:
         - name: RELEASE_VERSION
           value: "0.0.1-snapshot"
+        - name: AWS_POD_IDENTITY_WEBHOOK_IMAGE
+          value: quay.io/openshift/aws-pod-identity-webhook:latest
         image: quay.io/openshift/origin-cloud-credential-operator:latest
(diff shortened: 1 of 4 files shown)


derekwaynecarr:
can you get a document in openshift/enhancements that explains rationale for cloud cred operator to deploy the webhook.  i dont disagree with choice, but its good to clarify.

joelddiaz:
I think what you want here is to make the edit to the deployment in config/manager/deployment.yaml and then use the Makefile to generate manifests/05_deployment.yaml

sjenning:
@joelddiaz thanks, fixed :+1:

sjenning:
enhancement xref https://github.com/openshift/enhancements/pull/260

sjenning:
@joelddiaz would you be ok approving this? I'm still working on the enhancement but, unless we suddenly decide to do something _completely_ different from the current solution, the CCO will need this image wired in from the release payload.

Doing this will allow me to iterate in development more quickly and verify that all the image building linkages are working for CI and ART.

PR #171, opened 2020-03-30
add explicit --manifests-dir for render bootstrap

joelddiaz:
allow explicitly setting a --manifests-dir where the render process should look for the CCO configmap to determin whether or not to enable CCO during bootstrap.

### DIFF
--- a/cmd/manager/render.go
+++ b/cmd/manager/render.go
@@ -79,6 +79,7 @@ var (
 	}
 
 	renderOpts struct {
+		manifestsDir   string
 		destinationDir string
 		ccoImage       string
 		logLevel       string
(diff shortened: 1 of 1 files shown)


dgoodwin:
Can you clarify why we'd add this option and modify the installer, vs just assuming /asset/manifests?

joelddiaz:
Just following how other static pods are initialized that rely on "input" sources.

Some examples:
https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L131
https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L211

dgoodwin:
/lgtm

SOLD!

PR #172, opened 2020-03-31
document disabling CCO

joelddiaz:


joelddiaz:
/hold
holding until openshift/installer#3381 is merged

PR #177, opened 2020-04-07
switch to go modules, standard makefile, and bump

sjenning:
Unfortunately, this ended up being a huge PR.

The things it does:
- moves from `dep` managed vendor tree to Go modules
  - involves adding `dependencymagnet` to get tools in `vendor`
  - small change to azure unit tests due to sdk bump
- moves to a kube 1.17.2 base
- moves to standard Makefile (https://github.com/openshift/build-machinery-go)
- moves to conventional `bindata` in repo root
- rename binary to `cloud-credential-operator` to conform to standard Makefile and debugability (`manager` in the ps output vs `cloud-credential-operator`)
- split out `operator` subcommand to conform to convention on operators that have a `render` mode as well
- Rework `Dockerfile` to conform to convention
- remove `kustomize` from flow


sjenning:
might have to adjust prow in the short term to get this in as the make targets have changed

Makefile
```diff
@@ -1,80 +1,95 @@
...
-manager: generate fmt vet
-	go build -o bin/manager github.com/openshift/cloud-credential-operator/cmd/manager
-
-# Build without attempting any code generation. (CI)
-build-no-gen: fmt vet
-	go build -o bin/manager github.com/openshift/cloud-credential-operator/cmd/manager
-
-# Run against the configured Kubernetes cluster in ~/.kube/config
-run: generate fmt vet
-	go run ./cmd/manager --log-level=debug
```
dgoodwin, line 39: Make install and run are both used for devel and documented in README, we need to either preserve these or update the docs for how to do these standard dev tasks.

Makefile
```diff
@@ -1,80 +1,95 @@
...
-	kustomize build config | kubectl apply -f -
-
-# Generate manifests e.g. CRD, RBAC etc.
-manifests:
-	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd
-	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go rbac --name cloud-credential-operator
-	# kustomize and move to manifests dir for release image:
-	kustomize build config > manifests/05_deployment.yaml
-	cp config/manager/namespace.yaml manifests/00_namespace.yaml
-	cp config/crds/cloudcredential_v1_credentialsrequest.yaml manifests/00_v1_crd.yaml
```
dgoodwin, line 58: What target do we run for an equivalent to manifests? Looks like update-codegen?

Makefile
```diff
@@ -1,80 +1,95 @@
...
+
+# TODO targets for backward compatibility while we make the shift in CI
+test-no-gen: test
+.PHONY: test-no-gen
+
+vet: verify-govet
+.PHONY: vet
+
+build-no-gen: build
+.PHONY: build-no-gen
```
dgoodwin, line 95: Is this expected? 

```
❯ make build-no-gen
fatal: No names found, cannot describe anything.
go build -mod=vendor -trimpath -ldflags "-s -w -X github.com/openshift/cloud-credential-operator/pkg/version.versionFromGit="v0.0.0-unknown" -X github.com/openshift/cloud-credential-operator/pkg/version.commitFromGit="342cc25f" -X github.com/openshift/cloud-credential-operator/pkg/version.gitTreeState="clean" -X github.com/openshift/cloud-credential-operator/pkg/version.buildDate="2020-04-08T11:22:36Z" " github.com/openshift/cloud-credential-operator/cmd/cloud-credential-operator
```

Makefile
```diff
@@ -1,80 +1,95 @@
...
+# $3 - Dockerfile path
+# $4 - context directory for image build
+$(call build-image,ocp-cloud-credential-operator,$(IMAGE_REGISTRY)/ocp/4.5:cloud-credential-operator, ./Dockerfile,.)
+
+# This will call a macro called "add-crd-gen" will will generate crd manifests based on the parameters:
+# $1 - target name
+# $2 - apis
+# $3 - manifests
+# $4 - output
+$(call add-crd-gen,manifests,$(CRD_APIS),./manifests,./manifests)
```
dgoodwin, line 75: I like the move to just manifests/

Makefile
```diff
@@ -1,80 +1,95 @@
...
+
+update-codegen: update-codegen-crds
+.PHONY: update-codegen
```
dgoodwin, line 78: What is the difference between update-codegen and update-generated mentioned in one of the comments above?

manifests/05_deployment.yaml
```diff
@@ -203,7 +203,7 @@ spec:
               echo "Copying system trust bundle"
               cp -f /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
-          exec /root/manager --log-level=debug
+          exec /usr/bin/cloud-credential-operator operator --log-level=debug
```
dgoodwin, line 206: :+1:  thanks for cleaning this up.

pkg/dependencymagnet/doc.go
```diff
@@ -0,0 +1,10 @@
+// +build tools
+
+// go mod won't pull in code that isn't depended upon, but we have some code we don't depend on from code that must be included
+// for our build to work.
+package dependencymagnet
```
dgoodwin, line 5: Is "dependencymagnet" a tool? Or just this process of pinning a dep in doc.go? We recently did this in Hive, curious if there's a more official way.

pkg/operator/controller.go
```diff
@@ -0,0 +1,129 @@
+/*
```
dgoodwin, line 1: What was the rationale behind moving pkg/controller to pkg/operator

Makefile
sjenning: for `fatal: No names found, cannot describe anything.` yes it is expected.  `git describe` doesn't work if there are no tag, as is the case with this repo.  Probably not forever though.  There has been talk to tag all repos on a per release (or build) basis.

Makefile
sjenning: yes, `update-codegen` is the new target.  that is actually not working right now.  the command runs without error but it does not create the yaml file.  still have to run that down.

Makefile
sjenning: i can add those targets back

pkg/dependencymagnet/doc.go
sjenning: There isn't a more official way.  I copied what KASO did.
https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/dependencymagnet/doc.go
In upstream kube they do something similar in a different location
https://github.com/kubernetes/kubernetes/blob/master/build/tools.go

pkg/operator/controller.go
sjenning: This is the convention in other operators, mostly because operators grow to have more than one controller in them and each controller is organized in a new subdirectory of `pkg/operator`.
https://github.com/openshift/cluster-kube-apiserver-operator/tree/master/pkg/operator

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,204 @@
...
+const (
+	podTemplate = `apiVersion: v1
+kind: Pod
+metadata:
+  name: cloud-credential-operator
+  namespace: openshift-cloud-credential-operator
+spec:
+  containers:
+  - command:
+    - /root/manager
```
joelddiaz, line 45: point to new /usr/bin/cloud-cred-operator binary name

tmp/cco-bootstrap/bootstrap-manifests/cloud-credential-operator-pod.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
```
joelddiaz: files in 'tmp' dir?

tmp/cco-bootstrap/manifests/cco-cloudcredential_v1_credentialsrequest.yaml
```diff
@@ -0,0 +1,108 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
joelddiaz: 'tmp' dir again?

pkg/cmd/render/render.go
sjenning: good one!

sjenning:
meh, worth a shot :P

sjenning:
Great, e2e-aws passed.  Pushed once more because I failed to update the bindata on the last push.  I'll add a `make verify` test in prow after this merges.

Ready for review/lgtm/approve.

manifests/05_deployment.yaml
joelddiaz, line 206: down below, the container name of 'manager' might make better sense as 'cloud-credential-operator' (or just operator)

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,205 @@
...
+  containers:
+  - command:
+    - /usr/bin/cloud-credential-operator
+    args:
+	- operator
+    - --log-level=debug
+    - --kubeconfig=/etc/kubernetes/secrets/kubeconfig
+    image: %s
+    imagePullPolicy: IfNotPresent
+    name: manager
```
joelddiaz, line 52: same comment about the name of the container name of 'manager'

manifests/05_deployment.yaml
sjenning: sure, i'll do this in a follow on PR

pkg/cmd/render/render.go
sjenning: sure, i'll do this in a follow on PR

PR #180, opened 2020-04-13
move actuators to new subdirectory

sjenning:
@joelddiaz @dgoodwin 

This is a cleanup with no functional change.

There are now 6 actuators under `pkg`.  I thought it would be nice to have a `pkg/actuators` subdirectory to hold all on the cloud provider specific stuff.

There is still `pkg/operator/secretannotator` that has cloud specific stuff in it.  Would be nice to move that as well.


sjenning:
nice to see the new `verify` test is working :+1:

abhinavdahiya:
> pkg/aws/actuator/actuator_test.go → pkg/actuators/aws/actuator/actuator_test.go

 this reads wrong. if `pkg/actuators` contains all actuators, then why does `pkg/actuators/aws` has another package `actuator`

if there are multiple types of aws actuators, like for cred_minter, secret_minter, something else, i those should be the packages in `pkg/actuators/aws`

sjenning:
@abhinavdahiya yeah, it is weird.

`aws, gcp, vsphere` have subdirs named `actuator` to seperate them from their `mock` implementations.  `azure, openstack, ovirt` do not and only have `actuator.go`

I'm not attached to the name `actuators`.  We could name it `providers` like we do in other repos.  That would allow us to put provider-specific non-actuator functionally in here later like the aforementioned provider specific `secretannotator` stuff.

@abhinavdahiya you like `providers` better?

dgoodwin:
What about moving pkg/actuators/aws/actuators/* to pkg/actuators/aws/

joelddiaz:
my 2 cents would go for a directory scheme like pkg/platforms/aws/actuator

dgoodwin:
Needs update if we're going to get this in.

PR #183, opened 2020-04-15
Bug 1823631: fix bootstrap static pod

sjenning:
Fix regression from #177 

@dgoodwin @joelddiaz @dulek


dulek:
Nice, thank you!

PR #184, opened 2020-04-16
use alternate IAM endpoints if specified

joelddiaz:
Search through the cluster's Infrastructure object to see if an alternate IAM endpoint is specified, and use it when creating the AWS client.

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -31,6 +31,7 @@ import (
 	awsannotator "github.com/openshift/cloud-credential-operator/pkg/operator/secretannotator/aws"
 	annotatorconst "github.com/openshift/cloud-credential-operator/pkg/operator/secretannotator/constants"
 	"github.com/openshift/cloud-credential-operator/pkg/operator/utils"
+	awsutils "github.com/openshift/cloud-credential-operator/pkg/operator/utils/aws"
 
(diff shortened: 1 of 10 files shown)


joelddiaz:
> Looks good, is it feasible to unit test or excessively difficult?

The best I could think of was to put some test cases to cover that we're processing the contents of the Infrastructure object correctly.

Once we receive the awsClient object, I don't see a way to inspect the client to see what the connection settings are for it.

dgoodwin:
Ok that's reasonable.

/lgtm

joelddiaz:
@dgoodwin this should be good to go now. I can squash and re-push once you've had a look.

dgoodwin:
Looks good, squash away.

dgoodwin:
tide Pending — Not mergeable. Needs bugzilla/valid-bug label.

joelddiaz:
did all the supporting work this to work finally make it in time for 4.5 (and we need an exception for this)? or can this wait?

dgoodwin:
Need conflicts resolved here @joelddiaz

joelddiaz:
@sjenning fyi, i did mess with your oidc controller a bit to get this working. Not thrilled with having to add a controller-runtime.Client dependency on it, WDYT?

joelddiaz:
@dgoodwin FYI, rebased to fix merge conflicts. no functional changes.

PR #185, opened 2020-04-22
add static resource controller

sjenning:
Begins implementing https://github.com/openshift/enhancements/pull/260

This controller deploys the static and templated resources required for the AWS pod identity webhook.

The image pull spec is templated into the `Deployment` using the envvar injected into the operator by the CVO and the `service-ca` is templated into the `MutatingWebhookConfiguration`.

The controller reconciles the assets if they are changed and updates the `MutatingWebhookConfiguration` and does a forced rollout of the `Deployment` if the `service-ca` rotates.

/cc @dgoodwin @joelddiaz @derekwaynecarr @marun @deads2k


sjenning:
Still have some RBAC work to do.
/hold

marun:
@sjenning Is the `service-ca` related to the service-ca-operator or something else?

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+
+const (
+	controllerName    = "staticresource"
+	deploymentName    = "cloud-credential-operator"
+	operatorNamespace = "openshift-cloud-credential-operator"
+)
+
+var (
+	staticFiles = []string{
+		"v4.1.0/aws-pod-identity-webhook/sa.yaml",
```
dgoodwin, line 44: Is this v4.1.0 the version of aws-pod-identity-webhook? Seems like the directory structure should be inverted, aws-pod-identity-webhook/v4.1.0/sa.yaml

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+	return nil
+}
+
+func isManaged(meta metav1.Object) bool {
+	// all managed resources are named pod-identity-webhook
+	if meta.GetName() == "pod-identity-webhook" && meta.GetNamespace() == operatorNamespace {
+		return true
+	}
+	// we also care about service-ca rotation
+	if meta.GetName() == "service-ca" && meta.GetNamespace() == "openshift-config-managed" {
```
dgoodwin, line 147: Where is the watch setup for this? I can't spot it.

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+		err = c.Watch(&source.Informer{Informer: informer}, &handler.EnqueueRequestForObject{}, p)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func isManaged(meta metav1.Object) bool {
```
dgoodwin, line 141: I assume if we went ahead with a separate config CRD, we could use owner references to more reliably act on what we watch? But as it stands, there is no resource we could mark all of these as owned by.

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+		if err != nil {
+			return err
+		}
+
+		informer, err := cache.GetInformer(obj)
+		if err != nil {
+			return err
+		}
+
+		err = c.Watch(&source.Informer{Informer: informer}, &handler.EnqueueRequestForObject{}, p)
```
dgoodwin, line 132: This is a neat approach to watching everything we created, we may use this in Hive where our operator currently does not watch everything.

Any concerns about memory now that we're watching and thus caching a number of new types cluster wide?

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
+package staticresource
```
dgoodwin, line 1: Could I request a controller rename, this feels awspodidentity specific and not as generic as staticresource sounds. If we had additional static resource setup to do, we'd probably want that in another controller rather than forcing into this one.

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+	eventRecorder                          events.Recorder
+	deploymentGeneration                   int64
+	mutatingWebhookConfigurationGeneration int64
+	forceRollout                           bool
+}
+
+var _ reconcile.Reconciler = &staticResourceReconciler{}
+
+func (r *staticResourceReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.WithField("controller", controllerName).Debugf("reconciling after watch event %#v", request)
```
dgoodwin, line 165: Can use r.logger which already carries the "controller" field context. Several uses of this below.

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+	if request.Name == "service-ca" && request.Namespace == "openshift-config-managed" {
+		log.WithField("controller", controllerName).Infof("%s configmap change detected", request.NamespacedName)
+		r.forceRollout = true
+	}
+	r.ReconcileResources()
```
dgoodwin, line 170: Looks like errors are swallowed here, and thus would not re-try if there was a temporary conflict or similar.

manifests/03-deployment.yaml
```diff
@@ -0,0 +1,75 @@
...
+          fi
+          exec /usr/bin/cloud-credential-operator operator --log-level=debug
+        command:
+        - /bin/bash
+        - -ec
+        env:
+        - name: RELEASE_VERSION
+          value: 0.0.1-snapshot
+        - name: AWS_POD_IDENTITY_WEBHOOK_IMAGE
+          value: quay.io/openshift/aws-pod-identity-webhook:latest
```
dgoodwin, line 35: Is this the value that gets used? Is it acceptable for us to have this kind of implied image which is not reflected in the release image itself? Is the use of latest permanent as well?

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+			continue
+		}
+		log.WithField("controller", controllerName).Infof("%s reconciled successfully", result.Type)
+	}
+
+	// "v4.1.0/aws-pod-identity-webhook/deployment.yaml"
+	requestedDeployment := resourceread.ReadDeploymentV1OrDie(v410_00_assets.MustAsset("v4.1.0/aws-pod-identity-webhook/deployment.yaml"))
+	requestedDeployment.Spec.Template.Spec.Containers[0].Image = os.Getenv("AWS_POD_IDENTITY_WEBHOOK_IMAGE")
+	resultDeployment, modified, err := resourceapply.ApplyDeployment(r.clientset.AppsV1(), r.eventRecorder, requestedDeployment, r.deploymentGeneration, r.forceRollout)
+	r.forceRollout = false
```
dgoodwin, line 196: Could some strange things happen here with this forceRollout state being set on the reconciler if we were reconciling concurrently?

pkg/operator/staticresource/controller.go
sjenning: correct

pkg/operator/staticresource/controller.go
sjenning: that's because it isn't there.  good catch!

pkg/operator/staticresource/controller.go
sjenning: I do have concerns about memory and about privilege.  We have to put these abilities on a `ClusterRole` rather than a `Role`, which is not ideal.

I've been looking for a way to limit the scope of the watch to be namespace local, but haven't found it yet.

dgoodwin: @staebler sorry to pull you in here but where did we land with per-namespace watches and limiting cache, if memory serves there was no way to do this?

pkg/operator/staticresource/controller.go
sjenning: Yeah, I thought in the beginning I would make this more generic but... it didn't happen.  I can rename.

pkg/operator/staticresource/controller.go
sjenning: wonderful!

pkg/operator/staticresource/controller.go
sjenning: yep, needs fixing

manifests/03-deployment.yaml
sjenning: `quay.io/openshift/aws-pod-identity-webhook:latest` is a placeholder that is replaced by the CVO with the image pull spec during release image creation
https://github.com/openshift/cloud-credential-operator/blob/master/manifests/image-references#L9-L12

pkg/operator/staticresource/controller.go
sjenning: yes

pkg/operator/staticresource/controller.go
sjenning: So the cache is at the Manager level, not on the Controller. As other parts of the operator do watch things on a cluster wide basis (namespace creation, secrets, etc), we can't scope down the manager cache.

We could create a new controller-local cache, separate from the Manager cache, and base the Watches on it.  I'll play around with it.

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func Add(mgr manager.Manager, kubeconfig string) error {
+	// Currently the only static assets are AWS only
+	infraStatus, err := platform.GetInfraStatusUsingKubeconfig(mgr, kubeconfig)
+	if err != nil {
+		return err
+	}
+	platformType := platform.GetType(infraStatus)
+	if platformType != configv1.AWSPlatformType {
```
joelddiaz, line 76: Is it accurate to say that in one potential future (not implying that we'd want to support this...), we could have a cluster running on GCP accessing resources in AWS?

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+	applyResults := resourceapply.ApplyDirectly(
+		(&resourceapply.ClientHolder{}).WithKubernetes(r.clientset),
+		r.eventRecorder,
+		v410_00_assets.Asset,
+		staticFiles...,
+	)
+	for _, result := range applyResults {
+		if result.Error != nil {
+			log.WithField("controller", controllerName).WithError(result.Error).Errorf("error reconciling %s", result.Type)
+			continue
```
joelddiaz, line 184: Should we error rather than continue if we've failed to apply some required assets?

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+	deploymentGeneration                   int64
+	mutatingWebhookConfigurationGeneration int64
+	forceRollout                           bool
+}
+
+var _ reconcile.Reconciler = &staticResourceReconciler{}
+
+func (r *staticResourceReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.WithField("controller", controllerName).Debugf("reconciling after watch event %#v", request)
+	if request.Name == "service-ca" && request.Namespace == "openshift-config-managed" {
```
joelddiaz, line 166: can we constify these names since they're also used further below?

pkg/operator/staticresource/controller.go
```diff
@@ -0,0 +1,289 @@
...
+	if err != nil {
+		log.WithField("controller", controllerName).WithError(err).Error("error reading service-ca configmap")
+	}
+	caBundle, ok := serviceCAConfigMap.Data["ca-bundle.crt"]
+	if !ok {
+		log.WithField("controller", controllerName).Error("ca-bundle.crt key not found service-ca configmap")
+	}
+	requestedMutatingWebhookConfiguration := ReadMutatingWebhookConfigurationV1Beta1OrDie(v410_00_assets.MustAsset("v4.1.0/aws-pod-identity-webhook/mutatingwebhook.yaml"))
+	requestedMutatingWebhookConfiguration.Webhooks[0].ClientConfig.CABundle = []byte(caBundle)
+	resultMutatingWebhookConfiguration, modified, err := ApplyMutatingWebhookConfiguration(r.clientset.AdmissionregistrationV1beta1(), r.eventRecorder, requestedMutatingWebhookConfiguration, r.mutatingWebhookConfigurationGeneration)
```
joelddiaz, line 216: What is the purpose of storing and sending the mutatingWebhookConfigurationGeneration?

sjenning:
@marun yes, the service-ca-signer is the CA that signs `Services` that have the `service.beta.openshift.io/serving-cert-secret-name` annotation. It creates a cert/key and injects a secret so that TLS application can be accessed by name on the service network.

pkg/operator/staticresource/controller.go
sjenning: I think I'm going to pull the watching of the `service-ca` and redeployment logic for this PR.  We will need to add that but for simplicity of review, I think it will be better to leave that out for now, especially since I've switched the cache to be namespace-local (I'll have to add the Manager provided cache back in to access the `service-ca`)

pkg/operator/staticresource/controller.go
sjenning: actually no.  that is referring to the OCP version.  the idea is that if some point in the future, you need access to both the old asset and the new in order to do some non-trivial reconciliation, you'd have both versions.  not sure if i buy the organization/usefulness of that though.  it's just how the kube-* operators do it. eg https://github.com/openshift/cluster-kube-apiserver-operator/tree/master/bindata/v4.1.0/kube-apiserver

pkg/operator/staticresource/controller.go
sjenning: https://github.com/openshift/cloud-credential-operator/pull/185#discussion_r414758727

bindata/v4.1.0/aws-pod-identity-webhook/deployment.yaml
```diff
@@ -0,0 +1,37 @@
...
+        image: ${IMAGE}
+        imagePullPolicy: Always
+        command:
+        - /usr/bin/aws-pod-identity-webhook
+        - --in-cluster
+        - --namespace=openshift-cloud-credential-operator
+        - --port=6443
+        - --service-name=pod-identity-webhook
+        - --tls-secret=pod-identity-webhook
+        - --annotation-prefix=eks.amazonaws.com
```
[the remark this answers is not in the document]
sjenning: at some point before release, yes.  could we track this for a follow-on so that we aren't bogged down on naming; the hardest of issues in software development?

bindata/v4.1.0/aws-pod-identity-webhook/mutatingwebhook.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: pod-identity-webhook
+webhooks:
+- name: pod-identity-webhook.amazonaws.com
+  failurePolicy: Ignore
```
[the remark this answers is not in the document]
sjenning: This was just copied from upstream.  It is a whole different ball game if you use `Fail` here, as you can take out the cluster by denying pod admission if something goes wrong.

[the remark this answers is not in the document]
sjenning: I would prefer to keep this as is for now during devel.  I'll track this for revisiting before release.

bindata/v4.1.0/aws-pod-identity-webhook/deployment.yaml
```diff
@@ -0,0 +1,37 @@
...
+  template:
+    metadata:
+      labels:
+        app: pod-identity-webhook
+    spec:
+      serviceAccountName: pod-identity-webhook
+      containers:
+      - name: pod-identity-webhook
+        image: ${IMAGE}
+        imagePullPolicy: Always
```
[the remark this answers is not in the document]
sjenning: Yes and yes.

pkg/operator/staticresource/controller.go
sjenning: https://github.com/openshift/cloud-credential-operator/pull/185#discussion_r414758727

pkg/operator/staticresource/controller.go
sjenning: https://github.com/openshift/cloud-credential-operator/pull/185#discussion_r414758727

pkg/operator/staticresource/controller.go
sjenning: That is the easiest way to detect if something external has changed the `MutatingWebhookConfiguration` as the apiserver will increment the `Generation`.  `ApplyDeployment` does the same thing (rather than try to DeepEquals the `spec` or something like that)

pkg/operator/staticresource/controller.go
sjenning: @dgoodwin made the same point about this code not being generic.  I'm renaming the package to `awspodidentity`. 

And no, I don't see this code being used on any other platform since it is really only useful for assuming IAM roles, which... I guess you could do from instances on another cloud but... I can't see us supporting that.

bindata/v4.1.0/aws-pod-identity-webhook/deployment.yaml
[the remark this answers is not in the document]
joelddiaz: agreed. maybe just stick a # TODO comment if possible.

bindata/v4.1.0/aws-pod-identity-webhook/mutatingwebhook.yaml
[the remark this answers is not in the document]
joelddiaz: Understood about the potentially catastrophic consequences if the webhook starts to error. Just wanted confirmation that I understood why.

pkg/operator/staticresource/controller.go
joelddiaz: I suppose my question comes from not seeing ApplyMutatingWebhookConfiguration() do anything with that final parameter that is passed to it.

sjenning: ah yes, very true.  in fact, i'm going to take that out completely in light @marun sharing that i can just add an annotation to the `MutatingWebhookConfiguration` and the `caBundle` will be managed for me.  thus external `Generation` changes would be expected.

sjenning:
/hold cancel

no point since master is locked

sjenning:
Updates:
- use `imagePullPolicy: Always`
- use `service.beta.openshift.io/inject-cabundle` to inject service-ca
- create a namespace-local non-manager informer cache the `awspodidentity` controller can do a very narrow watch and we can use `Role` rather than `ClusterRole`
- `awsPodIdentityController` now retries the reconcilation in a loop every 10s before starting the namespace-local informer cache
- replaced `log.WithFields(log.Fields{"controller": controllerName})` with `r.logger`
- removed `Generation` tracking for `MutatingWebhookConfiguration`
- requeue on failed reconcilation
- rename package from `staticresource` to `awspodidentity`

pkg/operator/staticresource/controller.go
dgoodwin: What was this annotation and feature that manages the caBundle for you? I am fighting similar problems over in Hive.

dgoodwin: n/m found it on slack.

pkg/operator/awspodidentity/controller.go
```diff
@@ -0,0 +1,287 @@
...
+func ReadMutatingWebhookConfigurationV1Beta1OrDie(objBytes []byte) *admissionregistrationv1beta1.MutatingWebhookConfiguration {
+	requiredObj, err := runtime.Decode(defaultCodecs.UniversalDecoder(admissionregistrationv1beta1.SchemeGroupVersion), objBytes)
+	if err != nil {
+		panic(err)
+	}
+	return requiredObj.(*admissionregistrationv1beta1.MutatingWebhookConfiguration)
+}
+
+// ApplyMutatingWebhookConfiguration merges objectmeta, does not worry about anything else
+func ApplyMutatingWebhookConfiguration(client admissionregistrationclientv1beta1.MutatingWebhookConfigurationsGetter, recorder events.Recorder, required *admissionregistrationv1beta1.MutatingWebhookConfiguration) (*admissionregistrationv1beta1.MutatingWebhookConfiguration, bool, error) {
```
marun, line 244: As written, modification of the webhook config spec (e.g the url) by other actors will not prompt action by the operator. Is this intentional?

sjenning: yes, this is true.  i currently just ensures that the resource exists basically.

since an outside controller injects the service-ca, we really can't use `Generation` as Deployments do to detect changes and external changes are expected.

I would like to check more here.  I'll add a TODO.

sjenning:
trying to explain why the `AWS_POD_IDENTITY_WEBHOOK_IMAGE` env var doesn't seem to be set sometimes
```
38m         Normal    ServiceAccountCreated                 deployment/cloud-credential-operator              Created ServiceAccount/pod-identity-webhook -n openshift-cloud-credential-operator because it was missing
38m         Normal    ClusterRoleCreated                    deployment/cloud-credential-operator              Created ClusterRole.rbac.authorization.k8s.io/pod-identity-webhook because it was missing
38m         Normal    RoleCreated                           deployment/cloud-credential-operator              Created Role.rbac.authorization.k8s.io/pod-identity-webhook -n openshift-cloud-credential-operator because it was missing
38m         Normal    ClusterRoleBindingCreated             deployment/cloud-credential-operator              Created ClusterRoleBinding.rbac.authorization.k8s.io/pod-identity-webhook because it was missing
38m         Normal    RoleBindingCreated                    deployment/cloud-credential-operator              Created RoleBinding.rbac.authorization.k8s.io/pod-identity-webhook -n openshift-cloud-credential-operator because it was missing
38m         Normal    ServiceCreated                        deployment/cloud-credential-operator              Created Service/pod-identity-webhook -n openshift-cloud-credential-operator because it was missing
35m         Warning   DeploymentCreateFailed                deployment/cloud-credential-operator              Failed to create Deployment.apps/pod-identity-webhook -n openshift-cloud-credential-operator: Deployment.apps "pod-identity-webhook" is invalid: spec.template.spec.containers[0].image: Required value
30m         Warning   DeploymentCreateFailed                deployment/cloud-credential-operator              Failed to create Deployment.apps/pod-identity-webhook -n openshift-cloud-credential-operator: Deployment.apps "pod-identity-webhook" is invalid: spec.template.spec.containers[0].image: Required value
26m         Warning   DeploymentCreateFailed                deployment/cloud-credential-operator              Failed to create Deployment.apps/pod-identity-webhook -n openshift-cloud-credential-operator: Deployment.apps "pod-identity-webhook" is invalid: spec.template.spec.containers[0].image: Required value
25m         Normal    DeploymentCreated                     deployment/cloud-credential-operator              Created Deployment.apps/pod-identity-webhook -n openshift-cloud-credential-operator because it was missing
25m         Normal    MutatingWebhookConfigurationCreated   deployment/cloud-credential-operator              Created MutatingWebhookConfiguration.admissionregistration.k8s.io/pod-identity-webhook because it was missing

File: pkg/operator/awspodidentity/controller.go (hunk 17 above)
Comment (sjenning) [reply pkg/operator/awspodidentity/controller.go:251]: Actually, I might just add this.  Seems like if you iterate through `wehooks` and clear out `configConfig.caBundle`, we could just do a `DeepEquals` on it.

Comment (marun) [reply pkg/operator/awspodidentity/controller.go:251]: Please review the documentation for Apply{DaemonSet,Deployment}:

https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/apps.go#L21

I think it makes sense to set the input spec hash as an annotation to enable detection of changes across operator releases. 

Regarding the use of generation to detect changes made by other actors, you're right that it can't be used in the same manner as other apply methods due to the CA field being managed by service ca. Since this behavior depends on the service ca injection annotation, it should be possible to detect when to retain the CA field of the existing resource to avoid racing. At worst this implies a second operator-initiated update when the CA is changed by service ca (to pick up the generation). I think that's a small price to pay to ensure that the operator can ensure the intended spec.

cc: @deads2k

**Comment by sjenning** [comment]
Weird, installed a new cluster and now I'm not seeing it.  Added a proper fatal error to capture it though.

**Comment by sjenning** [comment]
flake https://bugzilla.redhat.com/show_bug.cgi?id=1817588

/test e2e-aws-upgrade

**Comment by sjenning** [comment]
ok, all tests are green.  I've deployed a cluster with the release image from this PR and everything looks good.  I'm going to lock this in and move on to adding the `serviceAccountRef` to the CRD.

File: pkg/operator/awspodidentity/controller.go [hunk 18]
```diff
@@ -0,0 +1,297 @@
...
+}
+
+var _ reconcile.Reconciler = &staticResourceReconciler{}
+
+func (r *staticResourceReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	r.logger.Debugf("reconciling after watch event %#v", request)
+	err := r.ReconcileResources()
+	if err != nil {
+		r.logger.Errorf("reconciliation failed, retrying in %s", retryInterval.String())
+		return reconcile.Result{RequeueAfter: retryInterval}, err
```
joelddiaz, line 191: Why a static retry schedule instead of using the built-in exponential back-off?

pkg/operator/awspodidentity/controller.go
```diff
@@ -0,0 +1,297 @@
...
+			continue
+		}
+		r.logger.Infof("%s reconciled successfully", result.Type)
+	}
+
+	// "v4.1.0/aws-pod-identity-webhook/deployment.yaml"
+	requestedDeployment := resourceread.ReadDeploymentV1OrDie(v410_00_assets.MustAsset("v4.1.0/aws-pod-identity-webhook/deployment.yaml"))
+	requestedDeployment.Spec.Template.Spec.Containers[0].Image = r.imagePullSpec
+	resultDeployment, modified, err := resourceapply.ApplyDeployment(r.clientset.AppsV1(), r.eventRecorder, requestedDeployment, r.deploymentGeneration, false)
+	r.deploymentGeneration = resultDeployment.Generation
```
joelddiaz, line 218: Should we read and store the Generation field before checking whether an error ocorred?

sjenning:
changes carried on in https://github.com/openshift/cloud-credential-operator/pull/187

And linking back to unaddressed review https://github.com/openshift/cloud-credential-operator/pull/187#issuecomment-623675188

sjenning:
reopening as the new head of current development and to get a CI release image with these changes

sjenning:
these CI release images get reclaimed quickly!

/test e2e-aws

PR #187, opened 2020-05-04
Api changes

sjenning:
/cc @dgoodwin @joelddiaz @derekwaynecarr @marun @deads2k

Implementing option #3 from https://github.com/openshift/enhancements/pull/260#issuecomment-622409273

Carrying everything from https://github.com/openshift/cloud-credential-operator/pull/185

This just makes the API change and needed wiring changes in the core credentailsrequest controller to minimize reviewer cognitive load.  There is no functional change.

This should make the next PR localized to the AWS actuator (creating Roles and linking ARNs to `ServiceAccount`s)


sjenning:
Note to myself.  Need to address @joelddiaz comments in https://github.com/openshift/cloud-credential-operator/pull/185#pullrequestreview-401998504

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
```diff
@@ -44,17 +44,55 @@ const (
...
 // CredentialsRequestSpec defines the desired state of CredentialsRequest
 type CredentialsRequestSpec struct {
 	// SecretRef points to the secret where the credentials should be stored once generated.
-	SecretRef corev1.ObjectReference `json:"secretRef"`
+	// +optional
+	SecretRef corev1.ObjectReference `json:"secretRef,omitempty"`
+
+	// StorageDestination specifies the type and location of a resource to which the credentials
+	// information can be stored or linked
+	StorageDestination StorageDestination `json:"storage,omitempty"`
```
dgoodwin, line 57: Apologies I may have misunderstood your other proposal (3), but I expected StorageDestination here to be more like StorageType, which is a const string just indicating which field is set on the CredentialsRequest. I don't think we should introduce a new way to set the secret target when we have to keep the old.

i.e.
credRequest.Spec.Type or StorageType
credRequest.Spec.SecretRef
credRequest.Spec.ServiceAccountRef

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
```diff
@@ -44,17 +44,55 @@ const (
...
+	// SecretStorageDestination specifies a serviceaccount to which the credentials will be linked
+	// This is currently only supported on AWS
+	// +optional
+	ServiceAccount *ServiceAccountStorageDestination `json:"serviceAccount,omitempty"`
+}
+
+type SecretStorageDestination struct {
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`
+}
```
dgoodwin, line 89: IMO we should just stick with ObjectReference here, even if we don't use or honor all the fields.

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
```diff
@@ -44,17 +44,55 @@ const (
...
+	SecretStorageDestinationType         = "Secret"
+	ServiceAccountStorageDestinationType = "ServiceAccount"
+)
+
+// StorageDestination defines the type and location of a resource to which the credentials
+// information can be stored or linked
+// +union
+type StorageDestination struct {
+	// StorageType contains the type of storage that should be used for the credentials
+	// +unionDiscriminator
```
dgoodwin, line 73: Whoah, what does this do?

sjenning: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190325-unions.md

However, afaict, it doesn't have an impact on the OpenAPIv3 schema that is output by the generator :confused: I figured I'd keep it in there though.

https://github.com/openshift/api/search?q=unionDiscriminator

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
sjenning: @deads2k how strong is your opinion that `ObjectReference` should not be used?

dgoodwin: If this is best practice I'd be curious to know.

However we should stay consistent with secretRef which is already ObjectReference. 

Perhaps we could switch it to a type though, as we only would be respecting the Name and Namespace, and API callers would not be impacted?

In any case, would like to see them the same.

dgoodwin: Just heard in another review that ObjectReference use is discouraged. So can we safely flip secretRef? Should we make them different?

sjenning: I'm not saying we should change `secretRef` but if we are going to create a new `storage` union as part of this work, I'm saying we shouldn't make the _new_ field `ObjectReference`

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
sjenning: We could flatten the `storage` field down into the `spec`.  The reason I didn't is that now we have to deal with the existing `SecretRef` at that level and if we continue to use that we can't deprecate it in favor of the better typed `secret` field in the `storage` union.

dgoodwin: I'm +1 for collapsing, I really don't want two secretRef fields with one deprecated on our v1 API, immediately making all our cred requests today in various repos technically using deprecated fields.

I'm +1 for ServiceAccountRef being ObjectReference, keeping it consistent with SecretRef, even though this is not recommended use now. +0 for ServiceAccountRef being typed instead of ObjectRef, it's weird to have Secret and ServiceAccount out of sync but I can live with it.

I'm +1 for making SecretRef typed as well *if* that's ok, from an API caller perspective I think it's valid as we're still just using name+namespace but I have no idea if this is an acceptable thing to do. I'm going to ask in forum-api-review.

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
deads2k: > @deads2k how strong is your opinion that `ObjectReference` should not be used?

It's pretty strong.  We should not use ObjectReference.  

```
//  1. Ignored fields.  It includes many fields which are not generally honored.  For instance, ResourceVersion and FieldPath are both very rarely valid in actual usage.
//  2. Invalid usage help.  It is impossible to add specific help for individual usage.  In most embedded usages, there are particular
//     restrictions like, "must refer only to types A and B" or "UID not honored" or "name must be restricted".
//     Those cannot be well described when embedded.
//  3. Inconsistent validation.  Because the usages are different, the validation rules are different by usage, which makes it hard for users to predict what will happen.
//  4. The fields are both imprecise and overly precise.  Kind is not a precise mapping to a URL. This can produce ambiguity
//     during interpretation and require a REST mapping.  In most cases, the dependency is on the group,resource tuple
//     and the version of the actual struct is irrelevant.
//  5. We cannot easily change it.  Because this type is embedded in many locations, updates to this type
//     will affect numerous schemas.  Don't make new APIs embed an underspecified API type they do not control.
```

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
deads2k: Yes, it's best to list this

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
deads2k: I recommend against collapsing the discriminated union in your top level spec.  It's easier to reason about what an API does if the "pick one" isn't co-located with "applies in to anything" (in this case your `providerSpec` for instance).

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
```diff
@@ -44,17 +44,55 @@ const (
...
+	Type string `json:"type"`
```
deads2k, line 74: explicitly required if you would.

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
```diff
@@ -44,17 +44,55 @@ const (
...
+	Secret *SecretStorageDestination `json:"secret,omitempty"`
+
+	// SecretStorageDestination specifies a serviceaccount to which the credentials will be linked
+	// This is currently only supported on AWS
+	// +optional
+	ServiceAccount *ServiceAccountStorageDestination `json:"serviceAccount,omitempty"`
+}
+
+type SecretStorageDestination struct {
+	Namespace string `json:"namespace"`
```
deads2k, line 87: help with an example.  If I create the credential request in namespace foo, I can reference a secret in namespace bar?  Does this work today?

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
```diff
@@ -44,17 +44,55 @@ const (
...
+	Name      string `json:"name"`
+}
+
+type ServiceAccountStorageDestination struct {
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`
```
deads2k, line 93: add help, indicate that this is for a serviceaccount and these are required.

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
dgoodwin: I believe this works today. All component cred requests land in our openshift-cloud-credential-operator namespace by convention, but I don't think we explicitly filter on this. If you have permissions to create CredRequests, you can do what you describe.

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
dgoodwin: The one thing I hoped to avoid in this is two places to set the secret ref, deprecating all our existing usage, and carrying that confusion forever.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -300,13 +306,54 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
...
+
 	// Maintain a copy, but work on a copy of the credentials request:
 	origCR := cr
 	cr = cr.DeepCopy()
 
+	// Copy deprecated secretRef to the new storage field for internal processing
+	if cr.Spec.StorageDestination.Type == "" {
+		cr.Spec.StorageDestination.Type = minterv1.SecretStorageDestinationType
+	}
+	cr.Spec.StorageDestination.Secret = &minterv1.SecretStorageDestination{
```
joelddiaz, line 318: always performing this operation means that the contents of the spec.SecretRef is always the canonical location of the secret data (as updating spec.storageDestination.Secret would be a no-op). is that the intended behavior?

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -300,13 +306,54 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
...
+		setInvalidStorageType(cr, r.platformType)
+		err := r.updateStatus(origCR, cr, logger)
+		if err != nil {
+			logger.WithError(err).Error("failed to update conditions")
+		}
+		return reconcile.Result{}, err
+	}
+
+	// CRs must contain a reference for their storage type (union)
+	if (storageType == minterv1.SecretStorageDestinationType && cr.Spec.StorageDestination.Secret == nil) ||
```
joelddiaz, line 334: how would spec.storageDestination.secret ever be nil if we unconditionally populate it a few lines above?

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -410,32 +473,48 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
...
-	} else {
-		logger.Debug("found secret namespace")
-		setMissingTargetNamespaceCondition(cr, false)
-
 	}
 
+	logger.Debug("found storage destination namespace")
+	setMissingTargetNamespaceCondition(cr, false)
+
 	// Check if the secret the credRequest wants created already exists
```
joelddiaz, line 481: update comment as we now handle more than secrets

pkg/operator/credentialsrequest/credentialsrequest_controller.go
sjenning: good call.  information in the new field should supercede information in the old field.  makes me wonder if we should fail validation and set and error condition on the CR if both are specified...

pkg/operator/credentialsrequest/credentialsrequest_controller.go
sjenning: yes, i moved the setting of the secret inside the `if type == ""` block above

pkg/apis/cloudcredential/v1/credentialsrequest_types.go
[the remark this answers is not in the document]
sjenning: i've been thinking the `Destination` bit is wordy.  i'll remove it.

sjenning:
* dropped `Destination` from field name (too wordy)
* setting of `spec.storage.secret` from `spec.secretRef` only happens if `spec.storage.type` is empty (new field supersedes)
* improved API field descriptions and added explicit `required` where needed

sjenning:
/hold
see https://github.com/openshift/enhancements/pull/260#issuecomment-628851030

PR #176, opened 2020-04-06
vendor: prune vendor tree

sjenning:
Currently the `vendor` tree is 574MB in size.  For comparison, kubernetes/kubernetes `vendor` is 113MB.

I massaged the `Gopkg.toml` and ran `dep prune` and got it to ~~96MB~~50MB.  Compile/fmt/vet times are all improved significantly.

I ran all the make targets to see if anything was missing after the prune.  Awaiting the official testing to make sure everything that was being used is still there.

@joelddiaz

PR #193, opened 2020-05-27
Bug 1808175: update vendor of golang.org/x/crypto

joelddiaz:
dep ensure -v -update golang.org/x/crypto

 Bringing vendor into sync
(1/1) Wrote golang.org/x/crypto@master: revision changed (e657309f52 -> 06a226fb4e)


joelddiaz:
nevermind. we'll try backporting https://github.com/openshift/cloud-credential-operator/pull/176

PR #194, opened 2020-05-27
[release-4.4] Bug 1808175: CVE-2020-9283 ose-cloud-credential-operator-container: golang.org/x/crypto: Processing of crafted ssh-ed25519 public keys allows for panic

joelddiaz:
manual cherry pick of openshift/cloud-credential-operator#176 as the cherrypick bot and github don't like the size of the original PR


sdodson:
Overriding bugzilla/valid-bug because this is a clean pick from upstream which was originally not intended to fix a bug but ultimately resolves the linked bug here by virtue of removing dead code which tripped security scanning on vulnerable libraries.

PR #195, opened 2020-05-27
Add OIDC discovery endpoint controller

sjenning:
@joelddiaz @dgoodwin @derekwaynecarr @abhinavdahiya @marun 

This adds a controller that
1. If `spec.serviceAccountIssuer` is not set in the `cluster` `Authentication` config, it sets it to the s3 location
1. Reconciles an s3 bucket with discovery and keys documents for the ODIC endpoint that can be used in IAM/STS

If `serviceAccountIssuer` is set to some other location, this controller disables itself and does not overwrite.  This allows the customer to set their own issuer without interference from the controller.

Remaining to be done
~~1. Presently only the `service-account-001.pub` is used in the `bound-sa-token-signing-certs` configmap, thus rotation of that signer is not currently supported.  I am going to change the code to just blindly put all `data` keys of the configmap as `keys` in the keys JSON document.~~  This is addressed now.

This also rebased all deps to 1.18 (one commit below the HEAD of this PR) to get the `serviceAccountIssuer` field on an openshift/api bump.

Just a reminder, I'm carrying previous PRs during the freeze so review is only requests for the most recent commit. ` pkg/operator/oidcdiscoveryendpoint/controller.go` is the heart of the change.


pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,413 @@
...
+	auth, err := r.configclientset.ConfigV1().Authentications().Get(context.TODO(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	serviceAccountIssuer := r.getIssuerURL()
+	if auth.Spec.ServiceAccountIssuer != "" {
+		r.logger.Debug("Authentication ServiceAccountIssuer already specified")
+		if auth.Spec.ServiceAccountIssuer != serviceAccountIssuer {
+			r.logger.Debug("Authentication ServiceAccountIssuer is not set to the s3 bucket location, skipping OIDC reconciliation")
```
dgoodwin, line 222: Should this be info or warn instead of debug?

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,413 @@
...
+		return false, nil
+	}
+
+	auth.Spec.ServiceAccountIssuer = serviceAccountIssuer
+	_, err = r.configclientset.ConfigV1().Authentications().Update(context.TODO(), auth, metav1.UpdateOptions{})
+	return false, err
+}
+
+func (r *s3EndpointReconciler) getIssuerURL() string {
+	return fmt.Sprintf("https://s3-%s.amazonaws.com/%s", r.region, r.getBucketName())
```
dgoodwin, line 234: Will there be issues here with govcloud or any non-standard AWS endpoints?

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,413 @@
...
+
+	awsClient, err := aws.NewClient(accessKey, secretKey, r.region, r.infrastructureName)
+	if err != nil {
+		return fmt.Errorf("error creating aws client: %v", err)
+	}
+
+	// Ensure bucket exists
+	bucketName := r.getBucketName()
+	createBucketInput := &s3.CreateBucketInput{
+		Bucket: awssdk.String(bucketName),
```
dgoodwin, line 268: Should this bucket be tagged with the infra ID for cleanup on deprovision?

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,413 @@
...
+	}
+
+	_, err = awsClient.CreateBucket(createBucketInput)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case s3.ErrCodeBucketAlreadyExists:
+				r.logger.Errorf("bucket %s already exists but it not owned by us", bucketName)
```
dgoodwin, line 276: Please add .WithError(err)

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,413 @@
...
+			case s3.ErrCodeBucketAlreadyOwnedByYou:
+				r.logger.Debugf("bucket %s already exists and is owned by us", bucketName)
+			default:
+				return aerr
+			}
+		} else {
+			return err
+		}
+	} else {
+		r.logger.Debugf("bucket %s created", bucketName)
```
dgoodwin, line 287: Suggest .WithField("bucket", bucketName).Info("bucket created")

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,413 @@
...
+}
+
+var _ reconcile.Reconciler = &s3EndpointReconciler{}
+
+func (r *s3EndpointReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	r.logger.Info("reconciling AWS S3 OIDC discovery endpoint on detected change to service account token signer")
+
+	skip, err := r.reconcileClusterAuthentication()
+	if err != nil {
+		r.logger.Error(err)
```
dgoodwin, line 194: Suggest changing both of these to WithError(err).Error("error reonciling cluster authentication") or similar, so we have a little more context as to where the error is coming from.

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,413 @@
...
+
+func (r *s3EndpointReconciler) getBucketName() string {
+	return fmt.Sprintf("%s-oidc", r.infrastructureName)
+}
+
+func (r *s3EndpointReconciler) reconcileS3Endpoint() error {
+	r.logger.Debugf("reconciling S3 endpoint resources")
+
+	// Get the root secret and create an AWS client
+	secret, err := r.kubeclientset.CoreV1().Secrets(constants.CloudCredSecretNamespace).Get(context.TODO(), awsannotator.AWSCloudCredSecretName, metav1.GetOptions{})
```
dgoodwin, line 245: Tangential discussion but we are in the process of documenting and supporting a mode of operation where the admin cloud creds can be removed from the cluster after install, and only recreated before an upgrade, leaving the creds to coast in the meantime. (and cco has a r/o cred to monitor and report problems)

Will this mode of operation work with the sts tokens?

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,413 @@
...
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			default:
+				return aerr
+			}
+		} else {
+			return err
+		}
+	}
+	r.logger.Debugf("discovery document updated")
```
dgoodwin, line 311: Info, no "f" required

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,413 @@
...
+
+	// Extract the token signer and create keys JSON s3 object for jwks_uri
+	tokenSignerConfigMap, err := r.kubeclientset.CoreV1().ConfigMaps(tokenSignerConfigMapNamespace).Get(context.TODO(), tokenSignerConfigMapName, metav1.GetOptions{})
```
dgoodwin, line 314: Curious why you're using the kubeclientset as opposed to the controller-runtime Client? Has always seemed easier to work with to me.

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: it is a persistent condition if the customer sets their own `ServiceAccountIssuer` so I didn't want it to be a warning since it just indicates an expected condition.  It would print each time the bound service account signing key rotated (I'm not sure how often that is off the top of my head).

i could make it an `info` if you like?

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: yes.  i might look at how image-registry handles this. it is the only other cluster component i know of that writes to s3.

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: ah, i was wondering about that was done :+1:

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: currently, i'm using those root creds to reconcile to the s3 bucket.  if we don't have those we will either need to 1) mint limited s3 creds for this controller to use or 2) disable ourselves and print a message saying the OIDC endpoint is not available

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: i had to create a clientset for the openshift config API (for accessing `Authentications`) so I just wanted to stay in that mindset.  no particular reason.  if you perfer the controller-runtime client, i switch to it for the kube API requests.

sjenning:
@dgoodwin confirmed that tagging the bucket results in deletion during cluster deprovision

In terms of unit tests, I'll have to think.  I know there is a mock aws client, but not sure how to leverage that yet.

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: registry is has a public block by default and doesn't need to know the public URL if it is made public.  so... hmmm.

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,420 @@
...
+	logger             log.FieldLogger
+	eventRecorder      events.Recorder
+	infrastructureName string
+	region             string
+}
+
+var _ reconcile.Reconciler = &s3EndpointReconciler{}
+
+func (r *s3EndpointReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	r.logger.Info("reconciling AWS S3 OIDC discovery endpoint on detected change to service account token signer")
```
joelddiaz, line 189: It's not just a change that triggers a Reconcile(), right (kube will periodically cause this to run even w/o a change)? Is it more accurate to remove the "detected change" text?

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,420 @@
...
+
+	skip, err := r.reconcileClusterAuthentication()
```
joelddiaz, line 191: rather than `skip` can we call this something like `userManagedOIDC`?

pkg/operator/oidcdiscoveryendpoint/controller.go
joelddiaz: I'm not sure how to interpret the data at https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html , but it seems to indicate that only `us-gov-west-1` would be valid...

And https://github.com/aws/aws-sdk-go/blob/master/aws/endpoints/defaults.go#L7876-L7891 makes it seem like we can just use the `region` as the code is presently doing.

pkg/operator/oidcdiscoveryendpoint/controller.go
joelddiaz: @sjenning I assume these proposed "limited s3 creds" would be permanent Secret-based creds?

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,420 @@
...
+
+	// Get the root secret and create an AWS client
+	secret, err := r.kubeclientset.CoreV1().Secrets(constants.CloudCredSecretNamespace).Get(context.TODO(), awsannotator.AWSCloudCredSecretName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	accessKey, ok := secret.Data[awsannotator.AwsAccessKeyName]
+	if !ok {
+		return fmt.Errorf("couldn't fetch key containing AWS_ACCESS_KEY_ID from cloud cred secret")
```
joelddiaz, line 251:
```suggestion
		return fmt.Errorf("couldn't fetch key containing %s from cloud cred secret", awsannotator.AwsAccessKeyName)
```

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+
+	secretKey, ok := secret.Data[awsannotator.AwsSecretAccessKeyName]
+	if !ok {
+		return fmt.Errorf("couldn't fetch key containing AWS_ACCESS_KEY_ID from cloud cred secret")
```
joelddiaz, line 256:
```suggestion
		return fmt.Errorf("couldn't fetch key containing %s from cloud cred secret", awsannotator.AwsSecretAccessKeyName)
```

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,420 @@
...
+
+	controllerRef := &corev1.ObjectReference{
+		Kind:      "deployment",
+		Namespace: operatorNamespace,
+		Name:      deploymentName,
+	}
+	eventRecorder := events.NewKubeRecorder(kubeclientset.CoreV1().Events(operatorNamespace), deploymentName, controllerRef)
+	logger := log.WithFields(log.Fields{"controller": controllerName})
+
+	if infraStatus.PlatformStatus == nil || infraStatus.PlatformStatus.AWS == nil {
```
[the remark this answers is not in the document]
sjenning: Hmmm... interesting.  That does throw a wrench in things.

It is also unsettling to know this kind of landmine exists :grimacing: I'm not sure how devs are supposed to know this.

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: as i attempted to switch to the controller client, i remember why i used the kubeclientset.  it was to get the `v1.EventInterface` needed for the `eventRecorder` on line 115

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+	controllerName                = "oidcdiscoveryendpoint"
+	deploymentName                = "cloud-credential-operator"
+	operatorNamespace             = "openshift-cloud-credential-operator"
+	tokenSignerConfigMapName      = "bound-sa-token-signing-certs"
+	tokenSignerConfigMapNamespace = "openshift-kube-apiserver"
+	credentialsSecretNamespace    = "openshift-cloud-credential-operator"
+	credentialsSecretName         = "cloud-credential-operator-s3-creds"
+	discoveryURI                  = ".well-known/openid-configuration"
+	keysURI                       = "keys.json"
+	discoveryTemplate             = `{
```
abhinavdahiya, line 55: we have bindata already loading assets, why not this template.

pkg/operator/oidcdiscoveryendpoint/controller.go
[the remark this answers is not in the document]
abhinavdahiya: we have fixed this with migration in 4.5, https://github.com/openshift/cluster-config-operator/tree/master/pkg/operator/migration_aws_status

pkg/operator/oidcdiscoveryendpoint/controller.go
abhinavdahiya: https://github.com/terraform-providers/terraform-provider-aws/blob/934b07ac2dc27a97e041e5f8558d2dd3bf369ae1/aws/resource_aws_s3_bucket.go#L1722-L1734

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+	err = r.reconcileS3Resources()
+	if err != nil {
+		r.logger.WithError(err).Error("failed reconciling S3 resources")
+		return reconcile.Result{Requeue: true}, err
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (r *s3EndpointReconciler) reconcileClusterAuthentication() (bool, error) {
```
abhinavdahiya, line 213: can we comment what the return values are?

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+	if auth.Spec.ServiceAccountIssuer != "" {
+		r.logger.Debug("Authentication ServiceAccountIssuer already specified")
+		if auth.Spec.ServiceAccountIssuer != serviceAccountIssuer {
+			r.logger.Info("Authentication ServiceAccountIssuer is not set to the s3 bucket location, skipping OIDC reconciliation")
+			return true, nil
+		}
+		return false, nil
+	}
+
+	auth.Spec.ServiceAccountIssuer = serviceAccountIssuer
```
abhinavdahiya, line 231: needs a deepcopy before modifying.

pkg/operator/oidcdiscoveryendpoint/controller.go
abhinavdahiya, line 213: Also personally i would like if this function is named to hint what it actually does, like `reconcileServiceAccountIssuer`

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+		return fmt.Errorf("error creating aws client: %v", err)
+	}
+
+	// Ensure bucket exists
+	bucketName := r.getBucketName()
+	_, err = awsClient.CreateBucket(&s3.CreateBucketInput{
+		Bucket: awssdk.String(bucketName),
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
```
abhinavdahiya, line 274: errors.As or errors.Is

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+	}
+
+	awsClient, err := aws.NewClient(accessKey, secretKey, r.region, r.infrastructureName)
+	if err != nil {
+		return fmt.Errorf("error creating aws client: %v", err)
+	}
+
+	// Ensure bucket exists
+	bucketName := r.getBucketName()
+	_, err = awsClient.CreateBucket(&s3.CreateBucketInput{
```
abhinavdahiya, line 270: have we though about how GET->CREATE vs CREATE only affects our rate limiting? Maybe AWS allows a lot more GETs and that would prevent us from sending CREATE, while depending on CREATE to check if exists might trigger more rate limiting.

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+				r.logger.WithField("bucket", bucketName).Info("bucket already exists and is owned by us")
+			default:
+				return aerr
+			}
+		} else {
+			return err
+		}
+	} else {
+		r.logger.WithField("bucket", bucketName).Info("bucket created")
+	}
```
abhinavdahiya, line 289:
```suggestion
	}
	r.logger.WithField("bucket", bucketName).Info("bucket created")
```

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+		Bucket: awssdk.String(bucketName),
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case s3.ErrCodeBucketAlreadyExists:
+				r.logger.WithField("bucket", bucketName).WithError(err).Error("bucket already exists but it not owned by us")
+				return aerr
+			case s3.ErrCodeBucketAlreadyOwnedByYou:
+				r.logger.WithField("bucket", bucketName).Info("bucket already exists and is owned by us")
```
abhinavdahiya, line 280: Don't we think this is going to be just a lot of positive logging?

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+	// Tag bucket for deprovisioning
+	_, err = awsClient.PutBucketTagging(&s3.PutBucketTaggingInput{
+		Bucket: awssdk.String(bucketName),
+		Tagging: &s3.Tagging{
+			TagSet: []*s3.Tag{
+				{
+					Key:   awssdk.String("kubernetes.io/cluster/" + r.infrastructureName),
+					Value: awssdk.String("owned"),
+				},
+			},
```
abhinavdahiya, line 300:
```suggestion
			TagSet: []*s3.Tag{{
					Key:   awssdk.String("kubernetes.io/cluster/" + r.infrastructureName),
					Value: awssdk.String("owned"),
			}},
```

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+		},
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
```
abhinavdahiya, line 304: errors.As or errors.Is

abhinavdahiya: also returning the aws error as-is no better than returning the err itself

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+		ACL:    awssdk.String("public-read"),
+		Body:   awssdk.ReadSeekCloser(strings.NewReader(discoveryJSON)),
+		Bucket: awssdk.String(bucketName),
+		Key:    awssdk.String(discoveryURI),
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			return aerr
+		}
+		return err
```
abhinavdahiya, line 324: same as https://github.com/openshift/cloud-credential-operator/pull/195/commits/985fc52ed17d4e37214df7acd57369dbee35429e#r432781039

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+
+	pemKeys := [][]byte{}
+	for _, pemKey := range tokenSignerConfigMap.Data {
+		pemKeys = append(pemKeys, []byte(pemKey))
+	}
+	if len(pemKeys) < 1 {
+		return fmt.Errorf("no signing keys found in config map %s/%s", tokenSignerConfigMapNamespace, tokenSignerConfigMapName)
+	}
+
+	keysJSON, err := encodeKeys(pemKeys)
```
abhinavdahiya, line 342: error is not handled.

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+	hasher := crypto.SHA256.New()
+	hasher.Write(publicKeyDERBytes)
+	publicKeyDERHash := hasher.Sum(nil)
+
+	keyID := base64.RawURLEncoding.EncodeToString(publicKeyDERHash)
+
+	return keyID, nil
+}
+
+type KeyResponse struct {
```
abhinavdahiya, line 378:
```suggestion
type keyResponse struct {
```

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+		}
+
+		pubKey, err := x509.ParsePKIXPublicKey(block.Bytes)
+		if err != nil {
+			return response, err
+		}
+		switch pubKey.(type) {
+		case *rsa.PublicKey:
+		default:
+			return response, errors.New("Public key was not RSA")
```
abhinavdahiya, line 398:
```suggestion
			return response, errors.New("public key was not RSA but %T", pubKey)
```

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+	Keys []jose.JSONWebKey `json:"keys"`
+}
+
+func encodeKeys(pemKeys [][]byte) ([]byte, error) {
+	var keys []jose.JSONWebKey
+	var response []byte
+	for _, key := range pemKeys {
+		block, _ := pem.Decode(key)
+		if block == nil {
+			return response, errors.New("Error decoding PEM key")
```
abhinavdahiya, line 388:
```suggestion
			return response, errors.New("error decoding PEM key")
```

abhinavdahiya, line 388: do we want to return return response, or should we explicity return nil with err

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -0,0 +1,424 @@
...
+			return response, errors.New("Public key was not RSA")
+		}
+
+		var alg jose.SignatureAlgorithm
+		switch pubKey.(type) {
+		case *rsa.PublicKey:
+			alg = jose.RS256
+		default:
+			return response, fmt.Errorf("invalid public key type %T, must be *rsa.PrivateKey", pubKey)
+		}
```
abhinavdahiya, line 407: this is repeat of https://github.com/openshift/cloud-credential-operator/pull/195/commits/985fc52ed17d4e37214df7acd57369dbee35429e#diff-8db9da33a38aa7a3147bc262c4e77abaR395-R399

pkg/operator/oidcdiscoveryendpoint/controller.go
abhinavdahiya: see https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/ the style you are using will be deprecated in Sep 2020

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: this was a little convoluted but in `ErrCodeBucketAlreadyOwnedByYou` case does not return so the `else` is needed

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: yes, moving to Debug

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: `response` is not a pointer atm

sjenning:
latest push incorporates much of @abhinavdahiya smaller changes.  still looking at moving from a CREATE to a GET, then optional CREATE if it doesn't exist.  Also making the discovery json template a bindata asset.

fyi, release-4.6 has branched and master is unfrozen now

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: the reconciliation happens very infrequently.  the only thing that can trigger it is a change to the service account token signer (which @marun informs me doesn't rotate on its own at any interval).  even if we were to get in a failure loop, the retry has backoff so we shouldn't cause any issues wrt rate limiting / throttling

pkg/operator/oidcdiscoveryendpoint/controller.go
sjenning: my reasoning was that it was not a kube resources and it was small and easier to load and access/understand without layers of indirection.

joelddiaz:
this gets an lgtm from me. should get a second approval for significant changes like this, so someone else can put this on the merge queue

pkg/operator/oidcdiscoveryendpoint/controller.go
abhinavdahiya: if you can add a note to the commit message on this, it will help in easy discoverability.  github comments aren't great after the fact.

pkg/operator/oidcdiscoveryendpoint/controller.go
abhinavdahiya: if you can add a note to the commit message on this, it will help in easy discoverability.  github comments aren't great after the fact.

PR #198, opened 2020-06-01
Ignore all idea files

akhil-rane:


staebler:
Is this something that would be better left to the user that is using an IntelliJ IDE to set up locally for themselves?

akhil-rane:
> Is this something that would be better left to the user that is using an IntelliJ IDE to set up locally for themselves?

May be. Or it can just be a guard to prevent against accidental push (for common Go IDEs)

PR #200, opened 2020-06-02
Fix credentials request cleanup failure after re-enabling CCO

akhil-rane:
Problem: Deleting the credentials request when CCO was disabled did not work
as expected, but the credentials request couldn't be deleted even when CCO was
re-enabled.

Solution: Creating a watch for cloud-credential-operator-config configmap to
monitor when CCO is re-enabled. Once CCO is re-enabled we reconcile all the
credentials request whose DeletionTimestamp is set.

Jira: https://issues.redhat.com/browse/CO-970

### DIFF
--- a/pkg/operator/credentialsrequest/credentialsrequest_controller.go
+++ b/pkg/operator/credentialsrequest/credentialsrequest_controller.go
@@ -240,7 +240,59 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Monitor the cloud-credential-operator-config configmap to check if operator is re-enabled.
+	// Check if attempts were made to delete the credentials requests when the operator was disabled.
+	// We have to have to reconcile all those credential requests.
+	// allCredRequestsMapFn simply looks up all CredentialsRequests and requests they be reconciled.
+	allCredRequestsMapFn := handler.ToRequestsFunc(
+		func(a handler.MapObject) []reconcile.Request {
+			log.Info("requeueing all CredentialsRequests")
+			crs := &minterv1.CredentialsRequestList{}
+			err := mgr.GetClient().List(context.TODO(), crs)
+			var requests []reconcile.Request
+			if err != nil {
+				log.WithError(err).Error("error listing all cred requests for requeue")
+				return requests
+			}
+			for _, cr := range crs.Items {
+				requests = append(requests, reconcile.Request{
+					NamespacedName: types.NamespacedName{
+						Name:      cr.Name,
+						Namespace: cr.Namespace,
+					},
+				})
+			}
+			return requests
+		})
+
+	// Check if operator is re-enabled in cloud-credential-operator-config configmap
+	configMapPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return isCloudCredOperatorConfigMap(e.MetaNew)
+		},
+		CreateFunc: func(e event.CreateEvent) bool {
+			return isCloudCredOperatorConfigMap(e.Meta)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return isCloudCredOperatorConfigMap(e.Meta)
+		},
+	}
+	err = c.Watch(
+		&source.Kind{Type: &corev1.ConfigMap{}},
+		&handler.EnqueueRequestsFromMapFunc{
+			ToRequests: allCredRequestsMapFn,
+		},
+		configMapPredicate)
+	if err != nil {
+		return err
+	}
+
 	return nil
+}
+
+// isCloudCredOperatorConfigMap returns true if given configmap is cloud-credential-operator-config configmap
+func isCloudCredOperatorConfigMap(cm metav1.Object) bool {
+	return cm.GetName() == operatorconstants.CloudCredOperatorConfigMap && cm.GetNamespace() == minterv1.CloudCredOperatorNamespace
 }
 
 var _ reconcile.Reconciler = &ReconcileCredentialsRequest{}


pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -239,9 +239,75 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Monitor the cloud-credential-operator-config configmap to check if operator is re-enabled.
+	// Check if attempts were made to delete the credentials requests when the operator was disabled.
+	// We have to have to reconcile all those credential requests.
+	configmapMapFn := handler.ToRequestsFunc(
```
dgoodwin, line 245: Kudos Akhil, this is better than what we talked about in scrum yesterday. Why restart the pod when we can just requeue when we detect a change we care about.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -239,9 +239,75 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false
+		},
+	}
+	err = c.Watch(
+		&source.Kind{Type: &corev1.ConfigMap{}},
+		&handler.EnqueueRequestsFromMapFunc{
+			ToRequests: configmapMapFn,
+		},
+		configMapPredicate)
```
dgoodwin, line 285: I was worried we were now watching all ConfigMaps globally and would blow memory up in some clusters, but I can see we already have a configmap_controller that does the same, so our cache should be unchanged with your patch. :+1:

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -239,9 +239,75 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Monitor the cloud-credential-operator-config configmap to check if operator is re-enabled.
+	// Check if attempts were made to delete the credentials requests when the operator was disabled.
+	// We have to have to reconcile all those credential requests.
+	configmapMapFn := handler.ToRequestsFunc(
+		func(a handler.MapObject) []reconcile.Request {
+			// Iterate all CredentialsRequests to determine if any of them have DeletionTimestamp set
+			log.Debug("checking for credentials requests marked for cleanup when operator was disabled")
```
dgoodwin, line 248: Let's make this simpler, just requeue them all, there's no need to only requeue the deleted ones.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -239,9 +239,75 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			crs := &minterv1.CredentialsRequestList{}
+			mgr.GetClient().List(context.TODO(), crs)
+			requests := []reconcile.Request{}
+			for _, cr := range crs.Items {
+				if cr.DeletionTimestamp != nil {
+					log.WithFields(log.Fields{
+						"cr": fmt.Sprintf("%s/%s", cr.Namespace, cr.Name),
+					}).Info("found credentials request marked for cleanup")
```
dgoodwin, line 256: Per above just skip checking if deleted, and log at Info that we're requeuing all cred requests because the config changed.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -239,9 +239,75 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	}
+
 	return nil
 }
 
+func isCloudCredOperatorRenabled(e event.UpdateEvent) bool {
+	if e.MetaNew.GetName() != minterv1.CloudCredOperatorConfigMap || e.MetaNew.GetNamespace() != minterv1.CloudCredOperatorNamespace {
+		return false
+	}
+	logger := log.WithField("configmap", minterv1.CloudCredOperatorConfigMap)
```
dgoodwin, line 297: I'd be tempted to go simpler here as well, if we saw an update to the correct configmap, requeue everything regardless if we're going from disabled to not disabled. This is to be generally consistent that we don't really care about the details of the event that triggered us, we just treat it as a point in time to go check observed state and try to reach desired state. This would mean there's no need to introspect the event and check specifically if we're going from disabled to enabled.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -239,9 +239,75 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		},
+		configMapPredicate)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
+func isCloudCredOperatorRenabled(e event.UpdateEvent) bool {
```
dgoodwin, line 293: If you agree with the below approach this function could just become (isCloudCredOperatorConfigMap)

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -239,9 +239,75 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 
+	// Monitor the cloud-credential-operator-config configmap to check if operator is re-enabled.
+	// Check if attempts were made to delete the credentials requests when the operator was disabled.
+	// We have to have to reconcile all those credential requests.
+	configmapMapFn := handler.ToRequestsFunc(
+		func(a handler.MapObject) []reconcile.Request {
+			// Iterate all CredentialsRequests to determine if any of them have DeletionTimestamp set
+			log.Debug("checking for credentials requests marked for cleanup when operator was disabled")
+			crs := &minterv1.CredentialsRequestList{}
+			mgr.GetClient().List(context.TODO(), crs)
```
dgoodwin, line 250: In borrowing your code for something I'm working on this morning my editor informed me you've got an unchecked error here. :)

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -239,9 +239,75 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			requests := []reconcile.Request{}
```
dgoodwin, line 251: This should also be "var requests []reconcile.Request", according to goland.

dgoodwin: I ended up with 

```go
	// allCredRequestsMapFn simply looks up all CredentialsRequests and requests they be reconciled.
	allCredRequestsMapFn := handler.ToRequestsFunc(
		func(a handler.MapObject) []reconcile.Request {
			log.Info("requeueing all CredentialsRequests")
			crs := &minterv1.CredentialsRequestList{}
			err := mgr.GetClient().List(context.TODO(), crs)
			var requests []reconcile.Request
			if err != nil {
				log.WithError(err).Error("error listing all cred requests for requeue")
				return requests
			}
			for _, cr := range crs.Items {
					requests = append(requests, reconcile.Request{
						NamespacedName: types.NamespacedName{
							Name:      cr.Name,
							Namespace: cr.Namespace,
						},
					})
			}
			return requests
		})
```

And I think we can just re-use the same map function, we both just want to requeue everything.

akhil-rane:
@dgoodwin theses changes also fix https://issues.redhat.com/browse/CO-971

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -239,9 +239,61 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			return requests
+		})
+
+	// Check if operator is re-enabled in cloud-credential-operator-config configmap
+	configMapPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return isCloudCredOperatorConfigMap(e.MetaNew)
+		},
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
```
dgoodwin, line 273: I would return isCloudCredOperatorConfigMap from create and delete as well, in each case the state we're in could have changed, so it makes sense to reprocess all CRs.

PR #203, opened 2020-06-05
report metric on mode of operator for CCO

joelddiaz:
Report with a 1 or 0 whether a specific credentials reconcile mode is running from Mint, Passthrough, CredsRemoved, Manual, Degraded, and Unknown.

Rework test cases to excercise more of metrics controller, and read the actual metrics recorded when making assertions.


pkg/operator/metrics/metrics_test.go
```diff
@@ -100,29 +196,138 @@ func TestCredentialsRequests(t *testing.T) {
...
+			test.validate(t)
+		})
+	}
+
+}
+
+func metricAssert(t *testing.T, metric *prometheus.GaugeVec, label string, expectedValue int) {
+	val, err := metric.GetMetricWithLabelValues(label)
+	assert.NoError(t, err, "unexpected error getting stored metric value")
+	assert.Equal(t, float64(expectedValue), promtest.ToFloat64(val))
```
dgoodwin, line 234: Best practice for metrics is that you should treat them like debug log statements, which would seldom be unit tested unless they were integral to the application. I think the mentality is that sprinkling instrumentation around should be quick and easy, which it won't if you need to cover it all with unit tests.

I think there could also be issues with the way they're exposed as something like a global singleton where state could leak from test to test. Not sure if the promtest lib you vendored helps there but regardless.

In our case we have some logic in play, but we can unit test that at the accumulator or calculator level instead of going right to prometheus.

pkg/operator/metrics/metrics.go
```diff
@@ -35,6 +42,12 @@ var (
 		Help: "Credentials requests with asserted conditions.",
 	}, []string{"condition"})
 
+	// Report on the mode CCO is operating under
+	metricCredentialsMode = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "cco_credentials_mode",
+		Help: "Track current mode the cloud-credentials-operator is functioning under.",
+	}, []string{"mode"})
```
dgoodwin, line 49: This seems a sufficient way to model it, if we get into more complicated scenarios in the future where more than one mode could be true, we can just report multiple 1's.

pkg/operator/metrics/metrics_test.go
joelddiaz: Yes, I do think there is the potential issue with state from test to test, but since the metric collection is written in a way to clear metrics before calculation, we are protected from this (as the code is currently written). I thought it was nice to get the extra code coverage, but if you think it's not worth it, I can switch back to having the tests wrap only the accumulator.

PR #204, opened 2020-06-08
restore DeepCopy generation

joelddiaz:
Bring back hooks to update/verify our generated DeepCopy files.

The DeepCopy generator does not work with the current AWSProviderSpec.StatementEntries[].PolicyCondition field being a map[string]map[string]interface{}. Annotate those fields so that there is no attempt to generate the DeepCopy methods, and provide custom DeepCopy methods where needed.


pkg/apis/cloudcredential/v1/aws_types.go
```diff
@@ -39,10 +40,10 @@ type StatementEntry struct {
 	// Resource specifies the object(s) this statement should apply to. (or "*" for all)
 	Resource string `json:"resource"`
 	// PolicyCondition specifies under which condition StatementEntry will apply
-	PolicyCondition IAMPolicyCondition `json:"policyCondition,omitempty"`
+	PolicyCondition *runtime.RawExtension `json:"policyCondition,omitempty"`
```
dgoodwin, line 43: Does this need to be a pointer? It feels like this could trigger the situation deads warned about where going to a pointer means old clients could suddenly start getting nil's, and break list+watch hard.

pkg/aws/actuator/actuator.go
```diff
@@ -882,12 +882,22 @@ func (a *AWSActuator) getDesiredUserPolicy(entries []minterv1.StatementEntry, us
...
+			// We're just passing through the policyCondition content right through to the AWS API
+			// so convert the rawextension content
+			var out interface{}
+			err := json.Unmarshal(se.PolicyCondition.Raw, &out)
+			if err != nil {
+				return "", fmt.Errorf("error unmarshaling PolicyCondition: %v", err)
+			}
+			statement.Condition = out
+		}
+		policyDoc.Statement = append(policyDoc.Statement, statement)
```
dgoodwin, line 900: Were you able to manually test with this PolicyCondition actually in use?

joelddiaz: Yes, with and without it. That's how I figured that I needed to annotate the StatementEntry struct to put the 'omitempty' (otherwise the JSON sent to AWS when there was no policy always included "Condition": null which would cause the AWS API calls to fail).

FWIW, this is a simple example:
```
spec:
  secretRef:
    name: my-cred-request-secret
  providerSpec:
    kind: AWSProviderSpec
    statementEntries:
    - effect: Allow
      action:
      - s3:CreateBucket
      - s3:DeleteBucket
      resource: "*"
      policyCondition:
        StringEquals:
          aws:userid: "testuser"
```

joelddiaz:
Here's a summary of where we stand with runtime.RawExtension being a pointer or not-a-pointer and comparing that to what the object definition is today.

Today, the object definition has the PolicyCondition field as a `map[string]map[string]interface{}`. I can't show an example of a CRD with that for two reasons. 1) The PolicyCondition field is buried under a *runtime.RawExtension 2) If I bring the PolicyCondition out of the RawExtension (for demonstration purposes), the CRD cannot be generated because of the `interface{}` (the primary motivator for this PR!).

Given that, I have created a new field (`TestField`) under CredentialsRequestSpec with type `map[string]int`. Generating the CRD for that results in:

```
testField:
  type: object
  additionalProperties:
    type: integer
```

Changing the `TestField` to `*runtime.RawExtension`:
```
testField:
  type: object
```

Changing it to a non-pointer `runtime.RawExtension` has no change in the generated CRD.

It is not obvious to me that there is a benefit (or even a change) in choosing between `runtime.RawExtension` and `*runtime.RawExtension`. I'd lean to the pointer-version simply because `CredentialsRequestSpec.ProviderSpec` is a pointer today.

Do we have any API experts who can review/comment on this proposed change?

csrwng:
@joelddiaz fwiw the machine-api also uses a pointer to runtime.RawExtension (so I was wrong thinking it wasn't):
https://github.com/openshift/machine-api-operator/blob/d5a738f9925fbf9ce64b5c20cf0a2058b09f9243/pkg/apis/machine/v1beta1/common_types.go#L35

joelddiaz:
@deads2k as one our resident "API people", can you PTAL and see if there are any concerns with this migration from a field which is presently listed as type `interface{}` (living under a `*runtime.RawExtension`) to having it also be a `*runtime.RawExtension` (also living under the "parent" RawExtension as the json-location of the field hasn't changed)?

Our primary concern here is that we don't make any breaking API changes here (specifically the change to `pkg/apis/cloudcredential/v1/aws_types.go`).

dgoodwin:
Good solve. Looks good to me, but will let @csrwng do the honors once he's had a chance to look.

hack/update-codegen.sh
```diff
@@ -0,0 +1,47 @@
...
+SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
+REPO_FULL_PATH=$(realpath ${SCRIPT_ROOT})
+cd ${REPO_FULL_PATH}
+
+CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../../../k8s.io/code-generator)}
+
+verify="${VERIFY:-}"
+
+valid_gopath=$(realpath $REPO_FULL_PATH/../../../..)
+if [[ "$(realpath ${valid_gopath}/src/github.com/openshift/cloud-credential-operator)" == "${REPO_FULL_PATH}" ]]; then
```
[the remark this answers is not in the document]
joelddiaz: It was an optimization suggested by @staebler so that we wouldn't have to wait for the various pieces to be downloaded to the GOPATH if your usable GOPATH already had them in there. Works out that the TMP_DIR holds around 174MB of data on disk.

I'll un-WIP and squash some things up and reduce the verbosity of the deepcopy generation in a moment.

PR #205, opened 2020-06-10
unify constants

joelddiaz:
bring everything under pkg/operator/constants (except for the list of FailureConditionTypes which will live with where the conditions are defined).


dgoodwin:
Probably waited long enough, unhold sounds ok to me if it's ok with you.

PR #206, opened 2020-06-10
Upgradable false when parent creds removed.

dgoodwin:
Enhance safety by making sure the recently documented root cred removal process marks upgradable false until the user restores the credentials.

We also now reconcile all cred requests on any change to a secret that looks like one of our supported cluster parent cred secrets.

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -28,6 +28,7 @@ import (
 	ccaws "github.com/openshift/cloud-credential-operator/pkg/aws"
 	minteraws "github.com/openshift/cloud-credential-operator/pkg/aws"
 	actuatoriface "github.com/openshift/cloud-credential-operator/pkg/operator/credentialsrequest/actuator"
+	"github.com/openshift/cloud-credential-operator/pkg/operator/credentialsrequest/constants"
 	awsannotator "github.com/openshift/cloud-credential-operator/pkg/operator/secretannotator/aws"
 	annotatorconst "github.com/openshift/cloud-credential-operator/pkg/operator/secretannotator/constants"
 	"github.com/openshift/cloud-credential-operator/pkg/operator/utils"
@@ -49,13 +50,10 @@ import (
 )
 
 const (
-	rootAWSCredsSecretNamespace = "kube-system"
-	rootAWSCredsSecret          = "aws-creds"
-	roAWSCredsSecretNamespace   = "openshift-cloud-credential-operator"
-	roAWSCredsSecret            = "cloud-credential-operator-iam-ro-creds"
-	clusterConfigNamespace      = "kube-system"
-	openshiftClusterIDKey       = "openshiftClusterID"
-	clusterVersionObjectName    = "version"
+	roAWSCredsSecretNamespace = "openshift-cloud-credential-operator"
+	roAWSCredsSecret          = "cloud-credential-operator-iam-ro-creds"
+	openshiftClusterIDKey     = "openshiftClusterID"
+	clusterVersionObjectName  = "version"
 )
 
 var _ actuatoriface.Actuator = (*AWSActuator)(nil)
@@ -293,9 +291,7 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
 			ErrReason: minterv1.InsufficientCloudCredentials,
 			Message:   msg,
 		}
-	}
-
-	if cloudCredsSecret.Annotations[annotatorconst.AnnotationKey] == annotatorconst.PassthroughAnnotation {
+	} else if cloudCredsSecret.Annotations[annotatorconst.AnnotationKey] == annotatorconst.PassthroughAnnotation {
 		logger.Debugf("provisioning with passthrough")
 		err := a.syncPassthrough(ctx, cr, cloudCredsSecret, logger)
 		if err != nil {
@@ -312,6 +308,8 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
 				Message:   fmt.Sprintf("%v: %v", msg, err),
 			}
 		}
+	} else {
+		logger.Infof("unknown or missing %s annotation on admin credentials Secret, skipping reconcile", annotatorconst.AnnotationKey)
 	}
 
 	return nil
@@ -399,7 +397,7 @@ func (a *AWSActuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequ
 			case iam.ErrCodeNoSuchEntityException:
 				logger.WithField("userName", awsStatus.User).Debug("user does not exist, creating")
 				if rootAWSClient == nil {
-					return fmt.Errorf("no root AWS client available, cred secret may not exist: %s/%s", rootAWSCredsSecretNamespace, rootAWSCredsSecret)
+					return fmt.Errorf("no root AWS client available, cred secret may not exist: %s/%s", constants.KubeSystemNS, annotatorconst.AWSCloudCredSecretName)
 				}
 
 				createOut, err := a.createUser(logger, rootAWSClient, awsStatus.User)
@@ -428,7 +426,7 @@ func (a *AWSActuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequ
 	// Check if the user has the expected tags:
 	if !userHasExpectedTags(logger, userOut, infraName, string(clusterUUID)) {
 		if rootAWSClient == nil {
-			return fmt.Errorf("no root AWS client available, cred secret may not exist: %s/%s", rootAWSCredsSecretNamespace, rootAWSCredsSecret)
+			return fmt.Errorf("no root AWS client available, cred secret may not exist: %s/%s", constants.KubeSystemNS, annotatorconst.AWSCloudCredSecretName)
 		}
 
 		err = a.tagUser(logger, rootAWSClient, awsStatus.User, infraName, string(clusterUUID))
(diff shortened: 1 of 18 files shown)


pkg/operator/awspodidentity/controller.go
```diff
@@ -109,7 +109,8 @@ func Add(mgr manager.Manager, kubeconfig string) error {
 	logger := log.WithFields(log.Fields{"controller": controllerName})
 	imagePullSpec := os.Getenv("AWS_POD_IDENTITY_WEBHOOK_IMAGE")
 	if len(imagePullSpec) == 0 {
-		return errors.New("AWS_POD_IDENTITY_WEBHOOK_IMAGE is not set")
+		logger.Warn("AWS_POD_IDENTITY_WEBHOOK_IMAGE is not set, skipping controller")
```
akhil-rane, line 112: Do we need doc update here saying we need to set 'AWS_POD_IDENTITY_WEBHOOK_IMAGE' environment variable while running from source?

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -178,13 +179,57 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			return isAdminCredSecret(e.Meta.GetNamespace(), e.Meta.GetName())
+		},
+	}
+	// Watch Secrets and reconcile if we see an event for an admin credential secret in kube-system.
+	err = c.Watch(
+		&source.Kind{Type: &corev1.Secret{}},
+		&handler.EnqueueRequestsFromMapFunc{
+			ToRequests: allCredRequestsMapFn,
+		},
+		adminCredSecretPredicate)
```
akhil-rane, line 228: Not sure about this but is there a need to reconcile all the credentials requests here? If suppose only AWS admin cred has been updated why do we need to reconcile GCP cred requests?

pkg/aws/actuator/actuator.go
```diff
@@ -896,9 +896,14 @@ func (a *AWSActuator) getDesiredUserPolicy(entries []minterv1.StatementEntry, us
...
 
+// GetParentCredSecretLocation returns the namespace and name where the parent credentials secret is stored.
+func (a *AWSActuator) GetParentCredSecretLocation() types.NamespacedName {
+	return types.NamespacedName{Namespace: constants.KubeSystemNS, Name: annotatorconst.AWSCloudCredSecretName}
+}
+
 func (a *AWSActuator) getCloudCredentialsSecret(ctx context.Context, logger log.FieldLogger) (*corev1.Secret, error) {
 	cloudCredSecret := &corev1.Secret{}
-	if err := a.Client.Get(ctx, types.NamespacedName{Name: rootAWSCredsSecret, Namespace: rootAWSCredsSecretNamespace}, cloudCredSecret); err != nil {
+	if err := a.Client.Get(ctx, types.NamespacedName{Name: annotatorconst.AWSCloudCredSecretName, Namespace: constants.KubeSystemNS}, cloudCredSecret); err != nil {
```
akhil-rane, line 906: instead of 'types.NamespacedName{Name: annotatorconst.AWSCloudCredSecretName, Namespace: constants.KubeSystemNS}' can we use a.GetParentCredSecretLocation()?

pkg/operator/credentialsrequest/status.go
```diff
@@ -263,6 +281,15 @@ func computeStatusConditions(
 		Status: configv1.ConditionTrue,
 		Type:   configv1.OperatorUpgradeable,
 	}
+	// If the operator is disabled, and we do not have a parent cred in kube-system, we are not
```
akhil-rane, line 284: Did you mean 'If the operator is not disabled' here?

pkg/aws/actuator/actuator.go
```diff
@@ -312,6 +310,8 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
 				Message:   fmt.Sprintf("%v: %v", msg, err),
 			}
 		}
+	} else {
+		logger.Infof("unknown or missing %s annotation on admin credentials Secret, skipping reconcile", annotatorconst.AnnotationKey)
```
joelddiaz, line 314: Should we set a condition indicating that the CredentialsRequest object wasn't able to be reconciled b/c of this?
Specifically, I'm thinking that it is precently provisioned (status.provisioned=true), and then we hit this else clause. How is this communicated (outside of the log message)?

pkg/gcp/actuator/actuator.go
```diff
@@ -555,10 +555,15 @@ func (a *Actuator) updateProviderStatus(ctx context.Context, logger log.FieldLog
 	return nil
 }
 
+// GetParentCredSecretLocation returns the namespace and name where the parent credentials secret is stored.
+func (a *Actuator) GetParentCredSecretLocation() types.NamespacedName {
+	return types.NamespacedName{Namespace: constants.KubeSystemNS, Name: annotatorconst.GCPCloudCredSecretName}
```
joelddiaz, line 560: global var?

pkg/openstack/actuator.go
```diff
@@ -209,11 +209,16 @@ func (a *OpenStackActuator) loadExistingSecret(cr *minterv1.CredentialsRequest)
 	return existingSecret, nil
 }
 
+// GetParentCredSecretLocation returns the namespace and name where the parent credentials secret is stored.
+func (a *OpenStackActuator) GetParentCredSecretLocation() types.NamespacedName {
+	return types.NamespacedName{Namespace: crconst.KubeSystemNS, Name: constants.OpenStackCloudCredsSecretName}
```
joelddiaz, line 214: global var?

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -178,13 +179,57 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 
+	// allCredRequestsMapFn simply looks up all CredentialsRequests and requests they be reconciled.
+	allCredRequestsMapFn := handler.ToRequestsFunc(
+		func(a handler.MapObject) []reconcile.Request {
+			log.Info("requeueing all CredentialsRequests")
+			crs := &minterv1.CredentialsRequestList{}
+			err := mgr.GetClient().List(context.TODO(), crs)
+			var requests []reconcile.Request
+			if err != nil {
+				log.WithError(err).Error("error listing all cred requests for requeue")
```
joelddiaz, line 197: This means that we never force the reconcile of all the CredentialsRequests?
Should we exit(1) here to force the Pod to restart?

pkg/operator/awspodidentity/controller.go
dgoodwin: I'm not sure just yet, I would like this to come up smoothly without having to do anything on 4.4, we probably need a separate PR to make that happen, there's an additional problem area I had to hack around to get it working.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
dgoodwin: It's cleaner IMO to just reconcile everything, the ones that do not match your cloud provider are currently a no-op, so there's virtually no cost to doing so. In the future we may have scenarios where CCO is managing creds for all clusters in Hive from the Hive cluster, and in this case we'd want to reconcile across clouds.

pkg/aws/actuator/actuator.go
dgoodwin: Good catch, I didn't think of going back to this after adding the actuator method.

pkg/operator/credentialsrequest/status.go
dgoodwin: Another good catch. :)

pkg/aws/actuator/actuator.go
dgoodwin: Feels like a YAGNI to me, maybe if we start hitting problems with it but I find setting and clearing conditions a bit cumbersome and not necessary to do for edge cases we don't expect to have any long lasting impact. In this case we'd be waiting for secretannotator to do it's work, which should not take long and the update should requeue everything. If this were a state we could be in for more than a second or two I'd say yes, but otherwise it feels unnecessary to me.

pkg/gcp/actuator/actuator.go
dgoodwin: Can you clarify? I'm not quite sure of the benefit there.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
dgoodwin: If the admin secret or config map changed yes, we would not reconcile them all, but that is the state we've been in for several releases. Exiting the pod seems like it could be bad precedent, I'm not sure what should be done in a map func when there's an error.

pkg/gcp/actuator/actuator.go
joelddiaz: Ignore this. This was a set of comments I was making before I saw that you extended the actuator API to include GetParentCredSecretLocation(), and this must have slipped through as I removed those comments.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
joelddiaz: Yeah, the inability to return an error really complicates things here. We're probably okay as things will eventually settle to the desired state as the objects get re-reconciled on whatever schedule we have it set to.

pkg/gcp/actuator/actuator.go
dgoodwin: Gotcha, thanks.

pkg/operator/awspodidentity/controller.go
sjenning: i'll open the PR for this targeted change that also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1846200

PR #208, opened 2020-06-11
Bug 1809194: expose metrics over https

joelddiaz:
Add kube-rbac-proxy container to CCO deployment listening on :8443 and forwarding traffic to :2112 (the CCO container's metrics port).
New ConfigMap with configuration for the kube-rbac-proxy container.
Update cco-metrics service to request certs to be created (for use by kube-rbac-proxy container).
Update ServiceMonitor to tell Prometheus to scrape using https.
Grant Prometheus role permissions for fetching metrics.
Update role permissions so that kube-rbac-proxy container can perform its tasks.

Also, remove unused port definition in CCO container definition (webook-server port number 9876).

### DIFF
--- a/manifests/0000_90_cloud-credential-operator_01_prometheusrole.yaml
+++ b/manifests/0000_90_cloud-credential-operator_01_prometheusrole.yaml
@@ -7,6 +7,12 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - namespace/metrics
+  verbs:
+  - get
+- apiGroups:
+  - ""
+  resources:
   - services
   - endpoints
   - pods
--- a/manifests/03-deployment.yaml
+++ b/manifests/03-deployment.yaml
@@ -22,6 +22,32 @@ spec:
     spec:
       containers:
       - args:
+        - --secure-listen-address=0.0.0.0:8443
+        - --upstream=http://127.0.0.1:2112/
+        - --tls-cert-file=/etc/tls/private/tls.crt
+        - --tls-private-key-file=/etc/tls/private/tls.key
+        - --config-file=/etc/kube-rbac-proxy/config-file.yaml
+        - --logtostderr=true
+        - -v=10
+        image: quay.io/openshift/origin-kube-rbac-proxy:latest
+        imagePullPolicy: IfNotPresent
+        name: kube-rbac-proxy
+        ports:
+        - containerPort: 8443
+          name: metrics
+          protocol: TCP
+        resources:
+          requests:
+            memory: 20Mi
+            cpu: 10m
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
+        volumeMounts:
+        - mountPath: /etc/kube-rbac-proxy
+          name: auth-proxy-config
+        - mountPath: /etc/tls/private
+          name: cloud-credential-operator-serving-cert
+      - args:
         - |
           if [ -s /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem ]; then
               echo "Copying system trust bundle"
@@ -39,10 +65,6 @@ spec:
         image: quay.io/openshift/origin-cloud-credential-operator:latest
         imagePullPolicy: IfNotPresent
         name: cloud-credential-operator
-        ports:
-        - containerPort: 9876
-          name: webhook-server
-          protocol: TCP
         resources:
           requests:
             cpu: 10m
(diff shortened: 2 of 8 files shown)


manifests/0000_90_cloud-credential-operator_01_prometheusrole.yaml
```diff
@@ -4,6 +4,12 @@ metadata:
   name: prometheus-k8s
   namespace: openshift-cloud-credential-operator
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - namespace/metrics
```
dgoodwin, line 10: Is this some special kube resource?

manifests/03-deployment.yaml
```diff
@@ -21,6 +21,32 @@ spec:
...
       containers:
+      - args:
+        - --secure-listen-address=0.0.0.0:8443
+        - --upstream=http://127.0.0.1:2112/
+        - --tls-cert-file=/etc/tls/private/tls.crt
+        - --tls-private-key-file=/etc/tls/private/tls.key
+        - --config-file=/etc/kube-rbac-proxy/config-file.yaml
+        - --logtostderr=true
+        - -v=10
+        image: quay.io/openshift/origin-kube-rbac-proxy:4.2.0
```
dgoodwin, line 32: One of my overall questions is where this approach came from, and here specifically is it ok to use 4.2 version hard coded? We should comment with why that's the case. 

Or is this one that is automatically replaced?

manifests/0000_90_cloud-credential-operator_01_prometheusrole.yaml
joelddiaz: Not sure. I was using the cluster-machine-approver repo as an example (b/c they also use controller-runtime) and they had it https://github.com/openshift/cluster-machine-approver/blob/master/manifests/0000_90_cluster-machine-approver_01_prometheusrole.yaml#L7-L12

manifests/03-deployment.yaml
joelddiaz: Again, using cluster-machine-approver as an example: https://github.com/openshift/cluster-machine-approver/blob/master/manifests/04-deployment.yaml#L34

FWIW, I think I looked at a running cluster post-install and it had this same unmodified image (I'd have to double check to make sure).

dgoodwin: There is a kube-rbac-proxy image in the release payload. I think @sjenning  did something similar with the AWS work for pod identity where he was able to have the image replaced on the fly. Could you check if this is something we could do here to get the latest image.

joelddiaz: It appears that the image is being replaced even with this :4.2.0 tag. Here is the running machine-approver Pod snippet:
```
  spec:
    containers:
    - args:
      - --secure-listen-address=0.0.0.0:9192
      - --upstream=http://127.0.0.1:9191/
      - --tls-cert-file=/etc/tls/private/tls.crt
      - --tls-private-key-file=/etc/tls/private/tls.key
      - --config-file=/etc/kube-rbac-proxy/config-file.yaml
      - --logtostderr=true
      - --v=10
      image: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:f7c7c750e954daef37c1271c2ad4ea5e2506638a7465cb251ee5b958ce290658
      imagePullPolicy: IfNotPresent
      name: kube-rbac-proxy
```
But it does seems more correct to use the :latest tag as found in https://github.com/openshift/cluster-ingress-operator/blob/master/manifests/02-deployment.yaml#L68 and https://github.com/openshift/cluster-dns-operator/blob/master/manifests/0000_70_dns-operator_02-deployment.yaml#L42 . Updated the PR to use :latest

manifests/0000_90_cloud-credential-operator_01_prometheusrole.yaml
joelddiaz: I don't see this in other repos, so I've removed it from the PR.

https://github.com/openshift/cluster-ingress-operator/blob/master/manifests/0000_90_ingress-operator_00_prometheusrole.yaml
https://github.com/openshift/cluster-dns-operator/blob/master/manifests/0000_90_dns-operator_00_prometheusrole.yaml

joelddiaz: So I tested this by removing this from the role, but you need to wait for the token to expire before the role changes take effect. It turns out that this is needed as the kube-rbac-proxy pod will show errors otherwise:

```
I0617 23:24:14.610086       1 proxy.go:102] Forbidden (user=system:serviceaccount:openshift-monitoring:prometheus-k8s, verb=get, resource=namespace, subresource=metrics)
```

What isn't clear to me is why other components do not need this permission, and I'll have to look into this.

joelddiaz: After discussion with the monitoring team, they asked for a BZ to investigate why this permissions is needed, but it seems like we're okay to go with the current role https://bugzilla.redhat.com/show_bug.cgi?id=1848600

PR #209, opened 2020-06-15
Bug 1846200: make AWS_POD_IDENTITY_WEBHOOK_IMAGE not set non-fatal

sjenning:
@dgoodwin @abhinavdahiya @wking 

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1846200

This changes the behavior from fatal termination of the whole operator when `AWS_POD_IDENTITY_WEBHOOK_IMAGE` is not set, to just disabling the `podidentitywebhook` controller within the operator.

### DIFF
--- a/pkg/operator/awspodidentity/controller.go
+++ b/pkg/operator/awspodidentity/controller.go
@@ -109,7 +109,7 @@ func Add(mgr manager.Manager, kubeconfig string) error {
 	logger := log.WithFields(log.Fields{"controller": controllerName})
 	imagePullSpec := os.Getenv("AWS_POD_IDENTITY_WEBHOOK_IMAGE")
 	if len(imagePullSpec) == 0 {
-		logger.Warn("AWS_POD_IDENTITY_WEBHOOK_IMAGE is not set, skipping controller")
+		logger.Warn("AWS_POD_IDENTITY_WEBHOOK_IMAGE is not set, AWS pod identity webhook will not be deployed")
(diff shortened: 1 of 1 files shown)


dgoodwin:
Tested on 4.4 and we can now startup cleanly there which is great.
However the oidcdiscoveryendpoint controller goes into an error loop: 

```
DEBU[0030] reconciling cluster Authentication ServiceAccountIssuer  controller=oidcdiscoveryendpoint
DEBU[0030] Authentication ServiceAccountIssuer already specified  controller=oidcdiscoveryendpoint
DEBU[0030] reconciling S3 resources                      controller=oidcdiscoveryendpoint
ERRO[0030] failed reconciling S3 resources               controller=oidcdiscoveryendpoint error="secrets \"cloud-credential-operator-s3-creds\" not found"
```

Could/should we lump that into this fix?

sjenning:
@dgoodwin that error should resolve when the limited `CredentialRequest` is processed.  is it not resolving?
https://github.com/openshift/cloud-credential-operator/blob/master/manifests/05-s3-credentialsrequest.yaml

sjenning:
The s3 limited cred that controller uses was added to `/manifest` and requires a new release image to deploy automatically.  If you want to test on an older cluster, you need to manually create https://github.com/openshift/cloud-credential-operator/blob/master/manifests/05-s3-credentialsrequest.yaml

PR #210, opened 2020-06-16
WIP: basic implementation of supporting applying profile patches to manifests

joelddiaz:
1) Make a patch for a profile under manifests/profile-patches/$PROFILE/ (name it with ending .json.patch or .yaml.patch)
2) Run hacks/update-profiles.sh and watch it generate manifests/$SOURCE_FILE-$PROFILE.{yaml|json}
3) Verify with hacks/verify-profiles.sh

This is a POC for possible approach to addressing the Profiles being proposed in https://github.com/openshift/enhancements/pull/200. Not intended for merging.

### DIFF
--- a/hack/verify-profiles.sh
+++ b/hack/verify-profiles.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -e
+
+SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
+
+${SCRIPT_ROOT}/hack/update-profiles.sh --verify
--- a/hack/update-profiles.sh
+++ b/hack/update-profiles.sh
@@ -1,0 +1,132 @@
+#!/usr/bin/env bash
+
+set -e
+
+SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
+FULL_REPO_ROOT=$(realpath ${SCRIPT_ROOT})
+
+YAML_PATCH_BIN=${SCRIPT_ROOT}/_output/tools/bin/yaml-patch
+JSON_PATCH_BIN=${SCRIPT_ROOT}/_output/tools/bin/json-patch
+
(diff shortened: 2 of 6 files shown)


joelddiaz:
@csrwng here's what i had in mind after our chat earlier today

csrwng:
@joelddiaz this looks awesome. 
The one thing I'd change is that instead of explicitly including the patch files in the script, I would go by convention and find patch files in a directory to apply to existing manifests. That way the script can be placed in a common library like `library-go` and get all projects to vendor it.

joelddiaz:
@csrwng Ok, I've set up the restrictions that you must place your patches in `./manifests/profile-patches/$PROFILE_NAME`, and the patch file must have the same name as the source file being patched except that it must end in `.patch`. And there's a primitive implementation of verify as well.
WDYT?

hack/verify-profiles.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -e
+
+SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
+
+${SCRIPT_ROOT}/hack/update-profiles.sh --verify
```
wking, line 7: Is this helper a hack to support autocompletion?

joelddiaz: just to make it easy for someone to run the "verify" step without having to know the details of how the update-profiles.sh script works

dgoodwin:
Could you update the commit message on this with some info on what this is intended for, what profiles are in this context, etc.

csrwng:
@dgoodwin this is just meant to be an example and  I don't think we want to actually merge it (at least not yet). If @joelddiaz agrees  we can close it for now and wait for the enhancement to get approved first.

joelddiaz:
Closing as this is just a POC

PR #212, opened 2020-06-19
no need for a non-default kube-rbac-proxy config

joelddiaz:
https://bugzilla.redhat.com/show_bug.cgi?id=1848600#c2 tells us that we do not need a custom config for the kube-rbac-proxy container.

Remove the custom configmap.
Update the Prometheus role to not need the namespace/metric permission.
Update the deployment to not mount the configmap, and run in its default mode (and reduce the verbosity of the container).


dgoodwin:
Much simpler, nice.

/lgtm

Issue #207, opened 2020-06-10
CloudCredentialsOperator Generating tons of logs 


raffamendes:
The CloudCredentialsOperator is generating tons of log entries, after searching in Kibana we're getting   a lot of hits from debug messages, is there a way to configure this log level to be less verbose?
 We're Using OCP 4.3.13

dgoodwin:
Can we get the logfile here? CCO should be dormant after starting up. I would like to know what was generating all the logs and the scale of verbosity we're looking at.

gregsheremeta:
@joelddiaz fixed?

joelddiaz:
Fix made it to the master branch. I don't believe we should expect this to ever make it to 4.3.x given where it is on the planned life cycle https://access.redhat.com/support/policy/updates/openshift#dates

gregsheremeta:
agree. Thanks!

PR #214, opened 2020-06-25
let CCO use the default info log-level

joelddiaz:
instead of forcing debug.

This is to address openshift/cloud-credential-operator#207

But it doesn't address how to actually set it to debug mode if desired.


dgoodwin:
If we needed debug logs we'd ask someone to scale down CVO and modify the deployment?

joelddiaz:
Scale down CVO, then modify the CCO deployment should be all that is necessary (to keep CVO from stomping our Deployment changes).

PR #215, opened 2020-06-29
WIP: really convert interface{} to RawExtension

joelddiaz:
Rather than introduce the API breaking change from `map[string]map[string]interface{}` to `runtime.RawExtension`, convert to `map[string]map[string]runtime.RawExtension`.

Update test cases, and remove manual DeepCopy pieces.

Don't register non-core types into the Scheme.


joelddiaz:
will be taking a different approach to get into openshift/api

joelddiaz:
@sttts WDYT about an approach like this to get rid of the `interface{}` out of the cloud-cred-operator API altogether? 
Specifically the changes here: https://github.com/openshift/cloud-credential-operator/pull/215/commits/61640c5491848aebd1bfabfc284c1f8ca2528b80

IIRC, the concern with this approach was whether the `interface{}` to rawExtension change would be considered non-API-breaking.

sttts:
This sounds like a good solution. Changing the Golang type does not necessarily change the external API. But be careful in testing that old and new object look the same. Note that RawExtension is meant for cases where the values have TypeMeta fields. This is the case for you?

joelddiaz:
> This sounds like a good solution. Changing the Golang type does not necessarily change the external API. But be careful in testing that old and new object look the same. Note that RawExtension is meant for cases where the values have TypeMeta fields. This is the case for you?

@sttts for our use-case we do not have TypeMeta for what is being embedded in this field as it simply a bundle of JSON to be passed as-is to AWS API calls. So I take that to mean that RawExtension wouldn't be appropriate for eliminating the `interface{}`? Is there another approach that can get us out of writing/maintaining custom DeepCopy methods?

PR #216, opened 2020-06-30
OpenStack: Add cloud provider secret key

Fedosin:
Now we include only clouds.yaml key in CCO minted secrets. But starting from OpenShift 4.6 we also need clouds.conf key data for Cinder CSI driver.

Both keys are generated by the installer: https://github.com/openshift/installer/blob/master/data/data/manifests/openshift/cloud-creds-secret.yaml.template#L33-L34

This commit adds clouds.conf key in CCO generated secrets.

### DIFF
--- a/pkg/openstack/actuator.go
+++ b/pkg/openstack/actuator.go
@@ -35,7 +35,8 @@ import (
 )
 
 const (
-	rootOpenStackCredsSecretKey = "clouds.yaml"
+	rootOpenStackCredsSecretKey          = "clouds.yaml"
+	cloudProviderOpenStackCredsSecretKey = "clouds.conf"
 )
 
 type OpenStackActuator struct {
@@ -95,7 +96,24 @@ func (a *OpenStackActuator) sync(ctx context.Context, cr *minterv1.CredentialsRe
 	logger := a.getLogger(cr)
 	logger.Debug("running sync")
 
-	clouds, err := a.getRootCloudCredentialsSecretData(ctx, logger)
+	// Get the base secret
+	cloudCredSecret := &corev1.Secret{}
+	if err := a.Client.Get(ctx, a.GetCredentialsRootSecretLocation(), cloudCredSecret); err != nil {
+		msg := "unable to fetch root cloud cred secret"
+		logger.WithError(err).Error(msg)
+		return &actuatoriface.ActuatorError{
+			ErrReason: minterv1.CredentialsProvisionFailure,
+			Message:   fmt.Sprintf("%v: %v", msg, err),
+		}
+	}
+
+	cloudsYAML, err := a.getCloudCredentialsSecretData(cloudCredSecret, rootOpenStackCredsSecretKey, logger)
+	if err != nil {
+		logger.WithError(err).Error("issue with cloud credentials secret")
+		return err
+	}
+
+	cloudsCONF, err := a.getCloudCredentialsSecretData(cloudCredSecret, cloudProviderOpenStackCredsSecretKey, logger)
 	if err != nil {
 		logger.WithError(err).Error("issue with cloud credentials secret")
 		return err
@@ -107,7 +125,7 @@ func (a *OpenStackActuator) sync(ctx context.Context, cr *minterv1.CredentialsRe
 		return err
 	}
 
-	err = a.syncCredentialSecret(cr, clouds, existingSecret, "", logger)
+	err = a.syncCredentialSecret(cr, cloudsYAML, cloudsCONF, existingSecret, "", logger)
 	if err != nil {
 		msg := "error creating/updating secret"
 		logger.WithError(err).Error(msg)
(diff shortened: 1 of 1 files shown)


bertinatto:
CC @openshift/storage

bertinatto:
I tested this patch and the operator correctly created a secret with a `clouds.conf` file. Then I used that secret with the Cinder CSI driver and it was able to successfully create/attach a volume.

Thanks for your work on this, @Fedosin!

joelddiaz:
Is there an example of the kind of data that is stored in clouds.conf (I've never come across this in my limited OpenStack experience)?

joelddiaz:
IIUC, clouds.conf is basically just the same contents as clouds.yaml but in INI-style format?

mandre:
> IIUC, clouds.conf is basically just the same contents as clouds.yaml but in INI-style format?

Not exactly - they're actually configuration files for different projects. The [clouds.conf](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#cloud-conf) (should be cloud.conf really, I do not know why we chose to name it clouds.conf here) is the configuration for the cloud provider, while [clouds.yaml](https://docs.openstack.org/python-openstackclient/latest/configuration/index.html) is the configuration for the openstack client, that is also used by gophercloud.

The change looks good to me, we can merge it when it passes e2e-openstack CI.

joelddiaz:
> > IIUC, clouds.conf is basically just the same contents as clouds.yaml but in INI-style format?
> 
> Not exactly - they're actually configuration files for different projects. The [clouds.conf](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#cloud-conf) (should be cloud.conf really, I do not know why we chose to name it clouds.conf here) is the configuration for the cloud provider, while [clouds.yaml](https://docs.openstack.org/python-openstackclient/latest/configuration/index.html) is the configuration for the openstack client, that is also used by gophercloud.

Perhaps I've misread things. My understanding came from how the installer builds this INI-looking content https://github.com/openshift/installer/blob/master/pkg/asset/manifests/openstack/cloudproviderconfig.go#L29

So is it more accurate to say that this clouds.conf file is really useful only for Cinder?

What I'm trying to understand is if in the future we decide to implement "mint" mode for cloud-cred-operator for OpenStack (today we only support "passthrough" mode where we blindly copy the contents of the secret kube-system/openstack-credentials), what do all these fields in the Secret that cloud-cred-operator would need to populate if CCO was making OpenStack API calls to create new credentials while processing a CredentialsRequest object?

Fedosin:
clouds.conf contains the in-tree cloud provider configuration. It includes auth parameters, cinder configuration and other sections: https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#typical-configuration

For Cinder CSI driver they decided to reuse this file as it already has everything they need. Other sections are ignored.

clouds.yaml has just auth parameters in the predefined format: https://docs.openstack.org/python-openstackclient/latest/configuration/index.html#configuration-files

mandre:
FYI e2e-openstack is currently broken due to https://bugzilla.redhat.com/show_bug.cgi?id=1853298. Patch at https://github.com/openshift/machine-config-operator/pull/1773 should solve it.

joelddiaz:
> clouds.conf contains the in-tree cloud provider configuration. It includes auth parameters, cinder configuration and other sections: https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#typical-configuration
> 
> For Cinder CSI driver they decided to reuse this file as it already has everything they need. Other sections are ignored.
> 
> clouds.yaml has just auth parameters in the predefined format: https://docs.openstack.org/python-openstackclient/latest/configuration/index.html#configuration-files

@dgoodwin WDYT? It seems like if we ever need to implement mint mode for OpenStack, it's going to be more involved than "give me creds" -> "save to secret" as there are files with specific formats that need to be built. I think my primary concern here is that we might be building files with storage configuration data/sections when the credentials request is for the ability to create/delete instances.

dgoodwin:
How does it differ from the json blobs we do for GCP?

joelddiaz:
> How does it differ from the json blobs we do for GCP?

Perhaps I simply do not understand how things are normally done in OpenStack, but to me the primary difference is that the GCP auth secret contains the auth info (urls, username, private keys), but not service configuration info (eg. no storage bucket encryption settings).

In my mental model, I would expect the operator that managed running ServiceX to ask for creds, and then combine the creds with the desired ServiceX configuration (assuming ServiceX's config and auth were both bundled into a single file).

With the present situation, for OpenStack mint-mode it feels like CCO would need to know:
1) This CredentialsRequest is for Cinder (or nova or whatever other OpenStack service)
2) Create a user/password that can do the actions requested in the CredRequest
3) Create a Cinder/service-specific secret payload

(and every new service could potentially need CCO-specific handling)

pkg/openstack/actuator.go
```diff
@@ -35,7 +35,8 @@ import (
 )
 
 const (
-	rootOpenStackCredsSecretKey = "clouds.yaml"
+	rootOpenStackCredsSecretKey          = "clouds.yaml"
+	cloudProviderOpenStackCredsSecretKey = "clouds.conf"
```
dgoodwin, line 39: Please add some comments here for the differences between the two keys, as this is definitely going to be confusing.

pkg/openstack/actuator.go
```diff
@@ -95,7 +96,24 @@ func (a *OpenStackActuator) sync(ctx context.Context, cr *minterv1.CredentialsRe
...
+		}
+	}
+
+	cloudsYAML, err := a.getCloudCredentialsSecretData(cloudCredSecret, rootOpenStackCredsSecretKey, logger)
+	if err != nil {
+		logger.WithError(err).Error("issue with cloud credentials secret")
+		return err
+	}
+
+	cloudsCONF, err := a.getCloudCredentialsSecretData(cloudCredSecret, cloudProviderOpenStackCredsSecretKey, logger)
```
dgoodwin, line 116: What's going to happen on a 4.5 to 4.6 upgrade here, presumably this key will not be populated, will all cred requests start failing?

Fedosin:
/hold
we need some time to consider if we really need this key in the CCO output

PR #218, opened 2020-07-01
WIP: rename provider types to be less generic

joelddiaz:


joelddiaz:
@dgoodwin is this valuable? I thought about this as I was moving things over to openshift/api, and the names just seemed generic for a repo that can possibly hold other provider types.

dgoodwin:
Seems ok to me.

joelddiaz:
Renaming the fields would break CredentialsRequest pre-rename objects. Abandoning this idea.

PR #219, opened 2020-07-01
don't register non-core types into scheme

joelddiaz:
Move that logic into the codec creation util.

### DIFF
--- a/pkg/apis/cloudcredential/v1/codec.go
+++ b/pkg/apis/cloudcredential/v1/codec.go
@@ -24,12 +24,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 )
 
-// NewScheme creates a new Scheme
-func NewScheme() (*runtime.Scheme, error) {
-	s := runtime.NewScheme()
-	return s, SchemeBuilder.AddToScheme(s)
-}
-
 // ProviderCodec is a runtime codec for providers.
 // +k8s:deepcopy-gen=false
 type ProviderCodec struct {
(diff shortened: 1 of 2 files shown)


staebler:
What is the reasoning for this? Aren't the provider types part of the API? For example, isn't the `AWSProviderStatus` part of the `cloudcredential.openshift.io/v1` API? Why wouldn't a user of the API expect the provider types to be added to the scheme when the API is added to the scheme?

joelddiaz:
> What is the reasoning for this? Aren't the provider types part of the API? For example, isn't the `AWSProviderStatus` part of the `cloudcredential.openshift.io/v1` API? Why wouldn't a user of the API expect the provider types to be added to the scheme when the API is added to the scheme?

This was just something I noticed while trying to move things to openshift/api. I saw us registering these generically-named objects, and noticed that some where missing (Ovirt and OpenStack). So since the code was working without needing to register these types, why register them (to be clear, I'm genuinely asking this question)?

Perhaps I'm missing the bigger picture, but what does registering types that we cannot directly query out of etcd get us feature-wise (since we have to go through the codec to unpack these provider types)?

joelddiaz:
After some more discussion this approach would be considered non-idiomatic and may end up surprising someone consuming CCO's API objects. Closing.

joelddiaz:
@staebler @dgoodwin let's revisit this PR. It looks like from the comments at https://github.com/openshift/api/pull/683#discussion_r455047315 , that there is some surprise that we're registering some of our embedded types. And it would make for a cleaner migration into openshift/api if we resolved this in CCO before attempting to migrate our types to openshift/api.

joelddiaz:
lets look into removing these from pkg/api (so they aren't part of what is submitted to openshift/api).

PR #224, opened 2020-07-09
WIP: migrate to using openshift/api/cloudcredential API

joelddiaz:
Vendor openshift/api to use the API defnitions located there.

Update import paths as needed.

Remove pkg/apis/cloudcredential API definition from this repo.


joelddiaz:
/hold
would need to revendor once openshift/api#683 merges

PR #226, opened 2020-07-15
[OpenStack] Include cacert in the output

Fedosin:
If OpenStack cloud uses self-signed certificates, the installer stores the cacert file in the configmap that is mounted to kube-apiserver and and kube-controller-manager pods by library-go.

Unfortunately, for all other pods that communicates with OpenStack we have to introduce various mechanisms to obtain this configmap.
It will be more correct to distribute this file using CCO means.

### DIFF
--- a/pkg/openstack/actuator.go
+++ b/pkg/openstack/actuator.go
@@ -36,6 +36,7 @@ import (
 
 const (
 	rootOpenStackCredsSecretKey = "clouds.yaml"
+	rootOpenStackCACertKey      = "ca-bundle.pem"
 )
 
 type OpenStackActuator struct {
@@ -101,13 +102,19 @@ func (a *OpenStackActuator) sync(ctx context.Context, cr *minterv1.CredentialsRe
 		return err
 	}
 
+	cacert, err := a.getRootCACertData(ctx, logger)
+	if err != nil {
+		logger.WithError(err).Error("issue with cacert configmap")
+		return err
+	}
+
 	logger.Debugf("provisioning secret")
 	existingSecret, err := a.loadExistingSecret(cr)
 	if err != nil {
 		return err
 	}
 
-	err = a.syncCredentialSecret(cr, clouds, existingSecret, "", logger)
+	err = a.syncCredentialSecret(cr, clouds, cacert, existingSecret, "", logger)
 	if err != nil {
 		msg := "error creating/updating secret"
 		logger.WithError(err).Error(msg)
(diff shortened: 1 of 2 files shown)


jsafrane:
I appreciate this PR, it helps me to write an operator that gets all data necessary to connect to OpenStack in a single Secret. I don't need to chase the CA certificate around the cluster.

pkg/openstack/actuator.go
```diff
@@ -213,6 +229,11 @@ func (a *OpenStackActuator) GetCredentialsRootSecretLocation() types.NamespacedN
 	return types.NamespacedName{Namespace: constants.CloudCredSecretNamespace, Name: constants.OpenStackCloudCredsSecretName}
 }
 
+// GetCredentialsRootSecretLocation returns the namespace and name where the CACert is stored.
+func (a *OpenStackActuator) GetCACertConfigMapLocation() types.NamespacedName {
+	return types.NamespacedName{Namespace: constants.OpenShiftConfigNamespace, Name: constants.CloudProvderConfigMapName}
+}
```
dgoodwin, line 235: This PR is just copying the CA cert, which we know must exist at this location, into every credentials secret? Is the goal to make it a little easier for someone to use the cred secret and not have to do a second get?

pkg/openstack/actuator.go
```diff
@@ -36,6 +36,7 @@ import (
 
 const (
 	rootOpenStackCredsSecretKey = "clouds.yaml"
+	rootOpenStackCACertKey      = "ca-bundle.pem"
```
joelddiaz, line 39: Is this really an OpenStack specific thing, or does the configmap contain CA cert data for any cloud environment using self-signed certificates?

pkg/openstack/actuator.go
```diff
@@ -236,11 +257,35 @@ func (a *OpenStackActuator) getRootCloudCredentialsSecretData(ctx context.Contex
...
 	clouds = string(keyBytes)
-	logger.Debug("found clouds.yaml in target secret")
+	logger.Debugf("found %v in target secret", rootOpenStackCredsSecretKey)
 
 	return clouds, nil
 }
 
+func (a *OpenStackActuator) getRootCACertData(ctx context.Context, logger log.FieldLogger) (string, error) {
+	caCertConfigMap := &corev1.ConfigMap{}
+	if err := a.Client.Get(ctx, a.GetCACertConfigMapLocation(), caCertConfigMap); err != nil {
```
joelddiaz, line 267: Is this configmap new in 4.6? I have a 4.5 RC and do not see this configmap.
Does this code need to account for the possibility that the configmap does not exist?

jsafrane: The certificate is optional, if the cluster uses self-signed one. https://docs.openshift.com/container-platform/4.5/installing/installing_openstack/installing-openstack-installer-custom.html

pkg/openstack/actuator.go
jsafrane: All public clouds don't use self-signed certificate. I don't have oVirt at hand to check where their certificate ends up.
https://docs.openshift.com/container-platform/4.5/installing/installing_rhv/installing-rhv-customizations.html#installation-launching-installer_installing-rhv-customizations

pkg/openstack/actuator.go
jsafrane: It requires the operator to have additional permissions to get stuff from openshift-config namespace that it does not need to have.

PR #227, opened 2020-07-15
start using the CCO config object

joelddiaz:
1. to determine whether CCO is in manual/disabled mode (old configmap is deprecated and the new config object takes precedence)
2. to bypass permissions checking when annotating the cloud secret (AWS-only presently)
3. update role to grant access to read the config
4. copy generated config CRD from openshift/api into manifests so that it gets registered
5. update clusteroperator status reporting for bad modes (and conflicting legacy settings and new modes)

Also consolidate scheme registration.
Fix conditions merging to not drop old conditions when new StatusHandler has no conditions to set.
Fix fallout from that fix where StatusHandlers registered during tests were stacking up and interfering with each other.


joelddiaz:
/hold
this PR depends on https://github.com/openshift/api/pull/692

joelddiaz:
@staebler PTAL, this set of changes is the minimal set (IMO) to get CCO to read/respect the new operator config

The other PR will handle making the operator config required for runtime operation and handling the bootstrap mode behavior

pkg/operator/secretannotator/aws/reconciler.go
```diff
@@ -178,8 +220,55 @@ func (r *ReconcileCloudCredSecret) updateSecretAnnotations(secret *corev1.Secret
...
+	secretAnnotations[constants.AnnotationKey] = value
 	secret.SetAnnotations(secretAnnotations)
 
 	return r.Update(context.Background(), secret)
 }
+
+// earlyAnnotationIfModeForced will read the CCO config, and go ahead and annotate the secret if
+// there is a forcedCredentialsMode set. It will return the fetched mode (empty string if no mode set),
+// whether the operator is in manual/disabled mode, and any errors encountered.
+func (r *ReconcileCloudCredSecret) earlyAnnotationIfModeForced(secretKey types.NamespacedName) (modeString string, disabled bool, err error) {
```
staebler, line 232: Is there a reason why there needs to be both a `modeString` and `disabled` return value? Can we consolidate into just a `modeString` which is set to "manual" when the operator is disabled?

pkg/operator/secretannotator/aws/reconciler.go
```diff
@@ -178,8 +220,55 @@ func (r *ReconcileCloudCredSecret) updateSecretAnnotations(secret *corev1.Secret
...
+		err = r.updateSecretAnnotations(secret, constants.MintAnnotation)
+		modeString = constants.MintAnnotation
+	case operatorv1.CloudCredentialsModePassthrough:
+		err = r.updateSecretAnnotations(secret, constants.PassthroughAnnotation)
+		modeString = constants.PassthroughAnnotation
+	case "":
+		// Mode not set, no worries; will have to do dynamic permissions checking
+	default:
+		r.Logger.Warnf("Unexpected mode: %s", mode)
+		err = fmt.Errorf("unexpected mode retrieved: %v", mode)
```
staebler, line 266: Should an unexpected mode put the operator into a degraded state?

pkg/operator/secretannotator/aws/reconciler.go
```diff
@@ -178,8 +220,55 @@ func (r *ReconcileCloudCredSecret) updateSecretAnnotations(secret *corev1.Secret
...
+	}
+
+	if err != nil {
+		r.Logger.WithError(err).Error("failed to update secret annotations")
```
staebler, line 270: This will log this error entry when there is an unexpected mode as well. It does not appear that that is the intention.

pkg/operator/secretannotator/aws/reconciler.go
```diff
@@ -178,8 +220,55 @@ func (r *ReconcileCloudCredSecret) updateSecretAnnotations(secret *corev1.Secret
...
+// there is a forcedCredentialsMode set. It will return the fetched mode (empty string if no mode set),
+// whether the operator is in manual/disabled mode, and any errors encountered.
+func (r *ReconcileCloudCredSecret) earlyAnnotationIfModeForced(secretKey types.NamespacedName) (modeString string, disabled bool, err error) {
+	var mode operatorv1.CloudCredentialsMode
+	mode, err = utils.GetOperatorMode(r.Client, r.Logger)
+	if err != nil {
+		return
+	}
+
+	disabled, err = utils.IsOperatorDisabled(r.Client, r.Logger)
```
staebler, line 239: Should it be an error if the mode was explicitly set to something other than "manual" and there is a configmap setting the operator to disabled?

pkg/util/scheme.go
```diff
@@ -0,0 +1,30 @@
...
+	if err := apis.AddToScheme(scheme); err != nil {
+		log.WithError(err).Fatal("unable to add APIs to scheme")
+	}
+
+	// Setup OpenShift operator configs
+	if err := operatorv1.Install(scheme); err != nil {
+		log.Fatal(err)
+	}
+
+	// Setup Openshift API scheme:
```
staebler, line 26:
```suggestion
	// Setup Openshift config scheme:
```

pkg/operator/secretannotator/aws/reconciler.go
joelddiaz: I've been treating the presence of an entry in the config CR as taking precedence over any legacy ConfigMap value.
WDYT about printing a warning if that is detected (explaining that the value in the config CR is being used)?

staebler: My thinking here is that the ConfigMap is something that the user had to deliberately create and the Mode is something that the user had to deliberately set. I don't think that one has precedence over the other. If they are both present but request different things, then that is a contradiction that CCO cannot resolve.

joelddiaz: My concern with that is b/c with the next PR where we start to require the presence of a config CR, on upgrade we will lay down an empty config CR and it would immediately put CCO in a bad state as a result of upgrade.

staebler: No, an empty config CR and a disabled configmap is fine. The only thing that should be an error is a config CR with the mode set to something other than "manual" and a disabled configmap.

staebler: I see where that last statement is a bit confusing. By "set to something other than manual", I mean a config CR that has a mode set (ie, not empty) and the mode is something other than "manual".

staebler: Config CR Mode | Disabled ConfigMap | Behavior
---- | ---- | ----
"" | no | permissions checks
"" | yes | manual
"manual" | no | manual
"manual" | yes | manual
"passthrough" | no | passthrough
"passthrough" | yes | error

pkg/azure/actuator.go
```diff
@@ -237,9 +237,7 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
 			ErrReason: minterv1.InsufficientCloudCredentials,
 			Message:   msg,
 		}
-	}
-
-	if cloudCredsSecret.Annotations[constants.AnnotationKey] == constants.PassthroughAnnotation {
+	} else if cloudCredsSecret.Annotations[constants.AnnotationKey] == constants.PassthroughAnnotation {
```
staebler, line 240: I find it easier to understand at a glance if a switch is used.
```
switch cloudSecret.Annotations[constants.AnnotationKey] {
case constants.InsufficientAnnotation:
   ...
case constants.PassthroughAnnotation:
   ...
case constants.MintAnnotation:
   ...
default:
   ...
}

File: pkg/operator/constants/constants.go [hunk 7]
```diff
@@ -32,6 +32,15 @@ const (
 	// running under (typically just haven't added support for the cloud/platform)
 	ModeUnknown CredentialsMode = "unknown"
 
+	// StatusModeMissmatch is used to set a clusteroperator condition when
+	// the legacy configmap setting of disabled: "true" conflicts with the
+	// specified operator config mode.
+	StatusModeMissmatch = "ModeMissmatch"
```
staebler, line 38:
```suggestion
	StatusModeMismatch = "ModeMismatch"
```

pkg/operator/utils/utils.go
```diff
@@ -147,30 +149,102 @@ func GenerateNameWithFieldLimits(infraName string, infraNameMaxLen int, crName s
...
+	return cm, err
+}
+
+// GetOperatorConfiguration will return the value in the operator config, and whether the operator has been disabled
+// via either the "manual" mode setting, or the legacy configmap, and whether there is a conflict between the legacy
+// and CCO config (in the even of a conflict, the disabledSetting field will reflect the value in the operator config
+// in the event of a conflict).
+func GetOperatorConfiguration(kubeClient client.Client, logger log.FieldLogger) (
+	operatorMode operatorv1.CloudCredentialsMode,
+	disabledSetting bool,
```
staebler, line 186: I still don't think there is much benefit is returning a separate bool for disabled. The caller can easily check `operatorMode == ManualMode`.

pkg/operator/secretannotator/aws/reconciler.go
```diff
@@ -78,37 +115,61 @@ type ReconcileCloudCredSecret struct {
...
+// In the event that the operator config resource has specified a mode to operate under (mint/passthrough)
+// then skip trying to determine the capabilities, and just annotate the secret.
 // +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;update
 func (r *ReconcileCloudCredSecret) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	start := time.Now()
 
-	r.Logger.Info("validating cloud cred secret")
-
-	operatorIsDisabled, err := utils.IsOperatorDisabled(r.Client, r.Logger)
+	mode, disabled, conflict, err := utils.GetOperatorConfiguration(r.Client, r.Logger)
```
staebler, line 128: I think this should be done after the defer to record metrics. We would want to record the metrics for the reconcile even if the mode is manual, there is a conflict, or there is an error determining the mode.

pkg/operator/secretannotator/aws/reconciler.go
```diff
@@ -78,37 +115,61 @@ type ReconcileCloudCredSecret struct {
...
+	} else if conflict {
+		r.Logger.Error("configuration conflict between legacy configmap and operator config")
+		return reconcile.Result{}, fmt.Errorf("configuration conflict")
+	} else if disabled {
+		r.Logger.Info("operator in disabled / manual mode")
 		return reconcile.Result{}, err
 	}
 
 	defer func() {
+		if err := r.syncOperatorStatus(); err != nil {
```
staebler, line 141: Separate this into its own defer since the metrics recording is a distinct action not related to syncing the operator status.

pkg/util/clusteroperator/status.go
```diff
@@ -36,6 +37,11 @@ var (
 	statusHandlers = []StatusHandler{}
 )
 
+// ClearHandlers so that test cases don't endlessly add handlers
```
staebler, line 40: This may lead to some flaky tests since tests should be able to be run in parallel.

pkg/operator/secretannotator/azure/status.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/openshift/cloud-credential-operator/pkg/util/clusteroperator"
+)
+
+// syncOperatorStatus computes the operator's current status and
+// creates or updates the ClusterOperator resource for the operator accordingly.
+func (r *ReconcileCloudCredSecret) syncOperatorStatus() error {
+	return status.SyncOperatorStatus(r.Client)
+}
+
+var _ clusteroperator.StatusHandler = &ReconcileCloudCredSecret{}
```
staebler, line 18: Is there a reason why we need the reconciler itself to be a `StatusHandler`? Could we define a type in pkg/operator/secretannotator/status/status.go that was the `StatusHandler` instead? That way we do not need this boilerplate code for each platform.

pkg/operator/secretannotator/status/status.go
```diff
@@ -0,0 +1,87 @@
...
+func GetConditions(kubeClient client.Client, logger log.FieldLogger) ([]configv1.ClusterOperatorStatusCondition, error) {
+	conditions := []configv1.ClusterOperatorStatusCondition{}
+
+	mode, _, conflict, err := utils.GetOperatorConfiguration(kubeClient, logger)
+	if err != nil {
+		return conditions, err
+	}
+
+	// shouldn't happen with the server-side enforcement of the CRDs enum specification
+	if valid := utils.IsValidMode(mode); !valid {
```
staebler, line 35:
```suggestion
	if !utils.IsValidMode(mode) {
```

pkg/operator/secretannotator/status/status.go
```diff
@@ -0,0 +1,87 @@
...
+		conditions = append(conditions, configv1.ClusterOperatorStatusCondition{
+			Type:    configv1.OperatorDegraded,
+			Status:  configv1.ConditionTrue,
+			Reason:  constants.StatusModeInvalid,
+			Message: fmt.Sprintf("operator mode of %s is invalid", mode),
+		})
+	}
+
+	if conflict {
+		conditions = append(conditions, configv1.ClusterOperatorStatusCondition{
```
staebler, line 45: If there is an invalid mode **and** a conflict, then the Degraded condition will be added twice.

pkg/operator/secretannotator/aws/reconciler.go
```diff
@@ -67,6 +87,23 @@ func Add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
+
+	// Watch for changes to CCO config
+	configPredicate := predicate.Funcs{
```
staebler, line 92: This watch should be added to the secretannotator for the other platforms as well. Extract this into a function that can be called by all of the secretannotators.

pkg/operator/constants/constants.go
joelddiaz: the irony :)

pkg/operator/utils/utils.go
joelddiaz: If operatorMode == "" the legacy configmap might still have an opinion about disabled/manual mode.

I think I ended up organizing everything behind this GetOperatorConfiguration() b/c I was originally returning an error when there was a conflict between the configmap and the operator config, and the status handlers needed a way to set conditions based on the various states that we can find ourselves in.

staebler: My contention is that the `operatorMode` returned should be "manual" if the config CR has not mode set and the disabled ConfigMap is set to disabled.

staebler: It does not matter to callers of `GetOperatorConfiguration` whether the mode was set via the config CR or via the disabled CofigMap.

joelddiaz:
@staebler addressed everything except the registering/clearing of the status handlers. Need to think about that one.

pkg/operator/secretannotator/aws/reconciler.go
joelddiaz: Yes. Implemented.

pkg/cmd/operator/cmd.go
```diff
@@ -71,7 +71,7 @@ func NewOperator() *cobra.Command {
 			log.Info("setting up manager")
 			mgr, err := manager.New(cfg, manager.Options{
 				MetricsBindAddress:      ":2112",
-				LeaderElection:          true,
+				LeaderElection:          false,
```
staebler, line 74: Was this an intentional change? Or something left in from testing?

pkg/operator/secretannotator/aws/reconciler.go
```diff
@@ -125,6 +108,17 @@ type ReconcileCloudCredSecret struct {
 func (r *ReconcileCloudCredSecret) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	start := time.Now()
 
+	defer func() {
+		dur := time.Since(start)
+		metrics.MetricControllerReconcileTime.WithLabelValues(constants.SecretAnnotatorControllerName).Observe(dur.Seconds())
+	}()
+
+	defer func() {
+		if err := r.Status.SyncOperatorStatus(); err != nil {
```
staebler, line 117: Should an error here cause the reconcile to return an error?

pkg/operator/secretannotator/status/status.go
```diff
@@ -16,36 +16,46 @@ import (
...
+		Client: kubeClient,
+	}
+}
+
+var _ clusteroperator.StatusHandler = &SecretStatusHandler{}
+
 // SyncOperatorStatus computes the operator's current status and
 // creates or updates the ClusterOperator resource for the operator accordingly.
-func SyncOperatorStatus(kubeClient client.Client) error {
+func (s *SecretStatusHandler) SyncOperatorStatus() error {
```
staebler, line 33: So that the reconcilers do not have to carry around the `StatusHandler` object.
```suggestion
func SyncOperatorStatus(c client.Client) error {
```

pkg/operator/secretannotator/status/status.go
```diff
@@ -16,36 +16,46 @@ import (
 	"github.com/openshift/cloud-credential-operator/pkg/util/clusteroperator"
 )
 
+type SecretStatusHandler struct {
+	client.Client
```
staebler, line 20: Name this field. I don't see any reason why `SecretStatusHandler` should be inheriting from `client.Client` instead of just composing it.

pkg/operator/secretannotator/vsphere/reconciler.go
```diff
@@ -129,6 +144,17 @@ func (r *ReconcileCloudCredSecret) Reconcile(request reconcile.Request) (reconci
 }
 
 func (r *ReconcileCloudCredSecret) validateCloudCredsSecret(secret *corev1.Secret) error {
+	// We only support passthrough so make sure the operator mode is either unset or "passthrough"
```
staebler, line 147: Thought it worth mentioning that the installer will not allow the mode to be set to "passthrough". Of course, the user can set the mode directly in the config CR afterwards.

pkg/operator/secretannotator/vsphere/reconciler.go
```diff
@@ -129,6 +144,17 @@ func (r *ReconcileCloudCredSecret) Reconcile(request reconcile.Request) (reconci
...
+	mode, _, _, err := utils.GetOperatorConfiguration(r.Client, r.Logger)
+	if err != nil {
+		return err
+	}
+	if mode != "" && mode != operatorv1.CloudCredentialsModePassthrough {
```
staebler, line 152: Not something that we should tackle for this PR, but this is also something that should cause the operator to be degraded.

pkg/operator/utils/utils.go
joelddiaz: Fair enough. I dropped the disabled field, and will just insert the "manual" mode if the conditions are right.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -372,14 +373,14 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
...
-	_, operatorIsDisabled, conflict, err := utils.GetOperatorConfiguration(r.Client, logger)
+	mode, conflict, err := utils.GetOperatorConfiguration(r.Client, logger)
 	if err != nil {
 		logger.WithError(err).Error("error checking if operator is disabled")
 		return reconcile.Result{}, err
 	} else if conflict {
 		logger.Error("configuration conflict betwen legacy configmap and operator config")
 		return reconcile.Result{}, fmt.Errorf("configuration conflict")
-	} else if operatorIsDisabled {
+	} else if mode == operatorv1.CloudCredentialsModeManual {
```
staebler, line 383: Don't need to change it, but I find the `else` distracting. There is no need for it to be an `else` since each of the previous blocks return from the function.
```
mode, conflict, err := ...
if err != nil {
   ...
}
if conflict {
   ...
}
if mode == operatatorv1.CloudCredentialsModeManual {
   ...
}
```

pkg/operator/metrics/metrics.go
```diff
@@ -118,11 +119,12 @@ func (mc *Calculator) metricsLoop() {
 
 	mc.log.Info("calculating metrics for all CredentialsRequests")
 
-	_, ccoDisabled, _, err := utils.GetOperatorConfiguration(mc.Client, mc.log)
+	mode, _, err := utils.GetOperatorConfiguration(mc.Client, mc.log)
 	if err != nil {
 		mc.log.WithError(err).Error("failed to determine whether CCO is disabled")
 		return
 	}
+	ccoDisabled := mode == operatorv1.CloudCredentialsModeManual
```
staebler, line 127: Also not something to change now, but we should make an effort in the future to remove the "disabled" concept from the CCO and replace it with "manual mode".

pkg/cmd/operator/cmd.go
joelddiaz: Nope, just testing stuff that snuck in. Good catch!

pkg/operator/secretannotator/aws/reconciler.go
joelddiaz: now returning a "manual" mode when only the legacy configmap is set.

pkg/operator/secretannotator/aws/reconciler.go
joelddiaz: done

pkg/operator/secretannotator/aws/reconciler.go
joelddiaz: done

pkg/operator/secretannotator/azure/status.go
joelddiaz: excellent suggestion. done.

pkg/operator/secretannotator/aws/reconciler.go
joelddiaz: done

joelddiaz:
/unhold
openshift/api#692 merged

PR #228, opened 2020-07-17
handle bootstrap user-defined mode

joelddiaz:
- [x] update Makefile to copy operator config from vendored openshift/api
- [x] copy config CRD into bindata so that render subcommand has access to the file
- [x] update render subcommand to check the ConfigMap containing the install-config to determine static Pod rendering
- [x] render will now also create the operator config CRD and a operator config CR (reflecting legacy configmap and/or operator config)
- [x] update disabled detection util to error if there is no operator config CR present
- [x] remove default operator configmap as it is now deprecated
- [x] add default operator config CR with default behavior (where forceCredentialsMode is "")
- [x] update test cases so that an operator config resource does exist
- [x] add test cases to cover the various render subcommand behaviors

New operator behavior will now refuse to take actions until the operator config CR (named "cluster") is present.


joelddiaz:
/hold
this depends on openshift/cloud-credential-operator#227

Makefile
```diff
@@ -76,7 +76,13 @@ $(call build-image,ocp-cloud-credential-operator,$(IMAGE_REGISTRY)/ocp/4.5:cloud
 $(call add-crd-gen,cloudcredential-manifests,./pkg/apis/cloudcredential/v1,./manifests,./manifests)
 $(call add-crd-gen,cloudcredential-bindata,./pkg/apis/cloudcredential/v1,./bindata/bootstrap,./bindata/bootstrap)
 
-update: update-codegen
+update: update-vendored-crds update-codegen update-bindata
+
+update-vendored-crds:
+	# copy config CRD from openshift/api
+	cp vendor/github.com/openshift/api/operator/v1/0000_40_cloud-credential-operator_00_config.crd.yaml ./manifests/00-config-crd.yaml
```
staebler, line 83: Would it work to make this a symbolic link instead so that we do not need the new make target?

manifests/01-operator-config.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: operator.openshift.io/v1
+kind: CloudCredential
+metadata:
+  name: cluster
+  annotations:
+    release.openshift.io/create-only: "true"
+spec:
+  forceCredentialsMode: ""
```
staebler, line 8: Delete this. It is not necessary.

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,218 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+	// Make sure deprecated configmap setting doesn't conflict with install-config contents.
+	// Otherwise there could be a race between the configmap and the operator config
+	// resources landing causing unpredictable runtime behavior of the CCO pod.
+	if operatorDisabledViaConfigmap && installConfigMode != "" {
+		if operatorDisabledViaConfigmap != installConfigModeManual {
+			return fmt.Errorf("config map asking for CCO to be disabled, and install-config asking for %s mode", installConfigMode)
+		}
+	}
+
+	operatorDisabled := operatorDisabledViaConfigmap || installConfigMode == "manual"
```
staebler, line 165: It would be good to extract this into a function that can be used by `GetOperatorConfiguration` as well.

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,218 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+		return fmt.Errorf("invalid mode defined: %s", installConfigMode)
+	}
+
+	installConfigModeManual := installConfigMode == string(operatorv1.CloudCredentialsModeManual)
+
+	// Make sure deprecated configmap setting doesn't conflict with install-config contents.
+	// Otherwise there could be a race between the configmap and the operator config
+	// resources landing causing unpredictable runtime behavior of the CCO pod.
+	if operatorDisabledViaConfigmap && installConfigMode != "" {
+		if operatorDisabledViaConfigmap != installConfigModeManual {
```
staebler, line 160:
```suggestion
		if !installConfigModeManual {
```
Or don't even define `installConfigModeManual`
```suggestion
		if installConfigMode == string(operatorv1.CloudCredentialsModeManual) {
```

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,218 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+	}
+
+	return disabled
+}
+
+type basicInstallConfig struct {
+	CredentialsMode string `yaml:"credentialsMode"`
+}
+
+func getModeFromInstallConfig() (string, error) {
```
staebler, line 285: Change this to return an `operatorv1.CloudCredentialsMode` so that you don't have to have casts littered throughout the code elsewhere.

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,218 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+	return nil
+}
+
+func isValidMode(mode string) bool {
+	switch mode {
+	case string(operatorv1.CloudCredentialsModeManual):
+	case string(operatorv1.CloudCredentialsModeMint):
+	case string(operatorv1.CloudCredentialsModePassthrough):
+	case "":
+		// all the above are acceptable
```
staebler, line 243:
```suggestion
	case string(operatorv1.CloudCredentialsModeManual),
		string(operatorv1.CloudCredentialsModeMint),
		string(operatorv1.CloudCredentialsModePassthrough),
		"":
		return true
```

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,218 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+	var operatorConfig bytes.Buffer
+	templateVars := operatorTemplateVars{
+		ForceCredentialsMode: installConfigMode,
+	}
+
+	// if either the legacy config map or the install-config have CCO as disabled/manual,
+	// then we must set the config appropriately.
+	if operatorDisabled {
+		templateVars.ForceCredentialsMode = string(operatorv1.CloudCredentialsModeManual)
+	}
```
staebler, line 198: Define a `mode` variable earlier in the function closer to where the install-config mode and disabled setting are determined. Use the `mode` variable here rather than having logic here looking at `operatorDisabled`.

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,218 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+	if err := writeFile(assetRenderPath, operatorConfig.Bytes()); err != nil {
+		return err
+	}
+
 	// need at least the empty dir so the installer bootkube.sh script works as expected
 	if err := os.Mkdir(filepath.Join(ccoRenderDir, bootstrapManifestsDir), 0775); err != nil {
-		log.WithError(err).Fatal("error creating bootstrap-manifests directory")
+		errors.Wrap(err, "error creating bootstrap-manifests directory")
 	}
 	if !operatorDisabled {
```
staebler, line 212: Outside of the scope of this work, but do we care if the static pod is run when disabled? Since we will now be waiting for the config CR to exist, there is no longer the possibility of the static pod winning a race and incorrectly doing work.

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,218 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+		log.WithError(err).Warnf("errored while searching for ConfigMap %s/%s, using defaults", minterv1.CloudCredOperatorNamespace, constants.CloudCredOperatorConfigMap)
 		return utils.OperatorDisabledDefault
 	}
+	if cm == nil {
+		log.Debugf("configmap %s/%s not found, using default mode", minterv1.CloudCredOperatorNamespace, constants.CloudCredOperatorConfigMap)
+		return utils.OperatorDisabledDefault
+	}
+	logger := log.New()
+	logger.SetLevel(log.GetLevel())
+	disabled, err := utils.CCODisabledCheck(cm, logger)
```
staebler, line 273:
```suggestion
	disabled, err := utils.CCODisabledCheck(cm, log.StandardLogger())
```
or
```suggestion
	disabled, err := utils.CCODisabledCheck(cm, log.WithFields(nil))
```

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,218 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		return utils.OperatorDisabledDefault
+	}
+
+	return disabled
+}
+
+type basicInstallConfig struct {
+	CredentialsMode string `yaml:"credentialsMode"`
```
staebler, line 282:
```suggestion
	CredentialsMode operatorv1.CloudCredentialsMode `json:"credentialsMode"`
```

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,218 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+
+	files, err := ioutil.ReadDir(manifestsDir)
+	if err != nil {
+		log.WithError(err).Errorf("failed to list files in %s", manifestsDir)
+		return nil, err
+	}
 
 	for _, fInfo := range files {
-		// non-recursive checking of all files where the the source manifests are located
+		// non-recursive checking of all files whewre the source manifests are located
```
staebler, line 324:
```suggestion
		// non-recursive checking of all files where the source manifests are located
```

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,218 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
-		if configMap.Namespace == minterv1.CloudCredOperatorNamespace && configMap.Name == constants.CloudCredOperatorConfigMap {
-			logger := log.New()
-			logger.SetLevel(log.GetLevel())
-			disabled, err := utils.CCODisabledCheck(configMap, logger)
-			if err != nil {
-				return utils.OperatorDisabledDefault
-			}
-
-			return disabled
+		if configMap.Namespace == namespace && configMap.Name == name {
```
staebler, line 344: This will end up matching things that are not ConfigMaps. There needs to be a check of the APIVersion and Kind as well.

pkg/cmd/render/render.go
joelddiaz: Definitely. I don't think GetOperatorConfiguration() had all the current capabilities when I first took a pass at this.

pkg/cmd/render/render.go
```diff
@@ -151,19 +151,12 @@ func render() error {
...
 
-	// Make sure deprecated configmap setting doesn't conflict with install-config contents.
-	// Otherwise there could be a race between the configmap and the operator config
-	// resources landing causing unpredictable runtime behavior of the CCO pod.
-	if operatorDisabledViaConfigmap && installConfigMode != "" {
-		if operatorDisabledViaConfigmap != installConfigModeManual {
-			return fmt.Errorf("config map asking for CCO to be disabled, and install-config asking for %s mode", installConfigMode)
-		}
+	if conflict {
+		return fmt.Errorf("config map asking for CCO to be disabled, and install-config asking for %s mode", installConfigMode)
```
staebler, line 157:
```suggestion
		return fmt.Errorf("config map asking for CCO to be disabled, and install-config asking for %q mode", installConfigMode)
```

pkg/cmd/render/render.go
```diff
@@ -188,13 +181,7 @@ func render() error {
...
 	var operatorConfig bytes.Buffer
 	templateVars := operatorTemplateVars{
-		ForceCredentialsMode: installConfigMode,
-	}
-
-	// if either the legacy config map or the install-config have CCO as disabled/manual,
-	// then we must set the config appropriately.
-	if operatorDisabled {
-		templateVars.ForceCredentialsMode = string(operatorv1.CloudCredentialsModeManual)
+		CredentialsMode: string(effectiveMode),
```
staebler, line 184: Does the template work when `CredentialsMode` has the type `operatorv1.CloudCredentialsMode` instead of `string`?

pkg/cmd/render/render.go
```diff
@@ -279,10 +264,10 @@ func isDisabledViaConfigmap() bool {
 }
 
 type basicInstallConfig struct {
-	CredentialsMode string `yaml:"credentialsMode"`
+	CredentialsMode operatorv1.CloudCredentialsMode `json:"credentialsMode,omitempty"`
```
staebler, line 267: I don't think that you need "omitempty" here. That is only useful for marshaling not for unmarshaling.

pkg/operator/utils/utils.go
```diff
@@ -180,41 +180,50 @@ func GetLegacyConfigMap(kubeClient client.Client) (*corev1.ConfigMap, error) {
...
 	// else see if there is a disconnect between the operator mode and the
 	// opt-in value of 'disabled: "true"' in the ConfigMap
-	if disabledViaConfigMap && disabledViaConfigMap != disabledViaOperatorConfig {
-		log.Errorf("legacy configmap disabled set to %v conflict with operator CR mode of %s",
-			disabledViaConfigMap, operatorMode)
-		// Allow the actual CCO config to be returned as that should take
-		// precedence for the purpose of reporting metrics data, but indicate the conflict.
-		configurationConflict = true
+	disabledViaOperatorConfig := operatorConfigMode == operatorv1.CloudCredentialsModeManual
+	if configMapDisabledValue && configMapDisabledValue != disabledViaOperatorConfig {
```
staebler, line 222:
```suggestion
	if configMapDisabledValue && !disabledViaOperatorConfig {
```

pkg/cmd/render/render.go
joelddiaz: This is a good point. I'll write up a card to explore retiring this conditional launch logic.

pkg/cmd/render/render.go
joelddiaz: It certainly does.

pkg/cmd/render/render.go
```diff
@@ -109,98 +134,206 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
...
+	}
+	assetRenderPath := filepath.Join(ccoRenderDir, manifestsDir, operatorConfigFilename)
+	if err := writeFile(assetRenderPath, operatorConfig.Bytes()); err != nil {
+		return err
+	}
+
 	// need at least the empty dir so the installer bootkube.sh script works as expected
 	if err := os.Mkdir(filepath.Join(ccoRenderDir, bootstrapManifestsDir), 0775); err != nil {
-		log.WithError(err).Fatal("error creating bootstrap-manifests directory")
+		errors.Wrap(err, "error creating bootstrap-manifests directory")
```
staebler, line 197: This is not doing anything. I assume that it should be returning the wrapped error.

joelddiaz:
@staebler i've squashed and it should be good assuming it survives CI

pkg/cmd/render/render.go
joelddiaz: Most definitely. Fixed.
I also slipped in some Makefile hacks to work around some issues I ran into regarding when the 'bindata' generation is happening.

joelddiaz:
/unhold
openshift/cloud-credential-operator#227 merged

PR #231, opened 2020-07-31
Bug 1858403: Use client-go leader election to write less.  

dgoodwin:
We were using the defaults from controller runtime previously:

lease duration: 15s
renew deadline: 10s
retry period: 2s

This meant that the active leader was writing to etcd every 2 seconds to
update the lease, which is excessive writing and spawned the bug above.

We now implement leader election using the underlying client-go code to                                                                                                                                                                                   
get access to ReleaseOnCancel, which is not presently exposed in                                                                                                                                                                                          
controller-runtime.                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                           
This allows us to immediately release the lock on normal shutdown                                                                                                                                                                                         
eliminating delay before another pod takes over, as well as startup                                                                                                                                                                                       
delay when doing development etc.

### DIFF
--- a/pkg/cmd/operator/cmd.go
+++ b/pkg/cmd/operator/cmd.go
@@ -17,11 +17,16 @@ limitations under the License.
 package cmd
 
 import (
+	"context"
 	"flag"
 	golog "log"
+	"os"
+	"os/signal"
+	"syscall"
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -30,8 +35,12 @@ import (
 	controller "github.com/openshift/cloud-credential-operator/pkg/operator"
 	"github.com/openshift/cloud-credential-operator/pkg/util"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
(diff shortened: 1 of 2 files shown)


smarterclayton:
/hold

Please see my comments on the BZ, this is not the right way to solve this problem (you should release the lock on shutdown)

dgoodwin:
PR updated, now uses client-go leader election directly so we can get access to ReleaseOnCancel which is not currently exposed in controller-runtime.

Now generating uuids for leader election ids, previously it was using a flat string which seems like it should have been busted, but somehow they were getting uuids appended in the actual configmap.

Tested bringing up multiple processes and switching between. Provided we exit cleanly there is just a couple seconds startup delay when the new process takes over, or just when restarting a single process.

pkg/cmd/operator/cmd.go
```diff
@@ -67,35 +76,95 @@ func NewOperator() *cobra.Command {
...
-			// Setup Scheme for all resources
-			util.SetupScheme(mgr.GetScheme())
-
-			// Setup all Controllers
-			log.Info("setting up controller")
-			kubeconfigCommandLinePath := cmd.PersistentFlags().Lookup("kubeconfig").Value.String()
-			if err := controller.AddToManager(mgr, kubeconfigCommandLinePath); err != nil {
-				log.WithError(err).Fatal("unable to register controllers to the manager")
+			// Leader election code based on:
+			// https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/examples/leader-election/main.go#L130
```
staebler, line 109: Make this a permalink, and cover the entire section that you are basing the code on.
https://github.com/kubernetes/kubernetes/blob/f7e3bcdec2e090b7361a61e21c20b3dbbb41b7f0/staging/src/k8s.io/client-go/examples/leader-election/main.go#L92-L154

pkg/cmd/operator/cmd.go
```diff
@@ -67,35 +76,95 @@ func NewOperator() *cobra.Command {
...
+				<-ch
+				log.Info("received termination, signaling shutdown")
+				cancel()
+			}()
+
+			id := uuid.New().String()
+			leLog := log.WithField("id", id)
+			leLog.Info("generated leader election ID")
+
+			lock := &resourcelock.ConfigMapLock{
```
staebler, line 131: Is this a good time to switch to using a `LeaseLock` instead of a `ConfigMapLock`? Or is this more complicated due to needing to upgrade existing CCOs that would still be using a `ConfigMapLock`?

dgoodwin: I wasn't totally aware how leaselock works, doesn't appear to be a resource in-cluster at least in 4.5 where I was running this. There is potential for some problems during upgrade though for sure, I'd vote to stay consistent for now.

pkg/cmd/operator/cmd.go
dgoodwin: Fixed.

dgoodwin:
Whoops that was supposed to be a 

/test e2e-aws

smarterclayton:
/hold cancel

since you step down

sdodson:
/hold
@dgoodwin ptql seems like a persistent failurr

joelddiaz:
/test e2e-aws
Looked through the artifacts from the most recent failure, and didn't come across anything pointing to CCO. Built a custom image with this PR's changes and the CI images from https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/releasestream/4.6.0-0.ci/release/4.6.0-0.ci-2020-08-18-103144. I saw no issues.

dgoodwin:
Failing for a week? This is really strange, there does not appear to be anything in here which could impact beyond the CCO, which we can see is running fine in the artifacts.

/test e2e-aws

dgoodwin:
Ok @joelddiaz had the idea that perhaps this PR is cursed and suggested we reopen it, indeed that passed e2e-aws on first try. I am replacing this PR with https://github.com/openshift/cloud-credential-operator/pull/239, I have no idea why this is seeing the failures it is but we can't see any reason why a change to our leader election would be causing mass API failures across the cluster.

PR #234, opened 2020-08-13
Bug 1871954: more auto deepcopy functions (less manual)

joelddiaz:
and fix up bug around incomplete DeepCopy (where changing the source object would change parts of the copied object).
add test cases
remove assert library for these tests (not appropriate for openshift/api)


pkg/apis/cloudcredential/v1/aws_manual.deepcopy.go
```diff
@@ -1,79 +1,50 @@
...
+	if *in == nil {
+		return
 	}
-	return nil
-}
 
-func deepCopyIAMPolicyCondition(ipc IAMPolicyCondition) IAMPolicyCondition {
-	cp := make(IAMPolicyCondition)
-	for key, val := range ipc {
+	*out = make(IAMPolicyCondition)
```
staebler, line 19:
```suggestion
	*out = make(IAMPolicyCondition, len(*in))
```

pkg/apis/cloudcredential/v1/aws_manual.deepcopy.go
```diff
@@ -1,79 +1,50 @@
...
+	tgt := *out
+
+	for key, val := range *in {
 		if val != nil {
-			cp[key] = make(IAMPolicyConditionKeyValue)
+			tgt[key] = make(IAMPolicyConditionKeyValue)
```
staebler, line 24:
```suggestion
			tgt[key] = make(IAMPolicyConditionKeyValue, len(val))
```

pkg/apis/cloudcredential/v1/aws_manual.deepcopy.go
```diff
@@ -1,79 +1,50 @@
...
+	case string:
+		// simple assignment creates copy
+		to = from
+	case []string:
+		toSlice := make([]string, len(v))
+		copy(toSlice, v)
+		to = toSlice
+	default:
+		// unexpected type, this could mean we're not
+		// doing a deepcopy
```
staebler, line 45: Is the expectation that the value in an IAMPolicyConditionKeyValue will always be a `string` or a `[]string`?

joelddiaz: Yes. Reading through the AWS documentation those were the only values I could see documented, and I also tried making policies with the AWS Web UI tooling, and I was also unable to make anything besides `string` and `[]string`.

PR #240, opened 2020-08-27
Bug 1873345: specify a serviceaccount in the deployment

joelddiaz:
allow downgrading from 4.6 (where we do specify a serviceaccount)

### DIFF
--- a/manifests/05_deployment.yaml
+++ b/manifests/05_deployment.yaml
@@ -238,6 +238,8 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
       priorityClassName: system-cluster-critical
+      serviceAccount: default
+      serviceAccountName: default
       terminationGracePeriodSeconds: 10
       tolerations:
       - effect: NoSchedule


joelddiaz:
@sdodson here's the proposed fix

sdodson:
So this was something that defaulted in 4.5, they upgrade to 4.6 and it gets removed, downgrade to 4.5 and needs to be set explicitly on downgrade?

joelddiaz:
@sdodson Yes. I built my own release image with this change, and was able to downgrade from 4.6 successfully (after fixing my typo in the original PR). The Deployment no longer references the serviceAccount that is new in 4.6, and instead references the `default` serviceAccount as was used pre-4.6.

```
  schedulerName: default-scheduler
  securityContext:
    fsGroup: 1000130000
    seLinuxOptions:
      level: s0:c11,c10
  serviceAccount: default
  serviceAccountName: default
  terminationGracePeriodSeconds: 10
  tolerations:
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
    operator: Exists
```

It appears the CVO is not doing a full replace of the objects, but is merely patching them which is why we need to explicitly set the values so they get reverted on downgrade.

sdodson:
/retitle Bug 1873345: specify a serviceaccount in the deployment

sdodson:
I've overridden bugzilla/valid-bug because this bug only happens when downgrading from 4.6 to 4.5, so while QE shall test that we've not regressed in 4.6 there's no code changes to be made on that branch and this bug was marked as a TestBlocker.

PR #244, opened 2020-09-10
Bug 1871713: Mark Manual Clusters  Upgradeable=False If 4.6 Cred Request Secrets Are Not Provisioned

dgoodwin:
There are two new credentials coming in 4.6, but if a manual mode user attempts to upgrade without them, the upgrade will fail.

This change encodes the secret names we know are coming, we ensure they exist and if not you are upgradeable=false. This condition does not affect 4.5.x upgrades, only 4.5 to 4.6.

Additionally to improve user experience, we are doing a hacky reconcile to get the status update that would clear upgradeable false when these secrets are created. This involves reconciling a fake cred request NamespacedName which we watch for and return early. In 4.7 we hope to break status updating out into it's own controller for a number of reasons, at which point this hack can be removed.

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -1237,10 +1237,9 @@ func (a *AWSActuator) Upgradeable(mode operatorv1.CloudCredentialsMode) *configv
 			upgradeableCondition.Status = configv1.ConditionFalse
 			upgradeableCondition.Reason = constants.MissingSecretsForUpgradeReason
 			upgradeableCondition.Message = fmt.Sprintf(
-				"Cannot upgrade manual mode cluster to %s due to missing secret(s): %v Please see documentation at: %s",
+				"Cannot upgrade manual mode cluster to %s due to missing secret(s): %v Please see the Manually Creating IAM for AWS OpenShift documentation.",
 				toRelease,
-				missingSecrets,
-				"https://docs.openshift.com/container-platform/4.5/installing/installing_aws/manually-creating-iam.html")
+				missingSecrets)
 		}
 	} else {
 		// If in mint or passthrough, make sure the root cred secret exists, if not it must be restored prior to upgrade.
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -1198,3 +1198,84 @@ func isAWSCredentials(providerSpec *runtime.RawExtension) (bool, error) {
 	}
 	return isAWS, nil
 }
+
+func (a *AWSActuator) Upgradeable(mode operatorv1.CloudCredentialsMode) *configv1.ClusterOperatorStatusCondition {
+	upgradeableCondition := &configv1.ClusterOperatorStatusCondition{
+		Status: configv1.ConditionTrue,
+		Type:   configv1.OperatorUpgradeable,
+	}
+	toRelease := "4.6"
+	if mode == operatorv1.CloudCredentialsModeManual {
+		// Check for the existence of credentials we know are coming in 4.6. If any do not exist, then we consider
+		// ourselves upgradable=false and inform the user why.
+		missingSecrets := []types.NamespacedName{}
+		for _, nsSecret := range a.GetUpcomingCredSecrets() {
+
+			secLog := log.WithField("secret", nsSecret).WithField("toRelease", toRelease)
+			secLog.Debug("checking that secrets exist for manual mode cluster upgrade")
+			secret := &corev1.Secret{}
+
+			err := a.Client.Get(context.Background(), nsSecret, secret)
+			if err != nil {
+				if errors.IsNotFound(err) {
+					secLog.Info("identified missing secret for upgrade")
+					missingSecrets = append(missingSecrets, nsSecret)
+				} else {
+					// If we can't figure out if you're upgradeable, you're not upgradeable:
+					secLog.WithError(err).Error("unexpected error looking up secret, marking upgradeable=false")
+					upgradeableCondition.Status = configv1.ConditionFalse
+					upgradeableCondition.Reason = constants.ErrorDeterminingUpgradeableReason
+					upgradeableCondition.Message = fmt.Sprintf("Error determining if cluster can be upgraded to %s: %v",
+						toRelease, err)
+					return upgradeableCondition
+				}
+			}
+		}
+
+		if len(missingSecrets) > 0 {
+			log.Infof("%d secrets missing, marking upgradeable=false", len(missingSecrets))
+			upgradeableCondition.Status = configv1.ConditionFalse
+			upgradeableCondition.Reason = constants.MissingSecretsForUpgradeReason
+			upgradeableCondition.Message = fmt.Sprintf(
+				"Cannot upgrade manual mode cluster to %s due to missing secret(s): %v Please see documentation at: %s",
+				toRelease,
+				missingSecrets,
+				"https://docs.openshift.com/container-platform/4.5/installing/installing_aws/manually-creating-iam.html")
+		}
+	} else {
+		// If in mint or passthrough, make sure the root cred secret exists, if not it must be restored prior to upgrade.
+		rootCred := types.NamespacedName{Namespace: rootAWSCredsSecretNamespace, Name: rootAWSCredsSecret}
+		secret := &corev1.Secret{}
+
+		err := a.Client.Get(context.Background(), rootCred, secret)
+		if err != nil {
+
+			if errors.IsNotFound(err) {
+				log.WithField("secret", rootCred).Info("parent cred secret must be restored prior to upgrade, marking upgradeable=false")
+				upgradeableCondition.Status = configv1.ConditionFalse
+				upgradeableCondition.Reason = constants.MissingRootCredentialUpgradeableReason
+				upgradeableCondition.Message = fmt.Sprintf("Parent credential secret must be restored prior to upgrade: %s/%s",
+					rootCred.Namespace, rootCred.Name)
+				return upgradeableCondition
+			}
+
+			log.WithError(err).Error("unexpected error looking up parent secret, marking upgradeable=false")
+			// If we can't figure out if you're upgradeable, you're not upgradeable:
+			upgradeableCondition.Status = configv1.ConditionFalse
+			upgradeableCondition.Reason = constants.ErrorDeterminingUpgradeableReason
+			upgradeableCondition.Message = fmt.Sprintf("Error determining if cluster can be upgraded to %s: %v",
+				toRelease, err)
+			return upgradeableCondition
+		}
+	}
+
+	return upgradeableCondition
+}
+
+func (a *AWSActuator) GetUpcomingCredSecrets() []types.NamespacedName {
+	return []types.NamespacedName{
+		// New in 4.6:
+		{Namespace: "openshift-cloud-credential-operator", Name: "cloud-credential-operator-s3"},
+		{Namespace: "openshift-cluster-csi-drivers", Name: "ebs-cloud-credentials"},
+	}
+}
--- a/pkg/azure/passthrough.go
+++ b/pkg/azure/passthrough.go
@@ -21,12 +21,18 @@ import (
 	"fmt"
 	"reflect"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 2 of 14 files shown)


dgoodwin:
WIP as I just want to take a cleanup pass tomorrow but this appears to be working well.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -86,6 +91,16 @@ func newReconciler(mgr manager.Manager, actuator actuator.Actuator, platType con
 	}
 }
 
+// isFutureCredRequestSecret is used to identify if a given secret namespace + name is one we're expecting in a future
+func isFutureCredRequestSecret(namespace, name string) bool {
+	for _, nsn := range awsactuator.GetUpcomingCredSecrets() {
```
joelddiaz, line 96: Should we just add GetUpcomingSecrets to the actuator interface and make this less AWS-specific?

pkg/operator/credentialsrequest/status.go
```diff
@@ -274,12 +276,7 @@ func computeStatusConditions(
 	conditions = clusteroperator.SetStatusCondition(conditions,
 		availableCondition)
 
-	// CCO doesn't have the idea of upgradeable vs not-upgradeable, but should report that condition nevertheless.
-	// Always be upgradeable.
-	upgradeableCondition := &configv1.ClusterOperatorStatusCondition{
-		Status: configv1.ConditionTrue,
-		Type:   configv1.OperatorUpgradeable,
-	}
+	upgradeableCondition := actuator.Upgradeable(mode)
```
joelddiaz, line 279: love it!

pkg/aws/actuator/actuator.go
```diff
@@ -1198,3 +1209,77 @@ func isAWSCredentials(providerSpec *runtime.RawExtension) (bool, error) {
...
+			log.Infof("%d secrets missing, marking upgradeable=false", len(missingSecrets))
+			upgradeableCondition.Status = configv1.ConditionFalse
+			upgradeableCondition.Reason = constants.MissingSecretsForUpgradeReason
+			upgradeableCondition.Message = fmt.Sprintf(
+				"Cannot upgrade manual mode cluster to %s due to missing secret(s): %v Please see documentation at: %s",
+				toRelease,
+				missingSecrets,
+				"https://docs.openshift.com/container-platform/4.5/installing/installing_aws/manually-creating-iam.html")
+		}
+	} else if mode == operatorv1.CloudCredentialsModeMint {
```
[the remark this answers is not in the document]
dgoodwin: Ah I wrongly assumed that mode calculation included a secret check and would be a reliable way to get the effective mode.

Excellent point though, I will make make it a blanket check, if not in manual then root secret is required for upgrade.

pkg/aws/actuator/actuator.go
```diff
@@ -1198,3 +1209,77 @@ func isAWSCredentials(providerSpec *runtime.RawExtension) (bool, error) {
 	}
 	return isAWS, nil
 }
+
+func (a *AWSActuator) Upgradeable(mode operatorv1.CloudCredentialsMode) *configv1.ClusterOperatorStatusCondition {
+	upgradeableCondition := &configv1.ClusterOperatorStatusCondition{
+		Status: configv1.ConditionTrue,
+		Type:   configv1.OperatorUpgradeable,
+	}
+	toRelease := "4.6"
```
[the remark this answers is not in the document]
dgoodwin: That looks like what you'll get if the list is empty to me, am I mistaken somewhere?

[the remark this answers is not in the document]
joelddiaz: Yes. I wrote this comment, and then forgot to remove this once I kept reading through the code and saw the looping over GetUpcomingCredSecrets().

pkg/operator/credentialsrequest/credentialsrequest_controller.go
dgoodwin: Done.

dgoodwin:
Ok I think feedback is addressed.

/assign @joelddiaz 
/cc @akhil-rane

joelddiaz (commented):
/lgtm

dgoodwin:
@joelddiaz sorry I added another separate commit, during demo prep I thought it might be a better demo to show that changing the mode immediately updated the condition, so we now watch the CloudCredential config and reconcile all CRs if anything changes.

dgoodwin:
As for bugzilla I'm going to clone the bz this week and get that ported to 4.6.

joelddiaz (commented):
/lgtm

dgoodwin:
Added a fix for the secret annotator controller erroring and requeuing in the perfectly valid state of the root admin cred being removed after minting.

Also switched the docs link from 4.5 to 4.6 after discussing with @jeana-redhat , this URL won't be valid until launch.

dgoodwin:
Updated per our slack discussion.

joelddiaz (commented):
/lgtm

dgoodwin:
We want this to merge after 4.6.

/hold

/bugzilla refresh

dgoodwin:
@sdodson so with this PR, I am essentially bringing the problem reported by @cgwalters in #248 back into 4.5. It is technically a separate issue where we should be instantiating a dummy actuator. Should this block this PR?

If so:

Option 1: Implement the dummy actuator and include in this 4.5 PR as it's technically going to cause that problem. I will also fix in 4.6 but via a separate PR aligned to a bug for that specific pre-existing issue.

Option 2: Skip upgradeable=false if you're in mint mode with a removed admin cred, but keep upgradeable=false if you're in manual mode and a 4.6 secret is missing. This would imply a mint move with removed admin cred user could enter into a 4.6 upgrade that will fail.

I think (1) is our best option but will take some more time.

dgoodwin:
Disregard previous, dummy actuator is already in play and we are backporting this code in the AWS actuator, so it should not introduce the problem.

pkg/aws/actuator/actuator.go
```diff
@@ -1198,3 +1198,84 @@ func isAWSCredentials(providerSpec *runtime.RawExtension) (bool, error) {
...
+				return upgradeableCondition
+			}
+
+			log.WithError(err).Error("unexpected error looking up parent secret, marking upgradeable=false")
+			// If we can't figure out if you're upgradeable, you're not upgradeable:
+			upgradeableCondition.Status = configv1.ConditionFalse
+			upgradeableCondition.Reason = constants.ErrorDeterminingUpgradeableReason
+			upgradeableCondition.Message = fmt.Sprintf("Error determining if cluster can be upgraded to %s: %v",
+				toRelease, err)
+			return upgradeableCondition
```
akhil-rane, line 1267: Is this return necessary considering we are returning right after the if-else block?

pkg/azure/passthrough.go
```diff
@@ -21,11 +21,14 @@ import (
...
 
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
 	minterv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
 	"github.com/openshift/cloud-credential-operator/pkg/operator/credentialsrequest/actuator"
 	actuatoriface "github.com/openshift/cloud-credential-operator/pkg/operator/credentialsrequest/actuator"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
```
akhil-rane, line 32: We have not separated the CCO and the other imports here

pkg/operator/constants/constants.go
```diff
@@ -43,6 +43,18 @@ const (
...
 
+	// MissingSecretsForUpgradeReason is used when a manual mode cluster is not upgradable due to missing
+	// secrets the operator knows will be required for the next minor release of OpenShift.
+	MissingSecretsForUpgradeReason = "ManualModeMissingSecrets"
+
+	// ErrorDeterminingUpgradeableReason is used when we encounter unexpected errors checking if a cluster can
+	// be updated.
+	ErrorDeterminingUpgradeableReason = "ErrorDeterminingUpgradeable"
+
+	// MissingRootCredentialUpgradeableReason is used a cluster is in mint mode with the root credential removed.
```
akhil-rane, line 54:
```suggestion
	// MissingRootCredentialUpgradeableReason is used when a cluster is in mint mode with the root credential removed.
```

pkg/operator/credentialsrequest/actuator/actuator.go
```diff
@@ -32,6 +36,12 @@ type Actuator interface {
 	Update(context.Context, *minterv1.CredentialsRequest) error
 	// Checks if the credentials currently exists.
 	Exists(context.Context, *minterv1.CredentialsRequest) (bool, error)
+	// Upgradeable returns a ClusterOperator Upgradeable condition to indicate whether or not this cluster can
+	// be safely upgraded to the next "minor" (4.y) Openshift release.
+	Upgradeable(operatorv1.CloudCredentialsMode) *configv1.ClusterOperatorStatusCondition
+	// GetUpcomingCredSecrets returns a slice of NamespacedNames for secrets holding credentials that we know are coming in the next OpenShift release.
+	// Used to pre-check if Upgradeable condition should be true or false for manual mode deployments of the credentials operator.
+	GetUpcomingCredSecrets() []types.NamespacedName
```
akhil-rane, line 44: Is this function temporary? How are we planning to maintain this function for future releases? For example 4.6 -> 4.7 will have different set of upcoming secrets. Does it makes sense to have something like  `GetUpcomingCredSecretsForRelease(release string)`

akhil-rane (commented):
Sorry for a late review. I completely missed it. Few minor comments.

pkg/aws/actuator/actuator.go
```diff
@@ -1198,3 +1198,84 @@ func isAWSCredentials(providerSpec *runtime.RawExtension) (bool, error) {
...
+				toRelease,
+				missingSecrets,
+				"https://docs.openshift.com/container-platform/4.5/installing/installing_aws/manually-creating-iam.html")
+		}
+	} else {
+		// If in mint or passthrough, make sure the root cred secret exists, if not it must be restored prior to upgrade.
+		rootCred := types.NamespacedName{Namespace: rootAWSCredsSecretNamespace, Name: rootAWSCredsSecret}
+		secret := &corev1.Secret{}
+
+		err := a.Client.Get(context.Background(), rootCred, secret)
```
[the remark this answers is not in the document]
dgoodwin: Good idea, tried just now but it looks like no, that function is wrapping the error in a way that I believe will break the IsNotFound logic below. Probably not worth the refactor.

pkg/aws/actuator/actuator.go
dgoodwin: Not necessary today, but I'm a little worried about things getting weird if someone comes later and adds a new check around line 1270 and doesn't realize that we kind of intended to return at line 1267. We'd proc this condition, not return, and then potentially overwrite this condition. That's not a crisis but it could be a little strange. Returning explicitly here after a problem seems better to me, not much downside other than 1 LOC.

pkg/azure/passthrough.go
dgoodwin: Fixing, thanks.

pkg/operator/constants/constants.go
dgoodwin: Fixing. Also saw another typo next line I'll get as well.

pkg/operator/credentialsrequest/actuator/actuator.go
dgoodwin: My thought was that we do not need to maintain past releases, once we're on 4.7, there's no need to keep checking for the secrets that were introduced in 4.7 because those are now handled by their respective CredentialsRequests. We now want to transition to 4.8, so I was thinking we just need one release at a time in here, whatever is coming next. The function should be updated every release but wouldn't cause major problems if we forgot because the secrets should exist by that point.

dgoodwin:
Updated for @akhil-rane comments, thanks for reviewing! A couple items I think are ok in current form but let me know if you disagree after reading.

sdodson:
/bugzilla refresh

Patch manger, FYI, this really needs to make the current window as this will be a prerequisite for upgrades to 4.6.

dgoodwin:
I swear this happened the last time I had a fix going back to 4.5 where suddenly the PR becomes corrupted and can't pass e2e, then re-opening a new one with the exact same branch passes immediately. A few more failures and I'll try this, we may need another approval.

sdodson:
There's some 36 PRs in the queue right now for 4.5, ever time something merges it will invalidate previous successful tests and require a retest. I wouldn't watch this pot boil, just check in on Monday.

PR #247, opened 2020-09-15
Bug 1878833: Update vendoring for 4.6

joelddiaz:
- [x] Update the version of controller-runtime being vendored (there is now a released version that supports the changes in k8s 1.19).
- [x] Update openshift/api vendoring to pull from the release-4.6 branch.
- [x] Update the CRD that we vendor from openshift/api
- [x] Update filenames for vendored CRDs (new version of openshift/build-machinery-go is now complaining during 'make verify' that it doesn't see us generate these files that we don't generate)
- [x] Update openshift/client-go and openshift/library-go to vendor from 4.6 branch


joelddiaz:
@dgoodwin alright, i *think* we've covered everything that needed updating (for the master branch)

bindata/bootstrap/cloudcredential_v1_operator_config_custresedef.yaml
```diff
@@ -0,0 +1,158 @@
+apiVersion: apiextensions.k8s.io/v1
```
dgoodwin, line 1: This file is appearing twice, custresdef and custresedef (an extra e)

dgoodwin (changes requested):

PR #248, opened 2020-09-16
Upgradeable false if upcoming secrets are not provisioned.

dgoodwin:
Sets the upgradeable false condition if you are in manual mode and
missing some secrets we know are coming in the next minor OpenShift
release. (i.e. 4.6 to 4.7) Also set upgradeable=false if you are in mind
mode and have removed your root admin cred.

Fixes an error and requeue in the secretannotator controller in the
perfectly valid state where the AWS root cred has been temporarily
removed.

Reconcile all CRs when CloudCredential config changes.

/assign @joelddiaz @akhil-rane


cgwalters:
I can't install a libvirt cluster because
```
    - lastTransitionTime: "2020-09-17T12:59:07Z"
      message: 'Cluster operator cloud-credential cannot be upgraded between minor
        versions: Parent credential secret kube-system/aws-creds must be restored
        prior to upgrade'
      reason: CredentialsRootSecretMissing
      status: "False"
      type: Upgradeable
```

Does this PR fix that?

dgoodwin:
@cgwalters can you clarify, I wouldn't expect you could see that error unless you'd already successfully installed a cluster.

cgwalters:
> @cgwalters can you clarify, I wouldn't expect you could see that error unless you'd already successfully installed a cluster.

This happens while trying to do a fresh install.

cgwalters:
> This happens while trying to do a fresh install.

IOW I suspect you need some platform-specific logic here, i.e. don't go upgradeable=false if the aws secret doesn't exist if we're not even on AWS.

pkg/aws/actuator/actuator_test.go
```diff
@@ -271,6 +275,98 @@ func TestGenerateUserName(t *testing.T) {
...
+		},
+		{
+			name:           "implicit mint mode with missing root cred",
+			mode:           v1.CloudCredentialsModeDefault,
+			existing:       []runtime.Object{},
+			expectedStatus: configv1.ConditionFalse,
+			expectedReason: constants.MissingRootCredentialUpgradeableReason,
+		},
+		{
+			name: "manual mode with 4.6 secrets pre-provisioned",
```
joelddiaz, line 325: maybe update the name to something like `manual mode with future version secrets pre-provisioned`

pkg/aws/actuator/actuator_test.go
```diff
@@ -271,6 +275,98 @@ func TestGenerateUserName(t *testing.T) {
...
+				testSecret("kube-system", "bar"),
+			},
+			expectedStatus: configv1.ConditionTrue,
+			overrideUpcomingSecrets: []types.NamespacedName{
+				{Namespace: "kube-system", Name: "foo"},
+				{Namespace: "kube-system", Name: "bar"},
+			},
+		},
+		{
+			name:           "manual mode without 4.6 secrets pre-provisioned",
```
joelddiaz, line 338: same comment about making the name non-version specific

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -69,6 +69,10 @@ const (
 	cloudCredDeprovisionSuccess = "CloudCredDeprovisionSuccess"
 
 	credentialsRequestInfraMismatch = "InfrastructureMismatch"
+
+	// Hack for 4.5 to get us a status update when we don't have a cred request to reconcile.
```
joelddiaz, line 73:
```suggestion
	// Hack to send a fake CredReq to get us a status update when we don't have a cred request to reconcile.
```

dgoodwin:
@cgwalters how is this blocking an install? Does upgradeable=false trigger an install failure? That shouldn't be the case, @sdodson is going to confirm with OTA team.

This change does not fix your issue though, in fact the 4.5 equivalent of this PR is about to backport this problem into 4.5. This change only adjusts the error message.

Your issue (assuming we can figure out why it's causing an install failure) is related to some logic where by default we assume to run the AWS actuator. This decision was taken when the initial Azure implementation was contributed and AWS remained the default so as not to cause any unintended breakages. The aws actuator probably should NOT be getting instantiated on libvirt. If we don't know the cloud, we probably should be instantiating a dummy actuator that just does nothing.

@cgwalters could you file a bugzilla for this?

Jumping over to the 4.5 PR, will tag you both there so I can figure out if it's safe to bring that back or not.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -129,12 +143,26 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	actuator := (r.(*ReconcileCredentialsRequest)).Actuator
+
 	// Define a mapping for secrets to the credentials requests that created them. (if applicable)
 	// We use an annotation on secrets that refers back to their owning credentials request because
 	// the normal owner reference is not namespaced, and we want to support credentials requests being
 	// in a centralized namespace, but writing secrets into component namespaces.
 	targetCredSecretMapFunc := handler.ToRequestsFunc(
 		func(a handler.MapObject) []reconcile.Request {
+
+			// Hack for 4.5, reconcile a fake cred request we're watching for in the Reconcile loop, intended to be
```
joelddiaz, line 155:
```suggestion
			// Hack: reconcile a fake cred request we're watching for in the Reconcile loop, intended to be
```

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -373,6 +424,13 @@ func (r *ReconcileCredentialsRequest) Reconcile(request reconcile.Request) (reco
 		metrics.MetricControllerReconcileTime.WithLabelValues(controllerName).Observe(dur.Seconds())
 	}()
 
+	// Hack for 4.5, reconcile a fake cred request we're watching for in the Reconcile loop, intended to be
```
joelddiaz, line 427:
```suggestion
	// Hack: reconcile a fake cred request we're watching for in the Reconcile loop, intended to be
```

joelddiaz:
IIUC, I think this removal resolves @cgwalters 's question https://github.com/openshift/cloud-credential-operator/pull/248/files#diff-845bfb17a5cd1ebfab38dbff6fbe7e21L199-L211
That was non-platform specific code.

pkg/operator/credentialsrequest/status_test.go
```diff
@@ -184,14 +184,6 @@ func TestClusterOperatorStatus(t *testing.T) {
 			cloudPlatform:      configv1.GCPPlatformType,
 			expectedConditions: []configv1.ClusterOperatorStatusCondition{},
 		},
-		{
-			name:              "upgradable false if parent cred removed",
-			credRequests:      []minterv1.CredentialsRequest{},
-			parentCredRemoved: true,
```
joelddiaz, line 190: the parentCredRemoved field from the test cases struct now appears to be unused.

pkg/operator/credentialsrequest/status_test.go
```diff
@@ -210,11 +202,19 @@ func TestClusterOperatorStatus(t *testing.T) {
...
+				operatorMode = v1.CloudCredentialsModeManual
+			}
+			clusterOperatorConditions := computeStatusConditions(
+				dummyActuator,
+				operatorMode,
+				test.credRequests,
+				test.cloudPlatform,
 				log.WithField("test", test.name))
+
+			//schemeutils.SetupScheme(scheme.Scheme)
```
joelddiaz, line 217: leftover debug code?

pkg/operator/credentialsrequest/status.go
```diff
@@ -40,15 +40,16 @@ func (r *ReconcileCredentialsRequest) syncOperatorStatus() error {
 var _ clusteroperator.StatusHandler = &ReconcileCredentialsRequest{}
 
 func (r *ReconcileCredentialsRequest) GetConditions(logger log.FieldLogger) ([]configv1.ClusterOperatorStatusCondition, error) {
-	_, credRequests, operatorIsDisabled, err := r.getOperatorState(logger)
+	_, credRequests, mode, err := r.getOperatorState(logger)
 	if err != nil {
 		return []configv1.ClusterOperatorStatusCondition{}, fmt.Errorf("failed to get operator state: %v", err)
 	}
-	parentSecretExists, err := r.parentSecretExists()
```
joelddiaz, line 47: parentSecretExists() seems to no longer be used

dgoodwin:
My mistake, great catch thanks @joelddiaz. @cgwalters looks like this *should* fix your issue after all due to where the code moved. We do default to a dummy actuator already, this status checking code was in the controller previously and this pr moves it to an aws specific location. Will address feedback and get this in so you can confirm.

dgoodwin:
PR updated.

joelddiaz (commented):
/lgtm

PR #255, opened 2020-10-05
move status calculation into its own controller

joelddiaz:
Rather than have each controller responsible for triggering a status calculation, move that responsiblity into its own controller.

- [x] Store the upcoming list of Secrets as a shared constant/variable (for use by both the Actuator's Upgradeable() and the status controller's Secret watch list).
- [x] Remove GetUpcomingSecrets() from the Actuator interface.
- [x] Stop setting default ClusterOperator status condition values from the Actuator's Upgradeable() API call (it is already set in the current status calculation codepath).
- [x]  Update controllers to register themselves to the new status controller for status calculation callbacks.
- [x] Convert previous pkg/util/status into the new status controller (now in pkg/operator/status).
- [x] Remove hack of sending fake CredentialsRequests just to trigger a status sync from credentialsrequest controller (new status calculator now will watch and react appropriately).
- [x] Update credentialsrequest test cases to only check for the setting of non-default conditions.
- [x] Update credentialsrequest test cases to no longer need to register and deregister status callbacks.
- [x] Move credentialsrequest status test cases to new status controller.
- [x] Update secretannotator actuators to no longer trigger a status sync.

Note: for situations where the status controller cannot watch for certain conditions that should cause a status recalculation, the controller can annotate the CCO's operator config object to add a timestamp (which will in turn trigger a status calc).
    
Also note: Ensure the managementState field is set on our CCO operator config (otherwise we run into problems trying to annotate the operator config). The operator config should not allow an empty string for the managementState field, but it absolutely is not set at all (which shows up as being set to the empty string after doing a Get() of the operator config). Detect this, and assume that the empty string means a managementState of "Managed", and send back "Managed" on Update().


pkg/operator/status/status_controller.go
```diff
@@ -0,0 +1,404 @@
...
+	defer func() {
+		dur := time.Since(start)
+		metrics.MetricControllerReconcileTime.WithLabelValues(controllerName).Observe(dur.Seconds())
+	}()
+
+	err := SyncStatus(r.Client, r.Logger)
+	return reconcile.Result{}, err
+}
+
+// SyncStatus allows triggering a status sync from outside of this controller.
```
dgoodwin, line 168: I'm really not sure if this is a bad thing. Seems fairly elegant, aside from error handling and requeues.

pkg/operator/status/status_controller.go
```diff
@@ -0,0 +1,404 @@
...
+func (r *ReconcileStatus) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	start := time.Now()
+
+	defer func() {
+		dur := time.Since(start)
+		metrics.MetricControllerReconcileTime.WithLabelValues(controllerName).Observe(dur.Seconds())
+	}()
+
+	err := SyncStatus(r.Client, r.Logger)
+	return reconcile.Result{}, err
```
dgoodwin, line 165: What do you think about returning a periodic requeue here, just to ensure status is pretty much always up to date? Maybe every 5 minutes?

pkg/operator/status/status_controller.go
```diff
@@ -0,0 +1,404 @@
...
+			logger.WithField("statushandler", handlerName).Errorf("failed to get related objects from status handler")
+			continue
+		}
+		relatedObjects = append(relatedObjects, handlerRelatedObjects...)
+	}
+
+	// sets defaults for condition not set by any handler
+	conditions = defaultUnsetConditions(conditions)
+
+	// at this point we know all conditoin types exist in conditions
```
dgoodwin, line 207: s/conditoin/condition

pkg/operator/constants/constants.go
```diff
@@ -118,4 +123,19 @@ var (
 		ModeDegraded,
 		ModeUnknown,
 	}
+
+	// Add known new credentials for next version upgrade
```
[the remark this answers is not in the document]
joelddiaz: Mostly seemed weird to instantiate the entire credentials actuator to just get a static list of objects. My first pass had me setting up the status controller to embed the actuator, but I ended up with this instead.

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -204,8 +204,12 @@ func (r *s3EndpointReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
 	// SyncStatus only if our degraded condition changed
 	defer func(initialDegradedCondition configv1.ClusterOperatorStatusCondition) {
-		if !clusteroperator.ConditionEqual(initialDegradedCondition, r.degraded) {
-			clusteroperator.SyncStatus(r.controllerRuntimeClient, r.logger)
+		if !status.ConditionEqual(initialDegradedCondition, r.degraded) {
+			// Would be nice to not have to trigger a status sync from this controller.
+			// Need to be able to force a status calculation here because the status
+			// controller has no Watch that could cause it to run for this situation.
+			// Could we annotate our secret to trigger a status calc???
+			status.SyncStatus(r.controllerRuntimeClient, r.logger)
```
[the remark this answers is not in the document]
joelddiaz: I'm really annoyed by this one. This thought occurred to me this morning: we could annotate the config CR with the current datestamp from the oidc controller???

pkg/operator/constants/constants.go
```diff
@@ -118,4 +123,19 @@ var (
...
+	AWSUpcomingSecrets = []types.NamespacedName{}
+	// AzureUpcomingSecrets contains the list of known new Azure credential secrets for the next version of OpenShift
+	AzureUpcomingSecrets = []types.NamespacedName{}
+	// GCPUpcomingSecrets contains the list of known new GCP credential secrets for the next version of OpenShift
+	GCPUpcomingSecrets = []types.NamespacedName{}
+	// OpenStackUpcomingSecrets contains the list of known new OpenStack credential secrets for the next version of OpenShift
+	OpenStackUpcomingSecrets = []types.NamespacedName{}
+	// OvirtUpcomingSecrets contains the list of known new oVirt credential secrets for the next version of OpenShift
+	OvirtUpcomingSecrets = []types.NamespacedName{}
+	// VsphereUpcomingSecrets contains the list of known new Vsphere credential secrets for the next version of OpenShift
```
akhil-rane, line 139:
```suggestion
	// VsphereUpcomingSecrets contains the list of known new vSphere credential secrets for the next version of OpenShift
```

joelddiaz:
@dgoodwin un-WIPed as I finally tested this and it looked fine to me. The working (but IMO non-ideal) syncing from the oidc controller is still in the current version of the PR.

pkg/operator/oidcdiscoveryendpoint/controller.go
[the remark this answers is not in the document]
dgoodwin: When this controller sees the degraded state change, annotated the CloudConfig CR with a timestamp to force a reconcile? That would put it into the normal controller flow, with error handling etc. Seems like a good idea to me.

pkg/operator/status/status_controller.go
```diff
@@ -0,0 +1,407 @@
...
+	return nil
+}
+
+func mergeConditions(existing, new []configv1.ClusterOperatorStatusCondition, handlerName string, logger log.FieldLogger) []configv1.ClusterOperatorStatusCondition {
+	for _, newCondition := range new {
+		existingCondition, index := findClusterOperatorCondition(existing, newCondition.Type)
+		if existingCondition == nil {
+			existing = append(existing, newCondition)
+		} else {
+			logger.WithField("statushandler", handlerName).Infof("condition already set for type %s by a previous handler, the new condition from the current handler will be accepted: %v", newCondition.Type, newCondition)
```
dgoodwin, line 273: Last write wins? Could you clarify this in the function comment a little. This should probably be a warning as well as info is being lost. The suggestion above for logging every condition should help when combined with that.

pkg/operator/status/status_controller.go
```diff
@@ -0,0 +1,407 @@
...
+	// Handlers return abnormal (non-default) conditions they wish to set.
+	// if the controller is functioning normally, it should return an empty slice of conditions.
+	conditions := []configv1.ClusterOperatorStatusCondition{}
+	relatedObjects := []configv1.ObjectReference{}
+	for handlerName, handler := range statusHandlers {
+		handlerConditions, err := handler.GetConditions(logger)
+		if err != nil {
+			logger.WithError(err).WithField("statushandler", handlerName).Errorf("failed to get conditions from status handler")
+			continue
+		}
```
dgoodwin, line 197: Could you info log each condition we got here and the status handler it came from. Then below when we see a conflict we can log that as a warning and still know the user/admin can see which handlers reported the same condition.

pkg/operator/status/status_controller.go
```diff
@@ -0,0 +1,407 @@
...
+		rootSecret.Name = constants.OpenStackCloudCredsSecretName
+		secrets = append(secrets, constants.OpenStackUpcomingSecrets...)
+	case configv1.OvirtPlatformType:
+		rootSecret.Name = constants.OvirtCloudCredsSecretName
+		secrets = append(secrets, constants.OvirtUpcomingSecrets...)
+	case configv1.VSpherePlatformType:
+		rootSecret.Name = constants.VSphereCloudCredSecretName
+		secrets = append(secrets, constants.VsphereUpcomingSecrets...)
+	default:
+		log.Errorf("unable to provide upcoming secrets for unknown platform: %v", platformType)
```
dgoodwin, line 401: This looks like it will be called on bare metal, so hitting this line is really not an error IMO, I'd go with info (or maybe even debug) and soften the message up a little bit.

pkg/operator/status/status_controller_test.go
```diff
@@ -0,0 +1,289 @@
...
+					conditionType: configv1.OperatorProgressing,
+					status:        corev1.ConditionFalse,
+				},
+				{
+					conditionType: configv1.OperatorDegraded,
+					status:        corev1.ConditionFalse,
+				},
+			},
+		},
+	}
```
dgoodwin, line 204: Coverage seems a little light in here, could we get something to simulate two handlers returning the same condition? Or is the handler registration making this infeasible?

dgoodwin (changes requested):

pkg/operator/oidcdiscoveryendpoint/controller.go
```diff
@@ -199,17 +203,24 @@ type s3EndpointReconciler struct {
...
-			// Would be nice to not have to trigger a status sync from this controller.
-			// Need to be able to force a status calculation here because the status
-			// controller has no Watch that could cause it to run for this situation.
-			// Could we annotate our secret to trigger a status calc???
-			status.SyncStatus(r.controllerRuntimeClient, r.logger)
+			// There are situations that cause this controller to go degraded that the status controller
+			// cannot watch for. So just update the config CR (which the status controller does watch)
+			// to trigger a status recalculation.
+			if err := r.updateConfigCRTimestamp(); err != nil {
+				r.logger.WithError(err).Error("failed to updated config CR to trigger status caluclation")
```
dgoodwin, line 216: small calculation typo here

dgoodwin (commented):
One small typo but I think we're gtg otherwise, please squash.

dgoodwin:
/lgtm

Nice work, looking forward to this as I have mass conflicts every time my creds rotate. :)

PR #256, opened 2020-10-06
Stop reporting mode metrics for unused modes.

dgoodwin:
We now will report 1 for any modes in use, and anything not in use will
be explicitly cleared in case we transitioned modes since the pod was
started.

This is to make the metric easier to consume across the entire fleet.

/assign @joelddiaz 
/cc @akhil-rane


dgoodwin:
Worth noting, status does not seem to be updating in master if you edit the mode, at least not quickly.

joelddiaz (commented):
/lgtm

PR #257, opened 2020-10-07
allow skipping leader election

joelddiaz:
When hard-stopping a debugging session, the leader election cleanup doesn't get to run. Provide a way to skip all the leader election startup to speed up the development loop.

### DIFF
--- a/pkg/cmd/operator/cmd.go
+++ b/pkg/cmd/operator/cmd.go
@@ -139,32 +139,36 @@ func NewOperator() *cobra.Command {
 				},
 			}
 
-			// start the leader election code loop
-			leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
-				Lock:            lock,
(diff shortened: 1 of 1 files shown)


joelddiaz:
> Do we need a doc update for this in [Developer Instructions](https://github.com/openshift/cloud-credential-operator#developer-instructions)?

The docs don't really address running with a debugger (which is the only reason I find this to be useful)... As long as CCO is shut down properly there shouldn't be any unnecessary waiting for leader elections, right?

akhil-rane:
> > Do we need a doc update for this in [Developer Instructions](https://github.com/openshift/cloud-credential-operator#developer-instructions)?
> 
> The docs don't really address running with a debugger (which is the only reason I find this to be useful)... As long as CCO is shut down properly there shouldn't be any unnecessary waiting for leader elections, right?

Yes, if CCO is shutdown properly there is no waiting :)

PR #259, opened 2020-10-07
enable GCP read only creds to be used when root creds missing

joelddiaz:
- [x] add new GCP read-only CredentialsRequest manifests
- [x] implement the ability to use the CCO's read-only GCP creds for verifying that creds need no updates
- [x] actually skip the service API checks when the CredReq says to skip the checking
- [x] add test case to cover the root creds missing but the read-only creds are available
- [x] update readme to mark GCP as supported for removing the root creds
- [x] (unrelated to GCP) return unexpected error in AWS actuator

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Cons:
   * Still requires admin credential in the cluster for brief periods of time.
   * Requires manually re-instating the Secret with admin credentials for each upgrade.
 
-Supported clouds: AWS
+Supported clouds: AWS, GCP
 
 ## 2. Passthrough Mode
 
@@ -200,7 +200,7 @@ Cloud | Mint | Mint + Remove Admin Cred | Passthrough | Manual | Token
 --- | --- | --- | --- | --- | ---
 AWS | Y | 4.4+ | Y | 4.3+ | 4.6+ (expected)
 Azure | Y | N | Y | unknown | N
-GCP | Y | N | Y | unknown | N
+GCP | Y | 4.7+ | Y | unknown | N
 OpenStack | N | N | Y | N | N
 oVirt | N | N | Y | N | N
 VMWare | N | N | Y | N | N
--- a/pkg/gcp/actuator/actuator.go
+++ b/pkg/gcp/actuator/actuator.go
@@ -48,7 +48,7 @@ import (
 
 const (
 	roGCPCredsSecretNamespace = "openshift-cloud-credential-operator"
-	roGCPCredsSecret          = "cloud-credential-operator-iam-ro-creds"
+	roGCPCredsSecret          = "cloud-credential-operator-gcp-ro-creds"
 
 	gcpSecretJSONKey = "service_account.json"
 )
@@ -320,16 +320,23 @@ func (a *Actuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequest
 		return fmt.Errorf("error gathering permissions for each role: %v", err)
 	}
 
-	// check that service APIs are enabled before we bother to make a
-	// new serviceaccount
-	serviceAPIsEnabled, err := checkServicesEnabled(rootGCPClient, permList, logger)
-	if err != nil {
-		return err
+	var serviceAPIsEnabled bool
+	if gcpSpec.SkipServiceCheck {
+		// Since we are skipping the checks, we will assume that the APIS are enabled.
+		serviceAPIsEnabled = true
+	} else {
+		// check that service APIs are enabled before we bother to make a
+		// new serviceaccount
+		var err error
+		serviceAPIsEnabled, err = checkServicesEnabled(rootGCPClient, permList, logger)
+		if err != nil {
+			return err
+		}
 	}
-	if !gcpSpec.SkipServiceCheck && !serviceAPIsEnabled {
+
+	if !serviceAPIsEnabled {
 		return fmt.Errorf("not all required service APIs are enabled")
 	}
-	serviceAPIsEnabled = true
 
 	// Create service account if necessary
 	var serviceAccount *iamadminpb.ServiceAccount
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -200,7 +200,7 @@ Cloud | Mint | Mint + Remove Admin Cred | Passthrough | Manual | Token
 --- | --- | --- | --- | --- | ---
 AWS | Y | 4.4+ | Y | 4.3+ | 4.6+ (expected)
 Azure | Y | N | Y | unknown | N
-GCP | Y | N | Y | unknown | N
+GCP | Y | Y | Y | unknown | N
```
dgoodwin, line 203: This should be 4.7+ rather than Y.

pkg/gcp/actuator/actuator.go
```diff
@@ -48,7 +48,7 @@ import (
 
 const (
 	roGCPCredsSecretNamespace = "openshift-cloud-credential-operator"
-	roGCPCredsSecret          = "cloud-credential-operator-iam-ro-creds"
+	roGCPCredsSecret          = "cloud-credential-operator-gcp-ro-creds"
```
dgoodwin, line 51: What was this var doing here already? Looks like there was some half r/o support?

pkg/gcp/actuator/actuator.go
```diff
@@ -493,52 +505,42 @@ func (a *Actuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequ
...
+func (a *Actuator) buildReadGCPClient(cr *minterv1.CredentialsRequest) (ccgcp.Client, error) {
+	logger := a.getLogger(cr).WithField("secret", fmt.Sprintf("%s/%s", roGCPCredsSecretNamespace, roGCPCredsSecret))
+	logger.Debug("loading GCP read-only credentials from secret")
 
-	logger.Debug("loading GCP credentials from secret")
-	jsonBytes, err := loadCredsFromSecret(a.Client, constants.CloudCredSecretNamespace, constants.GCPCloudCredSecretName)
+	jsonBytes, err := loadCredsFromSecret(a.Client, roGCPCredsSecretNamespace, roGCPCredsSecret)
 	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.Warn("read-only creds not found, using root creds cllient")
```
dgoodwin, line 522: cllient

pkg/operator/credentialsrequest/credentialsrequest_controller_gcp_test.go
```diff
@@ -332,6 +333,50 @@ func TestCredentialsRequestGCPReconcile(t *testing.T) {
...
+
+				return mockGCPClient
+			},
+			validate: func(c client.Client, t *testing.T) {
+				cr := getCredRequest(c)
+
+				lastSynced := cr.Status.LastSyncTimestamp.Time
+				assert.WithinDuration(t, time.Now(), lastSynced, time.Second*5, "expected a recent last synced status")
+			},
+		},
```
dgoodwin, line 379: Could we get tests for: cred needs a modification but no root cred, cred is missing entirely but no root cred.

pkg/gcp/actuator/actuator.go
```diff
@@ -493,52 +505,42 @@ func (a *Actuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequ
...
+			return a.buildRootGCPClient(cr)
+		}
+		logger.WithError(err).Error("failed to load in read-only creds Secret")
 		return nil, err
 	}
 
-	logger.Debug("creating root GCP client")
+	logger.Debug("creating read GCP client")
 	return a.GCPClientBuilder(a.ProjectName, jsonBytes)
 }
```
dgoodwin, line 531: Looks like we also need to set upgradeable=false when that root cred is gone somewhere in the status area. Is it feasible to implement that now in the status handler in this branch, or would we be better waiting until the status controller merges?

dgoodwin (changes requested):

pkg/gcp/actuator/actuator.go
joelddiaz: Yes. When I originally wrote the GCP support, I apparently plumbed just about everything through for read-only support and we were just missing an appropriate role that we could use. There was even an unused buildReadOnlyGCPClient() already written ¯\_(ツ)_/¯

pkg/gcp/actuator/actuator.go
joelddiaz: Good catch! I've put int he upgradeable() implementation to watch for the missing root creds, although one of these PRs might need some changes depending on which one gets in first.

dgoodwin:
Looks good, lets squash and I'll lgtm. Thanks!

pkg/gcp/actuator/actuator.go
```diff
@@ -775,6 +777,30 @@ func (a *Actuator) Upgradeable(mode operatorv1.CloudCredentialsMode) *configv1.C
...
 	}
+	toRelease := "4.7"
+
+	rootCred := a.GetCredentialsRootSecretLocation()
+	secret := &corev1.Secret{}
+
+	err := a.Client.Get(context.TODO(), rootCred, secret)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			log.WithField("secret", rootCred).Info("parent cred secret must be restored prior to upgrade, marking upgradable=false")
```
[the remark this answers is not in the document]
joelddiaz: Good question. I just copied what we presently use for the AWS actuator https://github.com/openshift/cloud-credential-operator/blob/master/pkg/aws/actuator/actuator.go#L1251

joelddiaz:
@dgoodwin don't know what it means for you to have outstanding changes requested on the PR, but I think we're good to go minus the question that @akhil-rane brought up about the text we use in the logging.

pkg/gcp/actuator/actuator.go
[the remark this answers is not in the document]
dgoodwin: I can't remember either. :) We should decide this after a call someday and put it in the readme.

PR #260, opened 2020-10-12
Add kubevirt platform

bardielle:
Our main goal is creating a tenant cluster on top of existing OpenShift/Kubernetes cluster by creating virtualMachines (using kubevirt) for every node in the tenant cluster (masters and also workers nodes) 
You can find more information here: https://github.com/openshift/enhancements/pull/417


gregsheremeta:
Hi, is there a Jira for this work? I would like to better understand the goal. I'm surprised to see kubevirt treated as a cloud platform -- I thought it was more of an addon to OCP.

bardielle:
@gregsheremeta The epic is https://issues.redhat.com/browse/OCPCNV-1 
The main goal is creating a nested openshift cluster (called tenant cluster) over an openshift cluster (called infra cluster) using kubevirt

pkg/kubevirt/actuator.go
```diff
@@ -0,0 +1,275 @@
...
+
+	return &KubevirtActuator{
+		Codec:  codec,
+		Client: client,
+	}, nil
+}
+
+// Exists checks if the credentials currently exist.
+// TODO: in the future validate the expiration of the credentials
+func (a *KubevirtActuator) Exists(ctx context.Context, cr *minterv1.CredentialsRequest) (bool, error) {
```
ravidbro, line 63: Don't you need to validate that the value inside the secret is as expected?

ravidbro (commented):
In general, I don't understand why do we need so much of code duplication between the actuators when the difference is only how to create the data for the secret when needed but all the flow is the same.
Of course, not for this PR

pkg/kubevirt/actuator.go
bardielle: The "getSecret" function validate that the secret contains the required key in the data
Moreover, in the other actoators there isn't such a validation

pkg/kubevirt/actuator.go
```diff
@@ -0,0 +1,275 @@
...
+	minterv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	kubernetesErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type KubevirtActuator struct {
+	Client client.Client
+	Codec  *minterv1.ProviderCodec
```
joelddiaz, line 41: i don't see this used anywhere in the non-test files

pkg/kubevirt/actuator.go
```diff
@@ -0,0 +1,275 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	minterv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	kubernetesErrors "k8s.io/apimachinery/pkg/api/errors"
```
joelddiaz, line 35: we don't generally use mixed case import naming ( https://blog.golang.org/package-names )

pkg/kubevirt/actuator.go
```diff
@@ -0,0 +1,275 @@
...
+		Client: client,
+	}, nil
+}
+
+// Exists checks if the credentials currently exist.
+// TODO: in the future validate the expiration of the credentials
+func (a *KubevirtActuator) Exists(ctx context.Context, cr *minterv1.CredentialsRequest) (bool, error) {
+	logger := a.getLogger(cr)
+	logger.Debug("running Exists")
+	var err error
```
joelddiaz, line 66: this declaration doesn't appear necessary

pkg/kubevirt/actuator.go
```diff
@@ -0,0 +1,275 @@
...
+		return nil, &actuatoriface.ActuatorError{
+			ErrReason: minterv1.CredentialsProvisionFailure,
+			Message:   fmt.Sprintf("%v: %v", msg, err),
+		}
+	}
+
+	// get the secret data - the infra kubeconfig
+	infraClusterKubeconfig, ok := kubevirtCredentialsSecret.Data[KubevirtCredentialsSecretKey]
+	if !ok {
+		return nil, errors.New("invalid mode")
```
joelddiaz, line 158: can we be more clear than "invalid mode". Maybe something like "credentials missing expected key".

pkg/kubevirt/actuator.go
```diff
@@ -0,0 +1,275 @@
...
+			},
+		},
+		Data: map[string][]byte{
+			KubevirtCredentialsSecretKey: *kubevirtCredentialData,
+		},
+	}
+
+	if err := a.Client.Create(ctx, secret); err != nil {
+		logger.WithError(err).Error("error creating secret")
+		return err
```
joelddiaz, line 230: probably should built up an ActuatorError here to return back to the credentialsrequest controller.

pkg/kubevirt/actuator_test.go
```diff
@@ -0,0 +1,400 @@
...
+limitations under the License.
+*/
+
+package kubevirt_test
+
+import (
+	"context"
+	"fmt"
+	"github.com/openshift/cloud-credential-operator/pkg/kubevirt"
+	kubernetesErrors "k8s.io/apimachinery/pkg/api/errors"
```
joelddiaz, line 23: mixed case import

pkg/kubevirt/actuator_test.go
```diff
@@ -0,0 +1,400 @@
...
+	openshiftapiv1 "github.com/openshift/api/config/v1"
+	minterv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+	"github.com/openshift/cloud-credential-operator/pkg/operator/constants"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	kubernetesClient "sigs.k8s.io/controller-runtime/pkg/client"
```
joelddiaz, line 36: mixed case import

pkg/kubevirt/actuator_test.go
```diff
@@ -0,0 +1,400 @@
...
+}
+
+func TestCreateCR(t *testing.T) {
+	if err := openshiftapiv1.Install(scheme.Scheme); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := minterv1.SchemeBuilder.AddToScheme(scheme.Scheme); err != nil {
+		t.Fatal(err)
+	}
```
joelddiaz, line 105: fyi, there is a utility that does all the scheme setup for this operator. https://github.com/openshift/cloud-credential-operator/blob/master/pkg/util/scheme.go

pkg/kubevirt/actuator_test.go
```diff
@@ -0,0 +1,400 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:                       constants.KubevirtCloudCredSecretName,
+			Namespace:                  constants.CloudCredSecretNamespace,
+		},
+		Data:       map[string][]byte{
+			kubevirt.KubevirtCredentialsSecretKey: kubevirtCredentialData,
+		},
+	}
+
+	kubevirtOpenshiftMmachineApiSecret = corev1.Secret{
```
joelddiaz, line 64:
```suggestion
	kubevirtOpenshiftMachineApiSecret = corev1.Secret{
```

pkg/kubevirt/actuator_test.go
```diff
@@ -0,0 +1,400 @@
...
+		errRegexp string
+		validate                   func(*testing.T, kubernetesClient.Client)
+	}{
+		{
+			name:               "Create CR happy flow",
+			existing:           defaultExistingObjects(),
+			credentialsRequest: testCredentialsRequest(t),
+			validate: func(t *testing.T, c kubernetesClient.Client) {
+				cr := getCredRequest(t, c)
+				assert.NotEqual(t, cr, nil)
```
joelddiaz, line 121: assert.NotNil()

pkg/kubevirt/actuator_test.go
```diff
@@ -0,0 +1,400 @@
...
+
+				secret := getSecret(t, c)
+				assert.NotEqual(t, secret, nil)
```
joelddiaz, line 124: assert.NotNil()

pkg/kubevirt/actuator_test.go
```diff
@@ -0,0 +1,400 @@
...
+		validate           func(*testing.T, kubernetesClient.Client, bool)
+	}{
+		{
+			name:               "Exists CR happy flow (true)",
+			existing:           existingObjectsAfterCreate(t),
+			credentialsRequest: testCredentialsRequest(t),
+			validate:           func(t *testing.T, c kubernetesClient.Client, isExists bool) {
+				secret := getSecret(t, c)
+				assert.Equal(t, isExists, true)
+				assert.NotEqual(t, secret, nil)
```
joelddiaz, line 235: assert.NotNil()

pkg/kubevirt/actuator_test.go
```diff
@@ -0,0 +1,400 @@
...
+		errRegexp string
+		validate                   func(*testing.T, kubernetesClient.Client)
+	}{
+		{
+			name:               "Update CR happy flow - non exists",
+			existing:           defaultExistingObjects(),
+			credentialsRequest: testCredentialsRequest(t),
+			validate: func(t *testing.T, c kubernetesClient.Client) {
+				cr := getCredRequest(t, c)
+				assert.NotEqual(t, cr, nil)
```
joelddiaz, line 295: assert.NotNil()

pkg/kubevirt/actuator_test.go
```diff
@@ -0,0 +1,400 @@
...
+			},
+		},
+		{
+			name:               "Update CR happy flow - exists",
+			existing:           existingObjectsAfterCreate(t),
+			credentialsRequest: testCredentialsRequest(t),
+			validate: func(t *testing.T, c kubernetesClient.Client) {
+				cr := getCredRequest(t, c)
+				assert.NotEqual(t, cr, nil)
```
joelddiaz, line 304: assert.NotNil()

pkg/kubevirt/actuator.go
bardielle: Im building it as part of "sync" function that calling to "syncCredentialSecret" function that calling to that function "createNewSecret".

If returned an error from "syncCredentialSecret" function than building an ActuatorError (line 124)

joelddiaz: I see what you're saying.

I suppose what I see is that from syncCredentialSecret() it can return an ActuatorError directly, and updateExistingSecret() also returns an ActuatorError, but only createNewSecret() doesn't return an ActuatorError. So I have to spend time thinking about why this function is special.

I'm okay with all returns from syncCredentialSecret() being regular `error` or all returns being ActuatorError. Lets just be consistent.

bardielle: You're right, better being consistently.
I fixed it, thanks

joelddiaz:
@dgoodwin you want to give this PR a look? I'm satisfied with how it is, but since it is adding a new platform type, this isn't the typical PR content.

pkg/kubevirt/actuator.go
```diff
@@ -0,0 +1,269 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package kubevirt
+
+import (
+	"context"
+	"fmt"
+	"github.com/openshift/cloud-credential-operator/pkg/operator/constants"
+	actuatoriface "github.com/openshift/cloud-credential-operator/pkg/operator/credentialsrequest/actuator"
```
dgoodwin, line 22: Imports should be grouped a little better, looks like standard library, random things like logrus, k8s, and openshift.

pkg/kubevirt/actuator.go
```diff
@@ -0,0 +1,269 @@
...
+func (a *KubevirtActuator) Delete(ctx context.Context, cr *minterv1.CredentialsRequest) error {
+	logger := a.getLogger(cr)
+	logger.Debug("running Delete")
+
+	existingSecret, err := a.getSecret(ctx, cr, logger)
+	if err != nil {
+		return err
+	}
+	if existingSecret != nil {
+		logger.Debug("Deleting existing secret")
```
dgoodwin, line 90: logger.Info for this one please.

README.md
```diff
@@ -17,7 +17,7 @@ instance roles)
 
 # Cloud Providers
 
-Currently the operator supports AWS, GCP, Azure, VMWare, OpenStack and oVirt.
+Currently the operator supports AWS, Azure, GCP, kubeVirt, OpenStack. oVirt and VMWare.
```
dgoodwin, line 20: The constant in the Infrastructure API looks like "KubeVirt" which would look a little more natural to me here, could we sync up on that casing across the docs/PR for consistency?

dgoodwin (commented):
Just a couple extremely minor requests, this looks good to me. Very interesting enhancement.

README.md
bardielle: I fixed all the cases in documentation

dgoodwin:
/lgtm

I think merging is currently blocked due to problems we're trying to fix in #263

PR #261, opened 2020-10-21
add current profile annotations to CVO manifests

guillaumerose:
This is matches openshift/enhancements#414 and doesn't change existing behavior


dgoodwin:
The verify test is failing on:

```
 diff vendor/github.com/openshift/api/operator/v1/0000_40_cloud-credential-operator_00_config.crd.yaml ./manifests/00-config-custresdef.yaml
4a5,6
>   annotations:
>     include.release.openshift.io/self-managed-high-availability: "true"
make: *** [verify-vendored-crds] Error 1 
```

This particular CRD definition is coming from openshift/api, and because our manifest does not match what is vendored it's complaining. It looks like you need to add the annotation to the upstream CRD in openshift/api. Unfortunately that may not be so easy as I believe it's autogenerated from go types in that repo. Frankly I think you may next hit that problem in this repo as well for the CredentialsRequest CRD definition.

It might be best to fix both by splicing the annotation in in the Makefile in update-vendored-crds, and/or update/update-codegen, I'm not totally sure when the CredReq crd file is generated in this inherited makefile structure.

guillaumerose:
I updated openshift/api vendoring and run `make update`. 
It forced me to bump k8s from 1.19 to 1.19.2. 

Let see if the CI is happy now.

guillaumerose:
Trying to stop the CI for the moment.

/hold

akhil-rane:
@guillaumerose CI issue is fixed #263  

/hold cancel

guillaumerose:
PR rebased

akhil-rane:
@guillaumerose This fixed the CI issue but was causing a lot of failures for release builds. So had to put it on hold for now. We are investigating.

dgoodwin:
We've seen this from time to time where a PR just can't pass CI, but if you close the PR, open an entirely new PR with the same change, it will pass almost immediately. You may be in this category. We've just had to do this with some other PRs and it worked. Suspect something persistent in the CI assigned to a PR can get out of sync or broken.

guillaumerose:
ok I will do this! thanks

guillaumerose:
Last chance for the CI, after that, I close it :)

PR #263, opened 2020-10-28
Fix /etc/pki/ca-trust/extracted/pem permissions issue

akhil-rane:
Currently, CCO e2es are failing because CI builds volume mount
'/etc/pki/ca-trust' at runtime. Previously we used Dockerfile RUN
command to make '/etc/pki/ca-trust/extracted/pem' world-writable.
Because this command is now being run with a volume mounted at
/etc/pki/ca-trust, the directories and files whose permissions are
modified are those in the volume and are not included in the final
image.

To fix this we add an archive of the contents (writable empty files)
of '/etc/pki/ca-trust/extracted' to the build context using Dockerfile 
ADD and then run 'update-ca-trust extract' as a container
command to populate these files

jira: https://issues.redhat.com/browse/CO-1272


akhil-rane:
Testing this fix

dgoodwin:
Let me know if I'm wrong but I wouldn't expect this to work, if something is now mounting to this location at runtime I don't think changing the perms during a build will help. I expect the fix needs to be a new way to get the cert in there, and in the command in the Deployment yaml where we do the copy today. It looked like there were some suggestions for a command to run to import certs now.

akhil-rane:
> Let me know if I'm wrong but I wouldn't expect this to work, if something is now mounting to this location at runtime I don't think changing the perms during a build will help. I expect the fix needs to be a new way to get the cert in there, and in the command in the Deployment yaml where we do the copy today. It looked like there were some suggestions for a command to run to import certs now.

Agreed. I just wanted to test what happens in CI for this case by fetching the image.

akhil-rane:
@dgoodwin what do you think of this approach? instead of /etc/pki/ca-trust/extracted we make /etc/pki/ca-trust/extracted/pem as our mounted volume and just copy trust bundle in there. Do not need to run `update-ca-trust extract` here. I am having hard time running  `update-ca-trust extract`  inside a container. The container just goes into error state with nothing in the logs or the status. I might be doing something wrong there, trying to figure it out.

joelddiaz (commented):
Can you update the commit text with an explanation of what this `writeable-extracted.tar.gz` file is and why/how it fixes things?
Othewise :+1:

dgoodwin:
Same here, just want a clear description of what that is and how it works, and it should probably be in the commit message and comments around the Dockerfile line.

Otherwise looks great indeed and nice work.

dgoodwin:
I believe they're going to revert the change that caused this, however it will probably take time to make it to the build/ci clusters, so I think we probably should still merge this?

akhil-rane:
> I believe they're going to revert the change that caused this, however it will probably take time to make it to the build/ci clusters, so I think we probably should still merge this?

Yes, I think we should merge. I will update the PR with the commit message.

akhil-rane:
@dgoodwin I have updated the commit message to the best of my knowledge. Let me know if it does not make sense. I might be wrong about some of the things here.

I believe changes to previous releases won't clear CI if we don't backport. So do we want to backport or wait for the CI fix?

dgoodwin:
/lgtm

Need to think on backporting next week, I have fixes blocked for 4.5 in the queue, we may have to clone a bug for 4.6 and 4.5 both, and backport to each.

PR #264, opened 2020-11-02
add new credentials field for AWS Secrets

joelddiaz:
Start storing a usable AWS credentials config file in the 'credentials' field of the Secret. This should allow a consumer of the credentials to just point to the config stored in that field when setting up an AWS client.

Also make sure we are re-queuing CredentialsRequest objects every 1hr10min (so that we are at least periodically doing a full reconcile to restore any lost credentials).

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -54,6 +54,10 @@ const (
 	roAWSCredsSecret          = "cloud-credential-operator-iam-ro-creds"
 	openshiftClusterIDKey     = "openshiftClusterID"
 	clusterVersionObjectName  = "version"
+
+	secretDataAccessKey      = "aws_access_key_id"
+	secretDataSecretKey      = "aws_secret_access_key"
+	secretDataCredentialsKey = "credentials"
 )
 
 var _ actuatoriface.Actuator = (*AWSActuator)(nil)
@@ -153,11 +157,18 @@ func (a *AWSActuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsR
 	}
 
 	// Various checks for the kinds of reasons that would trigger a needed update
-	_, accessKey, secretKey := a.loadExistingSecret(cr)
+	_, accessKey, secretKey, credentialsKey := a.loadExistingSecret(cr)
 	awsClient, err := a.AWSClientBuilder([]byte(accessKey), []byte(secretKey), a.Client)
 	if err != nil {
 		return true, err
 	}
+
+	// Make sure we update old Secrets that don't have the new "credentials" field
+	if credentialsKey == "" || credentialsKey != string(generateAWSCredentialsConfig(accessKey, secretKey)) {
+		logger.Infof("Secret %s key needs updating, will update Secret contents", secretDataCredentialsKey)
+		return true, nil
+	}
+
 	awsSpec, err := DecodeProviderSpec(a.Codec, cr)
 	if err != nil {
 		return true, err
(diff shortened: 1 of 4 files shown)


pkg/aws/actuator/actuator.go
```diff
@@ -54,6 +54,10 @@ const (
 	roAWSCredsSecret          = "cloud-credential-operator-iam-ro-creds"
 	openshiftClusterIDKey     = "openshiftClusterID"
 	clusterVersionObjectName  = "version"
+
+	secretDataAccessKey      = "aws_access_key_id"
+	secretDataSecretKey      = "aws_secret_access_key"
+	secretDataCredentialsKey = "credentials"
```
dgoodwin, line 60: There were some comments on one of our calls that you will need a second config file to point to this creds file, or possibly the section within? Did we determine if that was necessary or not?

pkg/aws/actuator/actuator_test.go
```diff
@@ -372,6 +377,88 @@ func TestUpgradeable(t *testing.T) {
...
+			}
+
+			cr := testCredentialsRequest()
+			logger := a.getLogger(cr)
+			err := a.syncAccessKeySecret(cr, test.accessKeyID, test.secretAccessKey, test.existingSecret, "exampleAWSPolicy", logger)
+
+			require.NoError(t, err, "unexpected error creating/updating Secret")
+
+			secret := &corev1.Secret{}
+			secretKey := types.NamespacedName{Name: cr.Spec.SecretRef.Name, Namespace: cr.Spec.SecretRef.Namespace}
```
dgoodwin, line 448: Suggest secretNSN or similar, key might be a little misleading here. I'm thinking Data[key] when I see that var.

pkg/aws/actuator/actuator_test.go
```diff
@@ -372,6 +377,88 @@ func TestUpgradeable(t *testing.T) {
...
+
+			err = fakeClient.Get(context.TODO(), secretKey, secret)
+			require.NoError(t, err, "unexpected error retriving Secret")
+
+			require.Contains(t, secret.Data, "credentials")
```
dgoodwin, line 453: Can we be pedantic and also check the old keys are there as expected as well.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -71,6 +71,8 @@ const (
 	credentialsRequestInfraMismatch = "InfrastructureMismatch"
 )
 
+var defaultRequeueTime = time.Hour + time.Minute*10
```
dgoodwin, line 74: An hour is conceptually easier to explain to people, is the 10 minute buying us anything? I could see a random jitter like we do in hive but really not a big concern for the scale CCO works at.

dgoodwin (commented):
Looking great just a few nits and questions.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
dgoodwin: Also please add a comment on why we're doing a requeue instead of just waiting for events.

pkg/aws/actuator/actuator.go
joelddiaz: I tried a couple of images that I and @cuppett made covering Python, Golang, PHP, and Java. I've seen all four work with the same credentials configuration file (although not every language uses the same environment variables or setup).

golang:  `AWS_SDK_LOAD_CONFIG=true AWS_SHARED_CREDENTIALS_FILE=<path to creds/config>`
python: `AWS_SHARED_CREDENTIALS_FILE=<path to creds/config>`
java and php couldn't be pointed to a non-default location so i had to: `mkdir -p ${HOME}/.aws && cp <path to credentials/config> ${HOME}/.aws/`

pkg/operator/credentialsrequest/credentialsrequest_controller.go
joelddiaz: Since we explicitly don't re-reconcile more than once every hour, I was just thinking if there is a small difference between when the object is requeued and the lastSync values (where it requeues just before 1 hour so the lastSync is still less than 1 hour ago), then we could effectively slip out to reconciling to every 2 hours. Not a huge deal, but it is double the intended reconcile rate.

dgoodwin:
Good to squash.

sjenning:
seems e2e-aws is failing because CCO clusteroperator is not reporting it's version
```
                "versions": [
                    {
                        "name": "operator",
                        "version": ""
                    }
                ]
```
even though the operator is Progressing=False and Available=True

dgoodwin:
I think all e2e is broken because of https://bugzilla.redhat.com/show_bug.cgi?id=1891759. @akhil-rane is working on it over in #263, and I think they're actually going to be reverting the OpenShift change entirely, though that will take a little longer to make it to the build clusters.

joelddiaz:
/retitle Bug 1896918: add new credentials field for AWS Secrets

PR #266, opened 2020-11-10
aws: support custom CA bundle for AWS API

staebler:
When a cluster is installed in a AWS C2S region, access to the AWS API requires a custom CA bundle for trust. The custom CA bundle is read from the "ca-bundle.pem" key of the kube-cloud-config ConfigMap in the openshift-config-managed namespace.

https://issues.redhat.com/browse/CORS-1584

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -17,8 +17,9 @@ limitations under the License.
 package aws
 
 import (
+	"strings"
+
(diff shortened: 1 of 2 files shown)


joelddiaz (commented):
looks good. just needs some attention to CI

joelddiaz (commented):
/lgtm

PR #270, opened 2020-11-17
Bug 1880785: Fix CredentialsRequest missing description in 'oc explain'

akhil-rane:
In order to make 'oc explain' work we need to enable field pruning
by setting preserveUnknownFields to false

jira: https://issues.redhat.com/browse/OCPBUGSM-17971


joelddiaz (commented):
/lgtm

PR #276, opened 2020-12-08
Bug 1906880: operator: remove odicdiscoveryendpoint controller

sjenning:
implements https://issues.redhat.com/browse/CO-1317

@marun @dgoodwin @joelddiaz


sjenning:
fyi @deads2k

joelddiaz:
@sjenning Do we care that future versions of OpenShift will handle changing the issuer field gracefully? (or at least I remember @marun mentioning something like this)

sjenning:
@joelddiaz I guess the two options are 1) this PR or 2) we disable the controller for 4.7 and re-enable for 4.8

@marun did say that 4.8 should be able to handle this gracefully.  However, my understanding is that it would require a complete restart of the cluster. Thus 4.8 cluster install time would regress if we didn't set the `serivceAccountIssuer` in the install manifests.

marun:
> @joelddiaz I guess the two options are 1) this PR or 2) we disable the controller for 4.7 and re-enable for 4.8
> 
> @marun did say that 4.8 should be able to handle this gracefully. However, my understanding is that it would require a complete restart of the cluster. Thus 4.8 cluster install time would regress if we didn't set the `serivceAccountIssuer` in the install manifests.

My understanding is that for 4.8 the goal is to remove the requirement to require a restart on issuer change. The apiserver would be updated to support multiple issuers to ensure that previously issued tokens would continue to be accepted until expiry. That would allow an issuer change without risk of breakage.

cc: @stlaz

joelddiaz:
@dgoodwin i'm fine with the PR (even after 4.8 gains the ability to gracefully handle the issuer change). With STS we will already require that these pieces be set up pre-installation anyway. WDYT?

dgoodwin:
I largely have to defer to you folks, but if you and Seth are on board then no objections from me, I'd say ship it.

joelddiaz:
> I largely have to defer to you folks, but if you and Seth are on board then no objections from me, I'd say ship it.

I think the TL;DR for what this means is that using the AWS pod identity stuff will require some extra steps going forward as CCO will not be performing this one step.

derekwaynecarr:
as discussed:

upgrade from 4.6 -> 4.7 is not impacted
fresh install on 4.7 will not have this component (no real option here)
release notes need to ensure change is clear
doc should provide steps on how to configure sts as part of install procedure

/lgtm

jeana-redhat:
so this will be a known issue in the 4.7 RNs? just keep me posted :+1: 

> doc should provide steps on how to configure sts as part of install procedure

@derekwaynecarr do you mean internal docs or user docs?

marun:
> as discussed:
> 
> upgrade from 4.6 -> 4.7 is not impacted
> fresh install on 4.7 will not have this component (no real option here)
> release notes need to ensure change is clear
> doc should provide steps on how to configure sts as part of install procedure

What about upgrade from 4.5 to 4.6? Or is this not enabled by default for 4.6?

sjenning:
>  What about upgrade from 4.5 to 4.6? Or is this not enabled by default for 4.6?
It is enabled by default in 4.6.

I think we should backport this to 4.6 z-stream asap.  That way new 4.6 clusters and clusters upgrading from 4.5 to latest 4.6.z have the same behavior as 4.7 clusters.

It will only be clusters that have already installed/upgraded to a 4.6.0-4.6.9ish that will have `serviceAccountIssuer` set by this controller and there really isn't a way to undo it.

joelddiaz:
> I think we should backport this to 4.6 z-stream asap. That way new 4.6 clusters and clusters upgrading from 4.5 to latest 4.6.z have the same behavior as 4.7 clusters.
> 


Can we put in 4.6 an update to this controller to undo the changes back to the internal issuer URL? (so that all clusters would have the internal issuer?)

sjenning:
@joelddiaz we can't do that without causing the issue we are avoiding here; invalidating all the TokenRequest API acquired tokens in the the cluster.

joelddiaz:
I now understand that it's too late for this, and that we're only finding out now because the first bound token users have appeared.

PR #280, opened 2021-01-11
Bug 1916454: checking 4.7 creds

joelddiaz:
Refactor AWS upgradeable checking so that other platforms can use the logic.

Add new GCP, OpenStack, and vSphere secrets to check/set upgradeable on.


akhil-rane (commented):
/lgtm

PR #283, opened 2021-01-13
Add support for Equinix Metal

detiber:
Start adding support for at least documenting the secret format for Equinix Metal.

Related to: https://github.com/openshift/machine-api-operator/pull/753#discussion_r556083729


README.md
```diff
@@ -59,6 +59,18 @@ data:
...
+### Equinix Metal
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: kube-system
+  name: equinix-metal-credentials
+data:
+  PACKET_API_KEY: EquinixMetalAPIKey
```
detiber, line 71: Seeing this, we should likely change the name of this variable to `EQUINIX_METAL_API_KEY` or similar. That would require changes to the cluster-api-provider-equinix-metal PR: https://github.com/openshift/cluster-api-provider-equinix-metal/pull/1#issuecomment-759025769 and installer PR: https://github.com/openshift/installer/pull/4472

README.md
```diff
@@ -212,6 +224,7 @@ Cloud | Mint | Mint + Remove Admin Cred | Passthrough | Manual | Token
 --- | --- | --- | --- | --- | ---
 AWS | Y | 4.4+ | Y | 4.3+ | 4.6+ (expected)
 Azure | Y | N | Y | unknown | N
+EquinixMetal | ? | ? | ? | ?
```
detiber, line 227: @elmiko does it make sense to add more functional support beside just documenting the format? If so, it would help to get some additional context into how CCO fits into the process and what methods might make sense for Equinix Metal to support.

README.md
elmiko: +1, i think that makes sense for clarity

README.md
elmiko: that's a good question, what were you thinking?

detiber: EM doesn't currently have fine grained credentials, so I suspect Passthrough may make the most sense, and then as additional support is added for finer grained credentials, we can likely add support for the other methods.

elmiko: passthrough seems to make sense (i don't a ton about EM though), but as long as fits those restrictions (eg can't mint new credentials) then i would agree.

i think it makes sense to add an asterisk with a note about EM. but i don't want to create an extra maintenance burden, my main goal was to make sure we have a document about what the credentials need in the resource.

> get some additional context into how CCO fits into the process 

just to make sure i don't skip on this part. as i understand it, the CCO holds the credentials that the Machine-API bits will use. when calls the cloud provider API are made then this object will get referenced to ensure we have the proper creds.

> what methods might make sense for Equinix Metal to support

another good question, i would like to confer with the installer team to see if there is anything specific we should be highlighting here.

in general, i think proceeding with the passthrough option makes sense to me but i will gather some more information from our teams.

README.md
```diff
@@ -212,6 +224,7 @@ Cloud | Mint | Mint + Remove Admin Cred | Passthrough | Manual | Token
 --- | --- | --- | --- | --- | ---
 AWS | Y | 4.4+ | Y | 4.3+ | 4.6+ (expected)
 Azure | Y | N | Y | unknown | N
+EquinixMetal | N | N | 4.x+ (expected) | N | N
```
detiber, line 227: I wasn't sure the right way to do this, whether it was to just add a 'Y' here or try to add the version like some of the fields for AWS and GCP...

go.mod
```diff
@@ -41,9 +40,11 @@ require (
...
-	k8s.io/client-go v0.19.2
-	k8s.io/code-generator v0.19.2
+	k8s.io/api v0.20.0
+	k8s.io/apimachinery v0.20.0
+	k8s.io/client-go v0.20.0
+	k8s.io/code-generator v0.20.0
 	sigs.k8s.io/controller-runtime v0.6.2
 )
+
+replace github.com/openshift/api => github.com/detiber/api v0.0.0-20210113181726-19e9a251beff
```
detiber, line 50: Temporary workaround until openshift/api#784 is merged

go.mod
```diff
@@ -41,9 +40,11 @@ require (
...
 	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013
 	google.golang.org/grpc v1.28.0
-	k8s.io/api v0.19.2
-	k8s.io/apimachinery v0.19.2
-	k8s.io/client-go v0.19.2
-	k8s.io/code-generator v0.19.2
+	k8s.io/api v0.20.0
+	k8s.io/apimachinery v0.20.0
+	k8s.io/client-go v0.20.0
+	k8s.io/code-generator v0.20.0
```
detiber, line 46: This was required when I added the replace for `github.com/openshift/api` below.

I believe the other updates came in transitively through the replace and these updates

bindata/bootstrap/cloudcredential_v1_operator_config_custresdef.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: cloudcredentials.operator.openshift.io
   annotations:
     include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
```
detiber, line 7: Updating this through `make update-vendored-crds` was required to pass `make verify`

manifests/00-config-custresdef.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: cloudcredentials.operator.openshift.io
   annotations:
     include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
```
detiber, line 7: Updating this through `make update-vendored-crds` was required to pass `make verify`

pkg/equinixmetal/actuator/actuator.go
```diff
@@ -0,0 +1,364 @@
+/*
```
detiber, line 1: Mostly copy/paste from `pkg/vsphere/actuator/actuator.go`

pkg/operator/secretannotator/equinixmetal/reconciler.go
```diff
@@ -0,0 +1,183 @@
+/*
```
detiber, line 1: Mostly copy/paste from `pkg/operator/secretannotator/vsphere/reconciler.go`

pkg/operator/secretannotator/equinixmetal/reconciler_test.go
```diff
@@ -0,0 +1,198 @@
+/*
```
detiber, line 1: Mostly copy/paste from `pkg/operator/secretannotator/vsphere/reconciler_test.go`

detiber:
@elmiko Updated to add support for passthrough creds using the vsphere implementation as a guide

elmiko:
thanks @detiber, i will be following up with some folks internally next week to make sure this is correct.

go.mod
elmiko: yeah, seems like you got caught in the middle of some changes

manifests/00-config-custresdef.yaml
elmiko: here's a little background on this annotation, https://github.com/openshift/enhancements/blob/master/enhancements/single-node-developer-cluster-profile.md

pkg/equinixmetal/actuator/actuator.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+	unknown := runtime.Unknown{}
+	err = codec.DecodeProviderSpec(providerSpec, &unknown)
+	if err != nil {
+		return false, err
+	}
+	isEquinixMetal := unknown.Kind == reflect.TypeOf(minterv1.EquinixMetalProviderSpec{}).Name()
+	if !isEquinixMetal {
+		log.WithField("kind", unknown.Kind).
+			Info("actuator handles only vsphere credentials")
```
displague, line 354: vsphere/equinixmetal

pkg/equinixmetal/actuator/actuator.go
```diff
@@ -0,0 +1,364 @@
...
+		msg := "unable to fetch root cloud cred secret"
+		logger.WithError(err).Error(msg)
+		return nil, &actuatoriface.ActuatorError{
+			ErrReason: minterv1.CredentialsProvisionFailure,
+			Message:   fmt.Sprintf("%v: %v", msg, err),
+		}
+	}
+
+	if !isSecretAnnotated(cloudCredSecret) {
+		logger.WithField("secret", fmt.Sprintf("%s/%s", constants.CloudCredSecretNamespace, constants.VSphereCloudCredSecretName)).Error("cloud cred secret not yet annotated")
```
displague, line 319: Vsphere/EquinixMetal

pkg/operator/secretannotator/equinixmetal/reconciler_test.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func testSecret() *corev1.Secret {
+	s := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testSecretName,
+			Namespace: testNamespace,
+		},
+		Data: map[string][]byte{
+			"username": []byte("someuser"),
```
displague, line 144: This seems decendent from:
 https://github.com/openshift/cloud-credential-operator/pull/283/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R137-R138 
while Equinix Metal uses:
https://github.com/openshift/cloud-credential-operator/pull/283/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R71
`EQUINIX_METAL_API_KEY`

pkg/operator/secretannotator/secretannotator_controller.go
```diff
@@ -49,6 +50,8 @@ func Add(mgr manager.Manager, kubeconfig string) error {
 		return gcp.Add(mgr, gcp.NewReconciler(mgr, infraStatus.PlatformStatus.GCP.ProjectID))
 	case configv1.VSpherePlatformType:
 		return vsphere.Add(mgr, vsphere.NewReconciler(mgr))
+	case configv1.EquinixMetalPlatformType:
+		return vsphere.Add(mgr, equinixmetal.NewReconciler(mgr))
```
displague, line 54: equinixmetal

README.md
elmiko: @detiber i followed up with some folks internally and passthrough is fine for now. in the future it would be nice if equinix supported the more granular options, but as it sounds like that does not currently exist we should be fine passthrough.

detiber:
Rebased on top of the latest changes, still caught in the middle of a 1.19 to 1.20 rebase.

detiber:
Rebased on top of the latest changes, still caught in the middle of a 1.19 to 1.20 rebase.

README.md
joelddiaz: Similar to my overall comment on this PR, we want to not add any in-cluster processing of CredentialsRequest CRs which means that the only column to be supported would be the Manual column.
This doesn't mean that you don't implement what is effectively Passthrough mode (where each Secret for each in-cluster component contains the same API key), but Passthrough mode implies in-cluster handling of CredentialsRequests...
```suggestion
EquinixMetal | N | N | N | 4.x+ (expected) | N
```

pkg/equinixmetal/actuator/actuator.go
joelddiaz: This is one of those pieces that we don't want to add to the in-cluster operator. As an example of what we would like, you can view the PRs that were done for IBMCloud (the first pass they made at this effectively implemented Passthrough mode, but their current - as of today - PR removes this Passthrough with support for fine-grained creds) https://github.com/openshift/cloud-credential-operator/pull/356

pkg/operator/secretannotator/equinixmetal/reconciler.go
joelddiaz: Same thing here. We don't need in-cluster processing of the Secret that idealy should not even exist.

joelddiaz (commented):
In general for new platforms, we want to move away from having CCO's in-cluster operator handling/processing CredentialsRequest CRs. We want to have cloud credentials provisioning/handling to be handled outside the cluster. This goes back to many customer discussions where they have concerns about a highly-privileged credential sitting inside the cluster.

As an alternative, there is a relatively new `ccoctl` binary in this repo that is meant to handle creating the cloud/platform resources (not applicable for Equinix b/c of supporting shared/passthrough credentials) and the manifests that need to be passed to the installer.

Also, would it be possible to break the vendoring into its own commit?

PR #286, opened 2021-01-14
require annotation before upgrade when in manual mode

joelddiaz:
Today CCO can mark itself Upgradeable=False if it sees that a Secret (containing credentials) that is required for the next minor upgrade is missing.

If an existing credentials in version 4.X simply needs new permissions in 4.Y, CCO in Manual mode cannot probe to determine this. (CCO being in Manual mode is basically CCO making minimal cloud API calls).
Introduce a system so that any/all platforms will set Upgradeable=False until an annotation has been appllied to CCO's CloudCredential config object.

This will set Upgradeable=False when in Manual mode until the user sets the annotation appropriately.

Additionally, this new annotation can be used to bypass situations that would otherwise have the cluster in Upgradeable=False (advanced use only).

Also: remove checks based on known upcoming secrets

moving forward, if CCO is in Manual mode, a user must set an appropriate annotation 'cloudcredential.openshift.io/upgradeable-to' to a higher version (semver-wise) than the most recently completed version installation/upgrade.

for mint/passthrough mode we will continue to only require the root cred secret to exist.

xref: https://issues.redhat.com/browse/CCO-15

### DIFF
--- a/pkg/openstack/actuator.go
+++ b/pkg/openstack/actuator.go
@@ -288,9 +288,5 @@ func (a *OpenStackActuator) getLogger(cr *minterv1.CredentialsRequest) log.Field
 // if the system is considered not upgradeable. Otherwise, return nil as the default
 // value is for things to be upgradeable.
 func (a *OpenStackActuator) Upgradeable(mode operatorv1.CloudCredentialsMode) *configv1.ClusterOperatorStatusCondition {
-	return utils.UpgradeableCheck(a.Client, mode, "4.7", a.getUpcomingCredSecrets(), a.GetCredentialsRootSecretLocation())
-}
-
-func (a *OpenStackActuator) getUpcomingCredSecrets() []types.NamespacedName {
-	return constants.OpenStackUpcomingSecrets
+	return utils.UpgradeableCheck(a.Client, mode, a.GetCredentialsRootSecretLocation())
 }
--- a/pkg/operator/constants/constants.go
+++ b/pkg/operator/constants/constants.go
@@ -47,9 +47,9 @@ const (
 	// the operator config CR specifies an invalide mode
 	StatusModeInvalid = "ModeInvalid"
 
-	// MissingSecretsForUpgradeReason is used when a manual mode cluster is not upgradable due to missing
-	// secrets the operator knows will be required for the next minor release of OpenShift.
-	MissingSecretsForUpgradeReason = "ManualModeMissingSecrets"
+	// MissingUpgradeableAnnotationReason is used when the cluster is not upgradeable due to
+	// the CCO's config object missing the appropriate annotation.
+	MissingUpgradeableAnnotationReason = "MissingUpgradeableAnnotation"
 
 	// ErrorDeterminingUpgradeableReason is used when we encounter unexpected errors checking if a cluster can
 	// be updated.
@@ -126,6 +126,10 @@ const (
 	// KubevirtCloudCredSecretName is the name of the secret where credentials
 	// for Kubevirt are stored.
 	KubevirtCloudCredSecretName = "kubevirt-credentials"
+
+	// UpgradeableAnnotation is the annotation CCO will check for on the cloudcredential.operator.openshift.io
+	// CR when determining upgradeability.
+	UpgradeableAnnotation = "cloudcredential.openshift.io/upgradeable-to"
 )
 
 var (
@@ -138,40 +142,6 @@ var (
 		ModeDegraded,
 		ModeUnknown,
 		ModeManualPodIdentity,
-	}
-
-	// Add known new credentials for next version upgrade
-
-	// AWSUpcomingSecrets contains the list of known new AWS credential secrets for the next version of OpenShift
-	AWSUpcomingSecrets = []types.NamespacedName{}
-	// AzureUpcomingSecrets contains the list of known new Azure credential secrets for the next version of OpenShift
-	AzureUpcomingSecrets = []types.NamespacedName{}
-	// GCPUpcomingSecrets contains the list of known new GCP credential secrets for the next version of OpenShift
-	GCPUpcomingSecrets = []types.NamespacedName{
-		{
-			Namespace: "openshift-cluster-csi-drivers",
-			Name:      "gcp-pd-cloud-credentials",
-		},
-		{
-			Namespace: "openshift-cloud-credential-operator",
-			Name:      "cloud-credential-operator-gcp-ro-creds",
-		},
-	}
-	// OpenStackUpcomingSecrets contains the list of known new OpenStack credential secrets for the next version of OpenShift
-	OpenStackUpcomingSecrets = []types.NamespacedName{
-		{
-			Namespace: "openshift-cluster-csi-drivers",
-			Name:      "openstack-cloud-credentials",
-		},
-	}
-	// OvirtUpcomingSecrets contains the list of known new oVirt credential secrets for the next version of OpenShift
-	OvirtUpcomingSecrets = []types.NamespacedName{}
-	// VsphereUpcomingSecrets contains the list of known new vSphere credential secrets for the next version of OpenShift
-	VsphereUpcomingSecrets = []types.NamespacedName{
-		{
-			Namespace: "openshift-cluster-storage-operator",
-			Name:      "openshift-vsphere-problem-detector",
-		},
 	}
 
 	// Add known stale credentials requests here
--- a/pkg/operator/utils/utils.go
+++ b/pkg/operator/utils/utils.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 
+	"golang.org/x/mod/semver"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metaerrors "k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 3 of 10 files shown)


joelddiaz:
/hold
this depends on #280

pkg/openstack/actuator.go
```diff
@@ -287,5 +288,15 @@ func (a *OpenStackActuator) getLogger(cr *minterv1.CredentialsRequest) log.Field
...
 // value is for things to be upgradeable.
 func (a *OpenStackActuator) Upgradeable(mode operatorv1.CloudCredentialsMode) *configv1.ClusterOperatorStatusCondition {
-	return nil
+	upgradeParams := &utils.UpgradeableParams{
+		Mode:               mode,
+		TargetRelease:      constants.UpgradeableNextRelease,
+		ExpectedNewSecrets: a.getUpcomingCredSecrets(),
+		RootSecretName:     a.GetCredentialsRootSecretLocation(),
+	}
+	return utils.UpgradeableCheck(a.Client, upgradeParams)
```
dgoodwin, line 297: Perhaps the actuators should return a list of problematic versions, and if none are set, the annotation is required. For example on AWS we need to do this, on other clouds we do not presently, so it would be a shame to make them do this work for no reason, especially in the absence of a "manual mode cred reconcile tool". (btw I really suggest keeping that in mind when looking into that STS tooling)

pkg/operator/constants/constants.go
```diff
@@ -47,6 +47,10 @@ const (
 	// secrets the operator knows will be required for the next minor release of OpenShift.
 	MissingSecretsForUpgradeReason = "ManualModeMissingSecrets"
 
+	// MissingUpgradeableAnnotationReason is used the cluster is not upgradeable due to
+	// the CCO's config object missing the appropriate annotation.
```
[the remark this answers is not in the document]
joelddiaz: It is intended for Manual mode for sure. But it was easy enough to move the place where we check for the annotation early in the Upgradeable() path so that non-Manual users could move forward in the event that CCO's Upgradeable() code was giving false negatives.

pkg/operator/constants/constants.go
```diff
@@ -118,6 +122,12 @@ const (
 	// KubevirtCloudCredSecretName is the name of the secret where credentials
 	// for Kubevirt are stored.
 	KubevirtCloudCredSecretName = "kubevirt-credentials"
+
+	// UpgradeableNextRelease is the version CCO will calculate whether it can upgrade to.
+	UpgradeableNextRelease = "4.7"
```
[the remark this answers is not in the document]
joelddiaz: I think we could make it current +1. Need to think about how this would work when we need to merge a commit into 4.7 (which should be benign in 4.7) before it gets into 4.6...

The annotation is only required when the upgradeParams.MissingAnnotationMessage is a non empty string. The string is the message to communicate what is expected from the user before setting the annotation to get CCO to Upgradeable=True.

Yes, for EUS, you would need to resolve any change from one release to the next. Without that, we would have to encode in .5 all the knowledge about getting from .5 to .8 (and a user might only be interested in the info for .5 to .6. This might be doable if .5.x is the last non-EUS update, and .5.y (and everything that follows) is exclusively EUS updates. Then it could make sense to encode all that information to upgrade to .8 in .5.y.

We could absolutely do the semver calculation. The challenge is communicating all of the info that might be needed to get from currentEUS to nextEUS. Perhaps the message is `Go read https://redhat.com/upgrading-4.5-to-4.8.html` ?

pkg/operator/utils/utils.go
```diff
@@ -287,3 +287,127 @@ func IsValidMode(operatorMode operatorv1.CloudCredentialsMode) bool {
...
+
+// UpgradeableParams encapsulates the data needed to determine setting the
+// ClusterOperator Upgradeable condition.
+type UpgradeableParams struct {
+	// Mode is the current CCO mode the cluster is configured to
+	Mode operatorv1.CloudCredentialsMode
+	// TargetRelease is the release being considered for upgrade
+	TargetRelease string
+	// ExpectedNewSecrets is the list of known new secrets in the upcoming release
+	ExpectedNewSecrets []types.NamespacedName
```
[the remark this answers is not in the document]
joelddiaz: For new creds, the presence of the Secret has been the flag indicating that it is okay to upgrade. It's only for those tricky cases where a pre-existing Cred needs new permissions that we need this annotation process.

pkg/openstack/actuator.go
joelddiaz: The sts-preflight tool kinda-sorta does this (or if it doesn't it can easily be extended to do this). You fetch the CredentialsRequest list from whatever version of OpenShift you're interested in, and the sts-preflight tool will update any existing Roles if they need new permissions. It also creates the Secret yaml, but doesn't presently automatically apply the Secrets.

dgoodwin: And does it do this in an additive fashion, i.e. it's ok to make new creds, or add permissions, but do not reconcile removal of any creds or permissions because i'm probably not on that version yet? IMO the tool for this purpose should have a mode/flag to opt-in to "reconcile and allow removing creds/perms" vs additive only preupgrade mode (which is probably the default)

pkg/operator/constants/constants.go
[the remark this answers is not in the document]
dgoodwin: Ooo I like that, ok cool.

pkg/openstack/actuator.go
joelddiaz: sts-preflight has no idea about CredRequest that you do not provide to it. It just converts a list of CredRequest yaml to IAM Roles and kubernetes Secrets (with predictable names). I suppose you could provide it CredRequests with non-nil deletionTimestamp to trigger a deletion, but it wasn't something that has been considered up to this point.
Maybe @akhil-rane 's cleanup work can annotate CredRequest that this tool can see and take action with.

pkg/operator/constants/constants.go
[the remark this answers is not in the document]
joelddiaz: See if this makes sense. If we make it current+.1

Cluster just upgraded to 4.8 by setting the annotation. To be more concrete, the user was instructed to add the s3:ListMultiPartUploadParts permission, and then the user set the annotation to unblock the upgrade to 4.8.

Upgradeable() in 4.8 (which 4.7 simply backported the check from (assuming we didn't know so far in advance that this new permission was needed until 4.8 was near release)) now sees that annotation indicating 4.8 (which of course is < 4.8 + .1). Now after a successful upgrade to 4.8, CCO marks things Upgradeable=False indicating the same steps (that were supposed to be for 4.8) are needed before upgrading to 4.9.

This scenario depends on the assumption that code needs to land in 4.8 before we can backport the 4.8 check into 4.7. If we're allowed to directly commit future-looking checks into 4.7 (bypassing 4.8) then we can avoid this.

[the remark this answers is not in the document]
dgoodwin: With the data defined in the cred request itself, and documented steps for how to upgrade, adding a third place where we convey the specifics is getting to be a lot to keep in sync. I'd be tempted to drop the idea of communicating exactly what needs to be done in the upgradeable message, it's fine for one permission, but it could be 20 across a bunch of cred requests someday. I think your idea of linking to docs that show how to extract the info from a release image *generically* would be ideal. The user should get familiar with that process if they're in manual mode. And of course ideally I'd love to let them have a tool to do it.

WDYT if it just said:  "You have not yet manually prepared credentials/permissions for release x.y. Please see https://redhat.com/manual-credentials-upgrade.html." And there we could show how to run the tool, or do it manually until it exists.

Tool would also have the same problem of not knowing what user identity belongs to a cred request, though we might be able to figure it out, and maybe even populate it.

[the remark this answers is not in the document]
joelddiaz: I do like this. It forces us to make sure the docs are in good shape (if they aren't already). All we have to do to be ready for a release is know whether there were any changes in the upcoming release, and we can turn on the "please annotate things before upgrading" path if something did change.

For STS, the tool (assuming the tool also made the IAM Roles) will know the names of the IAM Roles as they are named using a repeatable scheme (clusterInfra-targetNamespace-targetName https://github.com/sjenning/sts-preflight/blob/master/pkg/iamroles/iamrole.go#L68 ). I don't believe we have any plans to create tooling for the non-STS case so that situation would really be a "make sure you read the docs" situation.

joelddiaz:
/unhold
#280 merged

pkg/operator/constants/constants.go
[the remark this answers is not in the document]
dgoodwin: Should we defer the assumed target release or get it into this PR?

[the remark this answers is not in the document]
joelddiaz: This is the item I keep coming back to. How can we write this in a way that is backportable so that it works in 4.x (ie. set the Upgradeable=False until annotation is seen) and 4.x+1 (ie. don't set Upgradeable=False b/c the fact that we are in 4.x+1 means we must have already made the annotation to get to 4.x+1) w/o hardcoding these values?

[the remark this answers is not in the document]
dgoodwin: I'm thinking:

- current := 4.6 (introspect to find out)
- next := current+1 major (i.e. 4.7)
- if CR has no annotation, upgradable = false
- if CR has annotation that is semver >= next, upgradeable=true (i.e. annotation of 4.7, 4.8, 4.9, would all mark upgradable true if you're on 4.6)
- if CR has annotation that is semver < next, upgradeable=false (i.e. annotation of 4.6 would mark false if you're on 4.6)

Am I missing a case in there?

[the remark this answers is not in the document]
dgoodwin: I see what you mean now. 

We could take the stance, you need to get in the habit of running this tool to reconcile your creds, regardless if the release needs anything or not. This makes the code a lot simpler and gets users into good habits. Being upgradeable false in manual mode immediately though is something I'd consider to be fairly normal.

Or, we could encode info as to which specific releases do or don't have a problem on a platform. The AWS actuator could know a list of releases where changes are needed, i.e. ["4.7, "4.9"]. That's pretty much back to hardcoded which I see now is what you're thinking.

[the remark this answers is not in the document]
joelddiaz: I suppose the decision is whether Upgradeable=False should be the normal state of a cluster in Manual mode. Do we bring this up in our arch meeting?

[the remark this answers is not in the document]
dgoodwin: Yup lets get it onto the agenda for next week.

joelddiaz:
@dgoodwin this should be good for another review. we're now dropping the idea of knowing when there are upcoming new secrets, and just saying that if the cluster is in Manual mode, the annotation needs to be set approopriately.
There is some semver checking of the annotation against the currently running version of the cluster, and no hard coding of what the "next version" is.

pkg/operator/utils/utils.go
```diff
@@ -292,79 +294,123 @@ func IsValidMode(operatorMode operatorv1.CloudCredentialsMode) bool {
...
+	if !strings.HasPrefix(upgradeableTo, "v") {
+		upgradeableTo = "v" + upgradeableTo
+	}
+
+	upgradeableTo = semver.MajorMinor(upgradeableTo)
+
+	if semver.Compare(upgradeableTo, clusterSemVer) == 1 {
+		log.Info("Found annotation indicating upradeability, assuming upgradeable")
+		return nil
+	}
```
dgoodwin, line 355: Feels like all of this should be guarded on manual mode and moved into the block below.

pkg/operator/utils/utils.go
```diff
@@ -292,79 +294,123 @@ func IsValidMode(operatorMode operatorv1.CloudCredentialsMode) bool {
...
-				} else {
-					// If we can't figure out if you're upgradeable, you're not upgradeable:
-					secLog.WithError(err).Error("unexpected error looking up secret, marking upgradeable=false")
-					upgradeableCondition.Status = configv1.ConditionFalse
-					upgradeableCondition.Reason = constants.ErrorDeterminingUpgradeableReason
-					upgradeableCondition.Message = fmt.Sprintf("Error determining if cluster can be upgradead to %s: %s",
-						toRelease, err)
-					return upgradeableCondition
-				}
-			}
```
dgoodwin, line 327: I'm not sure we should drop this framework, it is somewhat redundant with the annotation check, but it's also an extra layer of security. In general I like the idea of the cco being intelligent in guiding you for what's wrong, and I think covering the case where you may have claimed you minted your creds for next release, but we see a secret that is clearly missing, I think that's worth keeping. WDYT?

dgoodwin: Then again this violates my above goal about not having to remember to maintain a release number for the next version each release. Curious what you and @akhil-rane think.

pkg/operator/utils/utils.go
joelddiaz: This occurrs before the Manual mode block so that someone can get a cluster unstuck (upgrade-wise) even for non-Manual cases. We can move it into the Manual mode block, but then we lose the escape hatch for all modes that this currently provides.

pkg/operator/utils/utils.go
joelddiaz: I'm not sure what the right answer is here. I can see an advanced customer who knows what they're doing not needing to worry about creating a Secret for a component that they have disabled. For exmaple, imagine that the new AWS CSI driver/operator was disableable.
It's a tough needle to thread here as Manual mode was originally intended for more advanced use cases, and going forward we might find that Manual mode might not be that advanced with STS stuff leveraging Manual mode.

On a related note, if we can get people used to running `ccoctl` before upgrades, we can use that opportunity to highlight anything that a user might have missed before upgrade. Someone running `ccoctl` would see the tool lay down the new Secret manifest for the AWS CSI component, so this might help someone having missed a pre-upgrade step.
(all this `ccoctl` talk of course means that the tool would grow to handle more than just the STS use case...)

dgoodwin: I definitely think ccoctl being able to do more than STS should be on the table eventually, it'd be great for manual mode in general.

The disabled or unused operators is a probably good enough reason to drop this bit of code for now.

dgoodwin:
Ok I think that's everything I was worried about settled. @akhil-rane can you take a pass and do the honors?

joelddiaz:
Let me know, and I'll squash before.

pkg/operator/utils/utils.go
joelddiaz: Maybe I've had a change of heart...on the other hand, we could leave this missing Secret stuff in. Setting the annotation will bypass any missing Secret stuff which allows advanced users to move an upgrade forward.

dgoodwin: Feels like it may not help, if you must set the annotation, and if setting the annotation overrides the missing secret code, the missing secret code couldn't really ever help. (except before you set the annotation maybe but it's use is pretty thin by that point) I was thinking more as an additional layer of checks after setting the annotation.

pkg/operator/utils/utils.go
akhil-rane: `someone can get a cluster unstuck (upgrade-wise) even for non-Manual cases` I did not understand what this means

pkg/operator/utils/utils.go
akhil-rane: I prefer manual mode to be "user knows what to do". Even though we can provide additional security by maintaining secrets, we do not provide complete security since permissions can always change. I am fine with removing this code.

akhil-rane (commented):
few questions, otherwise looks good to me

pkg/operator/utils/utils.go
joelddiaz: In the event that our upgradeable code path has an issue (no matter what mode CCO is in), setting the annotation will bypass all the Upgradeable checks and just mark CCO as upgradeable no matter what. This is just a failsafe so that we can get a cluster/customer unstuck in the unlikely ;) event that we mess up.

joelddiaz:
I've squashed and this should be ready for merging now.

PR #288, opened 2021-01-20
Bug 1920695: Update OpenShift CI registry in Makefile

akhil-rane:
Images prow job is failing with the old registry, moving to a new one


joelddiaz (commented):
/lgtm

PR #296, opened 2021-02-09
new binary for setting up credentials outside of cluster

joelddiaz:
Introduce 'credentials-provisioning' (we should pick a better name) binary. This will evolve to help with setting up and updating of credentials outside of the cluster.

Presently, it only implements the first useful command of 'create key-pair' to generate the ServiceAccount signing keys that will be fed to the cluster/installer.

xref: https://issues.redhat.com/browse/CCO-66


pkg/provisioning/create_test.go
```diff
@@ -0,0 +1,93 @@
...
+			},
+			verify: func(t *testing.T) {
+				pwd, err := os.Getwd()
+				require.NoError(t, err, "unexpected error getting current directory")
+
+				assert.Contains(t, CreateOpts.TargetDir, pwd)
+			},
+		},
+		{
+			name: "specify directory",
```
akhil-rane, line 41:
```suggestion
			name: "specify target directory",
```

pkg/provisioning/create_test.go
```diff
@@ -0,0 +1,93 @@
...
+			setup: func(t *testing.T) {
+				CreateOpts.TargetDir = tempDirName
+			},
+			verify: func(t *testing.T) {
+
+				assert.Equal(t, CreateOpts.TargetDir, tempDirName)
+			},
+		},
+		{
+			name: "specify directory that does not exist",
```
akhil-rane, line 51:
```suggestion
			name: "specify target directory that does not exist",
```

pkg/provisioning/keypair.go
```diff
@@ -0,0 +1,133 @@
...
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh"
+)
+
+const (
+	privateKeyFile = "serviceaccount-signer.private"
+	publicKeyFile  = "serviceaccount-signer.public"
+)
```
akhil-rane, line 20: Do we also want to give an option to users to name these keys?

pkg/provisioning/keypair.go
```diff
@@ -0,0 +1,133 @@
...
+
+	privateKeyFilePath := filepath.Join(prefixDir, privateKeyFile)
+	publicKeyFilePath := filepath.Join(prefixDir, publicKeyFile)
+	bitSize := 4096
+
+	_, err := os.Stat(privateKeyFilePath)
+	if err == nil {
+		log.Printf("Using existing RSA keypair found at %s", privateKeyFilePath)
+		return
+	}
```
akhil-rane, line 32: Is it okay to assume that if private key exists, public will also exist?

pkg/provisioning/keypair.go
```diff
@@ -0,0 +1,133 @@
...
+		Bytes:   pubKeyBytes,
+	})
+	f.Close()
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+}
+
+// generatePrivateKey creates a RSA Private Key of specified byte size
+func generatePrivateKey(bitSize int) (*rsa.PrivateKey, error) {
```
akhil-rane, line 79: Are we using this function anywhere?

pkg/provisioning/keypair.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, err
+	}
+
+	log.Println("Private Key generated")
+	return privateKey, nil
+}
+
+// generatePublicKey take a rsa.PublicKey and return bytes suitable for writing to .pub file
+// returns in the format "ssh-rsa ..."
+func generatePublicKey(privatekey *rsa.PublicKey) ([]byte, error) {
```
akhil-rane, line 98: Are we using this function anywhere?

pkg/provisioning/keypair.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	pubKeyBytes := ssh.MarshalAuthorizedKey(publicRsaKey)
+
+	log.Println("Public key generated")
+	return pubKeyBytes, nil
+}
+
+// writePemToFile writes keys to a file
+func writeKeyToFile(keyBytes []byte, saveFileTo string) error {
```
akhil-rane, line 111: Are we using this function anywhere?

pkg/provisioning/keypair.go
```diff
@@ -0,0 +1,133 @@
...
+	_, err := os.Stat(privateKeyFilePath)
+	if err == nil {
+		log.Printf("Using existing RSA keypair found at %s", privateKeyFilePath)
+		return
+	}
+
+	log.Print("Generating RSA keypair")
+	privateKey, err := rsa.GenerateKey(rand.Reader, bitSize)
+	if err != nil {
+		log.Fatal(err.Error())
```
akhil-rane, line 37:
```suggestion
log.WithError(err).Fatal("Failed to generate private key")
```

Just following the way we do it in hive.

docs/credentials-provision.md
```diff
@@ -0,0 +1,17 @@
...
+
+By default, the tool will output to the directory the command(s) were run in. To specify a directory, use the `--output-dir` flag.
+
+## Creating RSA keys
+
+To generate keys for use when setting up the cluster's OpenID Connect provider, run
+
+```bash
+$ credentials-provision create key-pair
+```
```
akhil-rane, line 15:
```suggestion
$ credentials-provisioning create key-pair
```

docs/credentials-provision.md
```diff
@@ -0,0 +1,17 @@
...
+## Global flags
+
+By default, the tool will output to the directory the command(s) were run in. To specify a directory, use the `--output-dir` flag.
+
+## Creating RSA keys
+
+To generate keys for use when setting up the cluster's OpenID Connect provider, run
+
+```bash
+$ credentials-provision create key-pair
```
akhil-rane, line 14:
```suggestion
credentials-provisioning create key-pair
```
As written in the code

akhil-rane (commented):
Looks good to me. Few comments. I am not good at naming stuff so may be @dgoodwin can suggest something here. Also, do you think it is possible to include some unit tests for keypair.go?

dgoodwin:
Ideas:
- cco
- ccoctl
- ccoadm

pkg/provisioning/keypair.go
joelddiaz: I think it just makes it more complicated when you run 'tool create cloud-pieces'. The step after creating keys will not be able to assume file names (unless we store some state file that says "i saved the key files as...), so you'll have to pass flags for the "happy path" (where you trust the tool to do everything).

If you want to have custom names, we will cover that will the bring-your-own-keys flag (something like 'tool create cloud-pieces --use-private-key=/path/to/private/key/filename').

WDYT?

pkg/provisioning/keypair.go
joelddiaz: I thought about this a bit. And I'm not sure what the right answer is. We can always regenerate the public key from the private key, so maybe we don't even need to save the public key file at this point. Or the future code that needs the public key can always just generate it from the private key as needed. :shrug:

docs/credentials-provision.md
joelddiaz: hmm..maybe 'credentials-provision' is a slightly better name...

joelddiaz:
>Also, do you think it is possible to include some unit tests for keypair.go?
I started to write test cases, but the code so linear. There are no branching paths, its just a long sequence of:
1) perform action A
2) if failed: fatal
3) perform action B
4) if failed: fatal
...

It is also a nightmare to test the error paths because of the log.Fatal() calls. @dgoodwin WDYT?

pkg/provisioning/keypair.go
akhil-rane: For now lets keep it simple and ignore this option. May be a good idea to put the names in the document so that user knows what to expect.

docs/credentials-provision.md
akhil-rane: +1 for credentials-provision

akhil-rane: > Ideas:
> 
> * cco
> * ccoctl
> * ccoadm

I will vote for ccoctl here

akhil-rane:
> > Also, do you think it is possible to include some unit tests for keypair.go?
> > I started to write test cases, but the code so linear. There are no branching paths, its just a long sequence of:
> 
> 1. perform action A
> 2. if failed: fatal
> 3. perform action B
> 4. if failed: fatal
>    ...
> 
> It is also a nightmare to test the error paths because of the log.Fatal() calls. @dgoodwin WDYT?
> 
> Technically that's probably an indication the code needs some work. We shouldn't log.Fatal outside of the main.go IMO, inside all other modules we should return errors, is that possible with cobra in this case? If so we test at that module level instead of main.go and problem is solved.

Can we make createKeys to return an error and unit test that function? This way we can make keyCmd to check for error and log fatal

cmd/ccoctl/main.go
```diff
@@ -0,0 +1,24 @@
...
+	"log"
+
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/cloud-credential-operator/pkg/provisioning"
+)
+
+func main() {
+	rootCmd := &cobra.Command{
+		Use:   "credentials-provisioning",
```
akhil-rane, line 13: `ccoctl` here?

pkg/provisioning/keypair_test.go
```diff
@@ -0,0 +1,99 @@
...
+		name        string
+		setup       func(*testing.T) string
+		verify      func(t *testing.T, tempDirName string)
+		cleanup     func(*testing.T)
+		expectError bool
+	}{
+		{
+			name: "private key file exists",
+			setup: func(t *testing.T) string {
+				tempDirName, err := ioutil.TempDir(os.TempDir(), "keypairtestdir")
```
akhil-rane, line 27: nit: `keypairtestdir` is used at few places. Can be made a constant

akhil-rane (commented):
Do you think it is better to squash last 3 commits into one?

joelddiaz:
> Do you think it is better to squash last 3 commits into one?

Definitely, just trying to keep the changes from one PR review to the next easy to see. But I'll squash once everything looks good.

akhil-rane:
@joelddiaz looks good to me now. Lets squash and merge it.

PR #297, opened 2021-02-09
Bug 1924477: Tag gogo/protobuf to v1.3.2

suhanime:


akhil-rane (commented):
/lgtm

akhil-rane:
@suhanime you might want to ask in the BZ ticket if this is required to be backported to 4.7 too

suhanime:
> @suhanime you might want to ask in the BZ ticket if this is required to be backported to 4.7 too

Looking at https://bugzilla.redhat.com/show_bug.cgi?id=1924477#c0 it seems 4.8 is fine
@joelddiaz was okay with 4.8 target last I checked

joelddiaz:
Whether we backport should be decided by whether or not we actually use this codepath. If it is just something that is imported and otherwise dead code, then fixing in the master branch is likely enough.

go.mod
```diff
@@ -47,3 +47,5 @@ require (
 	k8s.io/code-generator v0.19.2
 	sigs.k8s.io/controller-runtime v0.6.2
 )
+
+replace github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2 // Required to fix CVE-2021-3121
```
joelddiaz, line 51: Why the need for the manual override? I would have though a process like this would be enough:
```
go get -u github.com/gogo/protobuf
go mod tidy -v
go mod vendor -v
```

suhanime: This does not take care of the packages that are using protobuf - some of them seem to be still using older version because I can see multiple entries in go sum

joelddiaz (commented):
/lgtm

PR #299, opened 2021-02-11
Add a CredentialsRequest field to store service account names

akhil-rane:
The new field will not be used by CCO but is needed for ccoctl tool to
set up access control in the cloud provider.

### DIFF
--- a/pkg/apis/cloudcredential/v1/types_credentialsrequest.go
+++ b/pkg/apis/cloudcredential/v1/types_credentialsrequest.go
@@ -53,6 +53,13 @@ type CredentialsRequestSpec struct {
 	// ProviderSpec contains the cloud provider specific credentials specification.
 	// +kubebuilder:pruning:PreserveUnknownFields
 	ProviderSpec *runtime.RawExtension `json:"providerSpec,omitempty"`
+
+	// ServiceAccountNames contains a list of ServiceAccounts that will use permissions associated with this
+	// CredentialsRequest. This is not used by CCO, but the information is needed for being able to properly
+	// set up access control in the cloud provider when the ServiceAccounts are used as part of the cloud
+	// credentials flow.
+	// +optional
+	ServiceAccountNames []string `json:"serviceAccountNames,omitempty"`
 }
 
 // CredentialsRequestStatus defines the observed state of CredentialsRequest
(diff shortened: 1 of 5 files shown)


pkg/apis/cloudcredential/v1/types_credentialsrequest.go
```diff
@@ -53,6 +53,12 @@ type CredentialsRequestSpec struct {
 	// ProviderSpec contains the cloud provider specific credentials specification.
 	// +kubebuilder:pruning:PreserveUnknownFields
 	ProviderSpec *runtime.RawExtension `json:"providerSpec,omitempty"`
+
+	// ServiceAccountNames contains a list of service accounts that will use permissions granted by this
+	// CredentialsRequest. This is not used by CCO but instead needed for ccoctl tool to set up access
+	// control in the cloud provider.
```
joelddiaz, line 59: My suggestion is kind of wordy, but I'm trying to capture why this information is needed in a generic way...

```suggestion
	// ServiceAccountNames contains a list of ServiceAccounts that will use permissions associated with this
	// CredentialsRequest. This is not used by CCO, but the information is needed for being able to properly
	// set up access control in the cloud provider when the ServiceAccounts are used as part of the cloud
	// credentials flow.
```

pkg/apis/cloudcredential/v1/types_credentialsrequest.go
```diff
@@ -53,6 +53,12 @@ type CredentialsRequestSpec struct {
...
+	// +optional
+	ServiceAccountNames []string `json:"serviceAccountNames,omitempty"`
```
joelddiaz, line 61: Do we want `ServiceAccountNames` or just `ServiceAccounts`? My gut tells me `ServiceAccounts`, but the Pod spec does have fields for both `ServiceAccount` and `ServiceAccountName`.
WDYT?

akhil-rane: I would have gone with `ServiceAccounts` if it was a list of structs. But we are only storing names here (which is a specific attriibute of ServiceAccount) so `ServiceAccountNames` sounds more appropriate to me.

akhil-rane:
PR updated

joelddiaz (commented):
:+1: 
@dgoodwin you want to give this a once-over (since we're affecting API)?

pkg/apis/cloudcredential/v1/types_credentialsrequest.go
dgoodwin: SGTM

dgoodwin (commented):
/lgtm

PR #302, opened 2021-02-21
CCO-67: Add command to setup OpenID Connect provider

akhil-rane:
This PR adds following command to ccoctl to create OpenID connect provider along with the documentation for the same

`ccoctl create identity-provider --infra-name=<infra-name> --region=<aws-region> --public-key=/path/to/public/key/file`


pkg/cmd/provisioning/identity_provider.go
```diff
@@ -0,0 +1,335 @@
...
+		Use: "identity-provider",
+		Run: identityProviderCmd,
+	}
+
+	identityProviderSetupCmd.PersistentFlags().StringVar(&CreateOpts.InfraName, "infra-name", "", "Name prefix for all created AWS resources")
+	identityProviderSetupCmd.MarkPersistentFlagRequired("infra-name")
+	identityProviderSetupCmd.PersistentFlags().StringVar(&CreateOpts.Region, "region", "", "AWS region where the S3 OpenID Connect endpoint will be created")
+	identityProviderSetupCmd.MarkPersistentFlagRequired("region")
+	identityProviderSetupCmd.PersistentFlags().StringVar(&CreateOpts.PublicKeyPath, "public-key", "", "Path to public ServiceAccount signing key")
+	identityProviderSetupCmd.MarkPersistentFlagRequired("public-key")
```
joelddiaz, line 332: I believe we also want some kind of --dry-run or --skip-create flag so that the tool just places the AWS objects on the filesystem (so that a user can review and apply the objects if they don't trust ccoctl or they just want some extra tags, naming convention, etc).

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -0,0 +1,335 @@
...
+    ]
+}`
+)
+
+type JSONWebKeySet struct {
+	Keys []jose.JSONWebKey `json:"keys"`
+}
+
+func createIdentityProvider(client aws.Client, infraName, region, publicKeyPath, targetDir string) error {
+	bucketName := fmt.Sprintf("%s-installer", infraName)
```
joelddiaz, line 67: s/-installer/-oidc/

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -0,0 +1,335 @@
...
+	var bucketTagValue string
+	_, err := client.CreateBucket(&s3.CreateBucketInput{
+		Bucket: awssdk.String(bucketName),
+	})
+	if err != nil {
+		var aerr awserr.Error
+		if errors.As(err, &aerr) {
+			switch aerr.Code() {
+			case s3.ErrCodeBucketAlreadyOwnedByYou:
+				bucketTagValue = sharedCcoctlAWSResourceTagValue
```
joelddiaz, line 79: Help me understand these proposed tag values. If we create the bucket we call it "owned", but if the tagging fails and we we-run the tool, we then get tagged "shared"?
I don't think this is a huge deal, but what are we getting from these tagging values?

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -0,0 +1,335 @@
...
+type JSONWebKeySet struct {
+	Keys []jose.JSONWebKey `json:"keys"`
+}
+
+func createIdentityProvider(client aws.Client, infraName, region, publicKeyPath, targetDir string) error {
+	bucketName := fmt.Sprintf("%s-installer", infraName)
+	issuerURL := fmt.Sprintf("https://%s.s3.%s.amazonaws.com", bucketName, region)
+
+	var bucketTagValue string
+	_, err := client.CreateBucket(&s3.CreateBucketInput{
```
joelddiaz, line 71: For the `--dry-run` param (or whatever we end up calling it), we need to also provide the JSON (or YAML) of the AWS object that needs to be created/updated. Something like this (I believe you can leave out parameters that you want as defaults):
```
[jdiaz@minigoomba ~]$ aws s3api create-bucket --generate-cli-skeleton
{
    "ACL": "authenticated-read",
    "Bucket": "",
    "CreateBucketConfiguration": {
        "LocationConstraint": "ap-east-1"
    },
    "GrantFullControl": "",
    "GrantRead": "",
    "GrantReadACP": "",
    "GrantWrite": "",
    "GrantWriteACP": "",
    "ObjectLockEnabledForBucket": true
}

```

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -0,0 +1,335 @@
...
+		Key:     awssdk.String(keysURI),
+		Tagging: awssdk.String(fmt.Sprintf("%s/%s=%s", ccoctlAWSResourceTagKeyPrefix, infraName, ownedCcoctlAWSResourceTagValue)),
+	})
+
+	if err != nil {
+		return errors.Wrapf(err, "Failed to upload JSON web key set (JWKS) in the S3 bucket %s", bucketName)
+	}
+	log.Printf("JSON web key set (JWKS) in the S3 bucket %s at %s updated", bucketName, keysURI)
+
+	err = saveToFile("JSON web key set (JWKS)", filepath.Join(targetDir, "keys.json"), jwks)
```
joelddiaz, line 143: We can probably only save the file when run with `--dry-run`.

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	if len(providerARN) == 0 {
+		oidcOutput, err := client.CreateOpenIDConnectProvider(&iam.CreateOpenIDConnectProviderInput{
+			ClientIDList: []*string{
+				awssdk.String("openshift"),
+				awssdk.String("sts.amazonaws.com"),
+			},
+			ThumbprintList: []*string{
+				awssdk.String("A9D53002E97E00E043244F3D170D6F4C414104FD"), // root CA thumbprint for Amazon S3 (DigiCert)
```
joelddiaz, line 174: I'd rather fetch this dynamically from the bucket we just created (in the event that GovCloud or AWS China is using different CAs).

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -0,0 +1,335 @@
...
+	keys = append(keys, jose.JSONWebKey{
+		Key:       publicKey,
+		KeyID:     kid,
+		Algorithm: string(alg),
+		Use:       "sig",
+	})
+
+	keySet, err := json.MarshalIndent(JSONWebKeySet{Keys: keys}, "", "    ")
+	if err != nil {
+		return nil, errors.New("JSON encoding of web key set failed")
```
joelddiaz, line 244: should errors.Wrap to provide the original `err`

joelddiaz:
/lgtm
we'll merge since Akhil is out for a few weeks, and keep pushing the code forward with future PRs.

pkg/cmd/provisioning/constants.go
```diff
@@ -0,0 +1,18 @@
...
+	// privateKeyFile is the name of the private key file created by "ccoctl create key-pair" command
+	privateKeyFile = "serviceaccount-signer.private"
+	// publicKeyFile is the name of the public key file created by "ccoctl create key-pair" command
+	publicKeyFile = "serviceaccount-signer.public"
+	// discoveryDocumentURI is a URI for the OpenID configuration discovery document
+	discoveryDocumentURI = ".well-known/openid-configuration"
+	// keysURI is a URI for public key that enables client to validate a JSON Web Token issued by the Identity Provider
+	keysURI = "keys.json"
+	// ccoctlAWSResourceTagKeyPrefix is the prefix of the tag key applied to the AWS resources created/shared by ccoctl
+	ccoctlAWSResourceTagKeyPrefix = "openshift.io/cloud-credential-operator"
```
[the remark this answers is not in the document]
dgoodwin: Seems ok to me, ccoctl is specific to cloud-credential-operator even if spun out.

pkg/cmd/provisioning/identity_provider.go
dgoodwin: @abhinavdahiya do you know the details on this?

joelddiaz: IIUC, we should be tagging "shared" if the object already exists (so that any uninstall process does not try to delete it), and "owned" if ccoctl created the object.
But how do we handle the installation of cluster1 where ccoctl creates everything, then perhaps the install for cluster2 is run with the same infra/name prefix, ccoctl should do what?
a) unmark the "owned" and set it to "shared" when it sees the pre-existing objects?
b) something else?

abhinavdahiya: I don't see why we want to distinguish owned vs shared?
shared would be useful when multiple different OIDC providers were using the same s3 bucket, but it looks like there is one bucket per OIDC provider, so there doesn't seem to be requirement for shared.

I would say just keep it owned for now..

joelddiaz: On a related note, in #304 I'm adding the ability to just output JSON instead of making the AWS API calls (that you can then `aws iam identity-provider --cli-input-json` with). Since tagging is a separate operation, should we generate JSON files, or assume since the user is manually handling/reviewing the creation of the objects that they will set up whatever tagging they want?

PR #304, opened 2021-02-25
add generate-only mode for ccoctl create identity-provider

joelddiaz:
- [x] add NewClientFromSession() for creating AWS clients
- [x ] rename param from 'infra-name' to 'name-prefix'
- [x] Allow ccoctl to assume the location of the public key file instead of forcing to provide it (will assume that `ccoctl create key-pair` output files exist)
- [x ] name S3 bucket holding OIDC config "<bucket>-oidc" instead of "<bucket>-installer"
- [x] Allow someone to have ccoctl simply generate files instead of actually making the AWS API calls.
- [x ] Generate the issuerURL fingerprint instead of hardcoding

### DIFF
--- a/pkg/cmd/provisioning/identity_provider.go
+++ b/pkg/cmd/provisioning/identity_provider.go
@@ -306,10 +306,7 @@ func identityProviderCmd(cmd *cobra.Command, args []string) {
 		log.Fatal(err)
 	}
 
-	awsClient := &aws.AWSClient{
-		IAMClient: iam.New(s),
-		S3Client:  s3.New(s),
-	}
+	awsClient := aws.NewClientFromSession(s)
 
 	err = createIdentityProvider(awsClient, CreateOpts.InfraName, CreateOpts.Region, CreateOpts.PublicKeyPath, CreateOpts.TargetDir)
 	if err != nil {
--- a/pkg/cmd/provisioning/identity_provider.go
+++ b/pkg/cmd/provisioning/identity_provider.go
@@ -65,7 +65,7 @@ type JSONWebKeySet struct {
 }
 
 func createIdentityProvider(client aws.Client, namePrefix, region, publicKeyPath, targetDir string) error {
-	bucketName := fmt.Sprintf("%s-installer", namePrefix)
+	bucketName := fmt.Sprintf("%s-oidc", namePrefix)
 	issuerURL := fmt.Sprintf("https://%s.s3.%s.amazonaws.com", bucketName, region)
 
 	var bucketTagValue string
--- a/pkg/cmd/provisioning/identity_provider.go
+++ b/pkg/cmd/provisioning/identity_provider.go
@@ -30,7 +30,7 @@ import (
 var (
 	// IdentityProviderOpts captures the options that affect creation of the identity provider
 	IdentityProviderOpts = options{
-		InfraName:     "",
+		NamePrefix:    "",
 		Region:        "",
 		PublicKeyPath: "",
 	}
@@ -63,8 +63,8 @@ type JSONWebKeySet struct {
 	Keys []jose.JSONWebKey `json:"keys"`
 }
 
-func createIdentityProvider(client aws.Client, infraName, region, publicKeyPath, targetDir string) error {
-	bucketName := fmt.Sprintf("%s-installer", infraName)
+func createIdentityProvider(client aws.Client, namePrefix, region, publicKeyPath, targetDir string) error {
+	bucketName := fmt.Sprintf("%s-installer", namePrefix)
 	issuerURL := fmt.Sprintf("https://%s.s3.%s.amazonaws.com", bucketName, region)
 
 	var bucketTagValue string
@@ -94,7 +94,7 @@ func createIdentityProvider(client aws.Client, infraName, region, publicKeyPath,
 		Tagging: &s3.Tagging{
 			TagSet: []*s3.Tag{
 				{
-					Key:   awssdk.String(fmt.Sprintf("%s/%s", ccoctlAWSResourceTagKeyPrefix, infraName)),
+					Key:   awssdk.String(fmt.Sprintf("%s/%s", ccoctlAWSResourceTagKeyPrefix, namePrefix)),
 					Value: awssdk.String(bucketTagValue),
 				},
 			},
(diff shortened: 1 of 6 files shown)


joelddiaz:
@dgoodwin i implemented all the PR feedback I had for #302 
The remaining bits are deciding on the label/tag naming (should it encode "cloud-credential-operator" in the name). See https://github.com/openshift/cloud-credential-operator/pull/302#discussion_r581309138

And whether the "owned" / "shared" tagging values buys us anything: https://github.com/openshift/cloud-credential-operator/pull/302#discussion_r581394190

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -57,145 +67,273 @@ var (
...
+						Key:   awssdk.String(fmt.Sprintf("%s/%s", ccoctlAWSResourceTagKeyPrefix, namePrefix)),
+						Value: awssdk.String(ownedCcoctlAWSResourceTagValue),
+					},
+				},
+			})
+			if err != nil {
+				return errors.Wrapf(err, "Failed to tag the identity provider with arn: %s", providerARN)
+			}
+
+			log.Printf("Identity Provider created with ARN: %s", providerARN)
```
dgoodwin, line 217: I probably missed this in the other pr but is log.Printf the way we're intentionally printing output in ccoctl?

dgoodwin (approved):
/lgtm

/hold

Feel free to cancel once the other issues are resolved, one I'm ok with, may want to hear from Abhinav on the tagging shared/owned.

pkg/cmd/provisioning/identity_provider.go
joelddiaz: It seems there is no log.Info in this log package (https://golang.org/pkg/log/), so I suppose Printf is what we're left with.

joelddiaz:
@dgoodwin after the discussions, i've updated the PR to remove the "shared" tag value, and make sure we only try to tag something that we've created.

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -314,23 +312,21 @@ func createOIDCBucket(client aws.Client, bucketName, namePrefix, region, targetD
...
-						Key:   awssdk.String(fmt.Sprintf("%s/%s", ccoctlAWSResourceTagKeyPrefix, namePrefix)),
-						Value: awssdk.String(bucketTagValue),
+			_, err = client.PutBucketTagging(&s3.PutBucketTaggingInput{
+				Bucket: awssdk.String(bucketName),
+				Tagging: &s3.Tagging{
+					TagSet: []*s3.Tag{
+						{
+							Key:   awssdk.String(fmt.Sprintf("%s/%s", ccoctlAWSResourceTagKeyPrefix, namePrefix)),
+							Value: awssdk.String(ownedCcoctlAWSResourceTagValue),
+						},
```
akhil-rane, line 323: @joelddiaz This will also tag preexisting buckets as `owned`. How should we handle a scenario where in a user has created a bucket with other objects (other then STS)? We should not allow `ccoctl delete` to remove that bucket right? This was my thinking when tagging preexisting buckets as `shared`. If bucket is shared `ccoctl delete` should only remove the tag and sts objects and not delete the entire bucket.

joelddiaz: the code lives under the `else` which should only run when ccoctl actually created the bucket.

akhil-rane: Oops my bad. Looks good

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -110,7 +110,7 @@ func createIdentityProvider(client aws.Client, infraName, region, publicKeyPath,
 		Body:    awssdk.ReadSeekCloser(strings.NewReader(discoveryDocumentJSON)),
 		Bucket:  awssdk.String(bucketName),
 		Key:     awssdk.String(discoveryDocumentURI),
-		Tagging: awssdk.String(fmt.Sprintf("%s/%s=%s", ccoctlAWSResourceTagKeyPrefix, infraName, ownedCcoctlAWSResourceTagValue)),
+		Tagging: awssdk.String(fmt.Sprintf("%s/%s=%s", ccoctlAWSResourceTagKeyPrefix, namePrefix, ownedCcoctlAWSResourceTagValue)),
```
abhinavdahiya: adding nameprefix to the tag that's being added for ownership seems incorrect. this should be use for names only then?

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -321,8 +321,8 @@ func NewIdentityProviderSetup() *cobra.Command {
 		Run: identityProviderCmd,
 	}
 
-	identityProviderSetupCmd.PersistentFlags().StringVar(&CreateOpts.InfraName, "infra-name", "", "Name prefix for all created AWS resources")
-	identityProviderSetupCmd.MarkPersistentFlagRequired("infra-name")
+	identityProviderSetupCmd.PersistentFlags().StringVar(&CreateOpts.NamePrefix, "name-prefix", "", "Name prefix for all created AWS resources")
+	identityProviderSetupCmd.MarkPersistentFlagRequired("name-prefix")
```
abhinavdahiya, line 464: my personal preference is keeping infra name, and explaining that resource names use this as prefix and tags use this to mark owned

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -58,145 +64,244 @@ var (
...
+	if generateOnly {
+		oidcIdentityProviderJSON := fmt.Sprintf(iamIdentityProviderTemplate, issuerURL)
 
-	if err != nil {
-		return errors.Wrapf(err, "Failed to upload JSON web key set (JWKS) in the S3 bucket %s", bucketName)
-	}
-	log.Printf("JSON web key set (JWKS) in the S3 bucket %s at %s updated", bucketName, keysURI)
+		err := saveToFile("AWS IAM Identity Provider", filepath.Join(targetDir, iamIdentityProviderFilename), []byte(oidcIdentityProviderJSON))
+		if err != nil {
+			errors.Wrap(err, "failed to save IAM Identity Provider file")
```
abhinavdahiya, line 164: return err here? not requiring us to do the if else ?

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -58,145 +64,244 @@ var (
...
 	}
+	return nil
+}
 
-	oidcProviderList, err := client.ListOpenIDConnectProviders(&iam.ListOpenIDConnectProvidersInput{})
+func createJSONWebKeySet(client aws.Client, publicKeyFilepath, bucketName, namePrefix, targetDir string, generateOnly bool) error {
+	jwks, err := buildJsonWebKeySet(publicKeyFilepath)
 	if err != nil {
-		return errors.Wrap(err, "Failed to fetch list of Identity Providers")
+		return errors.Wrap(err, "Failed to build JSON web key set from the public key")
```
abhinavdahiya, line 226: https://github.com/golang/go/wiki/CodeReviewComments#error-strings error wraps should not start with capital letters.

pkg/cmd/provisioning/identity_provider.go
dgoodwin: We don't really know the infra name at this point though do we? It's generated during install and we're pre-install at this point. Cluster name perhaps but gotta watch length carefully.

pkg/cmd/provisioning/identity_provider.go
joelddiaz: If we don't use some unique identifier, then all objects created by ccoctl will be tagged identically. How should we uniquely tag the objects?

pkg/cmd/provisioning/identity_provider.go
joelddiaz: because the OIDC pieces can be shared across clusters, it seemed like there was a potential for confusion if we linked the cluster infra name to the OIDC infra resources. that's why i was pushing away from the 'infra-name' terminology.

PR #306, opened 2021-03-09
iam-roles subcommand

joelddiaz:
This will take a directory with files representing the
CredentialsRequests that need IAM Roles/Policies created and the AWS IAM
Identity Provider ARN and create/update the Roles.

It also supports a '--generate-files-only' param which will place JSON
files of the IAM Roles and the IAM Role Policies that would have been
created (these can be used with the aws cli with the --cli-input-json
parameter).

Also, add make target for running 'go generate' (to update mock clients)

xref: https://issues.redhat.com/browse/CCO-70

### DIFF
--- a/pkg/cmd/provisioning/identity_provider.go
+++ b/pkg/cmd/provisioning/identity_provider.go
@@ -306,10 +306,7 @@ func identityProviderCmd(cmd *cobra.Command, args []string) {
 		log.Fatal(err)
 	}
 
-	awsClient := &aws.AWSClient{
-		IAMClient: iam.New(s),
-		S3Client:  s3.New(s),
-	}
+	awsClient := aws.NewClientFromSession(s)
 
 	err = createIdentityProvider(awsClient, CreateOpts.InfraName, CreateOpts.Region, CreateOpts.PublicKeyPath, CreateOpts.TargetDir)
 	if err != nil {
--- a/pkg/cmd/provisioning/identity_provider.go
+++ b/pkg/cmd/provisioning/identity_provider.go
@@ -65,7 +65,7 @@ type JSONWebKeySet struct {
 }
 
 func createIdentityProvider(client aws.Client, namePrefix, region, publicKeyPath, targetDir string) error {
-	bucketName := fmt.Sprintf("%s-installer", namePrefix)
+	bucketName := fmt.Sprintf("%s-oidc", namePrefix)
 	issuerURL := fmt.Sprintf("https://%s.s3.%s.amazonaws.com", bucketName, region)
 
 	var bucketTagValue string
--- a/pkg/cmd/provisioning/identity_provider.go
+++ b/pkg/cmd/provisioning/identity_provider.go
@@ -30,7 +30,7 @@ import (
 var (
 	// IdentityProviderOpts captures the options that affect creation of the identity provider
 	IdentityProviderOpts = options{
-		InfraName:     "",
+		NamePrefix:    "",
 		Region:        "",
 		PublicKeyPath: "",
 	}
@@ -63,8 +63,8 @@ type JSONWebKeySet struct {
 	Keys []jose.JSONWebKey `json:"keys"`
 }
 
-func createIdentityProvider(client aws.Client, infraName, region, publicKeyPath, targetDir string) error {
-	bucketName := fmt.Sprintf("%s-installer", infraName)
+func createIdentityProvider(client aws.Client, namePrefix, region, publicKeyPath, targetDir string) error {
+	bucketName := fmt.Sprintf("%s-installer", namePrefix)
 	issuerURL := fmt.Sprintf("https://%s.s3.%s.amazonaws.com", bucketName, region)
 
 	var bucketTagValue string
@@ -94,7 +94,7 @@ func createIdentityProvider(client aws.Client, infraName, region, publicKeyPath,
 		Tagging: &s3.Tagging{
 			TagSet: []*s3.Tag{
 				{
-					Key:   awssdk.String(fmt.Sprintf("%s/%s", ccoctlAWSResourceTagKeyPrefix, infraName)),
+					Key:   awssdk.String(fmt.Sprintf("%s/%s", ccoctlAWSResourceTagKeyPrefix, namePrefix)),
 					Value: awssdk.String(bucketTagValue),
 				},
 			},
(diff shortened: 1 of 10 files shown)


pkg/cmd/provisioning/roles.go
```diff
@@ -0,0 +1,337 @@
...
+		Run: iamRolesCmd,
+	}
+
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.NamePrefix, "name-prefix", "", "Name prefix for all created AWS resources")
+	iamRolesSetupCmd.MarkPersistentFlagRequired("name-prefix")
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.Region, "region", "", "AWS region where the S3 OpenID Connect endpoint will be created")
+	iamRolesSetupCmd.MarkPersistentFlagRequired("region")
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.CredRequestDir, "credentials-requests-dir", "", "Directory containing files of CredentialsRequests to create IAM Roles for")
+	iamRolesSetupCmd.MarkPersistentFlagRequired("credentials-requests-dir")
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.IdentityProviderARN, "identity-provider-arn", "", "ARN of IAM Identity provider for IAM Role trust relationship")
```
dgoodwin, line 332: "can be created with create identity-provider sub-command"

pkg/cmd/provisioning/roles.go
```diff
@@ -0,0 +1,337 @@
...
+	iamRolesSetupCmd := &cobra.Command{
+		Use: "iam-roles",
+		Run: iamRolesCmd,
+	}
+
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.NamePrefix, "name-prefix", "", "Name prefix for all created AWS resources")
+	iamRolesSetupCmd.MarkPersistentFlagRequired("name-prefix")
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.Region, "region", "", "AWS region where the S3 OpenID Connect endpoint will be created")
+	iamRolesSetupCmd.MarkPersistentFlagRequired("region")
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.CredRequestDir, "credentials-requests-dir", "", "Directory containing files of CredentialsRequests to create IAM Roles for")
```
dgoodwin, line 330: "can be created with oc adm release extract --credentials-requests --cloud=aws"

pkg/cmd/provisioning/roles.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// NewIAMRolesSetup provides the "create iam-roles" subcommand
+func NewIAMRolesSetup() *cobra.Command {
+	iamRolesSetupCmd := &cobra.Command{
+		Use: "iam-roles",
+		Run: iamRolesCmd,
+	}
+
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.NamePrefix, "name-prefix", "", "Name prefix for all created AWS resources")
```
dgoodwin, line 326: With sts-preflight and this, I wasn't totally clear what this name prefix was meant to be. Do I just make it up? It feels like infraID but that's not known until later. Presumably this is just something the user can pick freeform? 

If so lets just note that in the description.

pkg/cmd/provisioning/roles.go
```diff
@@ -0,0 +1,337 @@
...
+	iamRolesSetupCmd.MarkPersistentFlagRequired("name-prefix")
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.Region, "region", "", "AWS region where the S3 OpenID Connect endpoint will be created")
+	iamRolesSetupCmd.MarkPersistentFlagRequired("region")
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.CredRequestDir, "credentials-requests-dir", "", "Directory containing files of CredentialsRequests to create IAM Roles for")
+	iamRolesSetupCmd.MarkPersistentFlagRequired("credentials-requests-dir")
+	iamRolesSetupCmd.PersistentFlags().StringVar(&CreateOpts.IdentityProviderARN, "identity-provider-arn", "", "ARN of IAM Identity provider for IAM Role trust relationship")
+	iamRolesSetupCmd.MarkPersistentFlagRequired("identity-provider-arn")
+	iamRolesSetupCmd.PersistentFlags().BoolVar(&CreateOpts.GenerateOnly, "generate-files-only", false, "Skip creating objects, and just save what would have been created into files")
```
dgoodwin, line 334: WDYT about -o json here?

pkg/cmd/provisioning/identity_provider.go
```diff
@@ -324,12 +454,12 @@ func NewIdentityProviderSetup() *cobra.Command {
 		Run: identityProviderCmd,
 	}
 
-	identityProviderSetupCmd.PersistentFlags().StringVar(&CreateOpts.InfraName, "infra-name", "", "Name prefix for all created AWS resources")
-	identityProviderSetupCmd.MarkPersistentFlagRequired("infra-name")
+	identityProviderSetupCmd.PersistentFlags().StringVar(&CreateOpts.NamePrefix, "name-prefix", "", "Name prefix for all created AWS resources")
+	identityProviderSetupCmd.MarkPersistentFlagRequired("name-prefix")
```
dgoodwin, line 458: This helps clear up it's not infra id :) Does the name prefix here have to match future executions of the command? I'm guessing not. As per below I'd just clarify that this is just to help users identify resources visually.

pkg/cmd/provisioning/roles.go
```diff
@@ -0,0 +1,337 @@
...
+				default:
+					return "", err
+				}
+
+			}
+		} else {
+			role = outRole.Role
+			log.Printf("Existing role %s found", *role.Arn)
+
+			// TODO: implement idemponent apply/update for role
```
dgoodwin, line 171: Future PR?

pkg/cmd/provisioning/roles.go
joelddiaz: If we did something like '-o json' it might get a little weird with a high-level 'ccoctl create all' command where not all of the output files will be of type JSON (the public/private key). And I was planning on having the installer manifests generation output yaml files (not for any particular reason other than they're easier to edit IMO). The reason I'm generating JSON for the aws CLI files, is that it seems like that is what the tool provides when you ask for a skeleton object (I would've preferred YAML to be honest) WDYT?

pkg/cmd/provisioning/identity_provider.go
joelddiaz: If you run ccoctl with different name-prefix values, you'll end up with a second set of objects altogether..

pkg/cmd/provisioning/roles.go
joelddiaz: Possibly. Doing a GetRolePolicy() and comparing what we get back to what we generate is a bit involved as AWS will reformat things so you can't do a straight string comparison. I just wanted to note here that if we were really motivated, we could avoid making this mutating call.

joelddiaz:
/hold
wait for #304

pkg/cmd/provisioning/roles.go
dgoodwin: Yup that's fair. I'd like a more concise CLI param though. What about --dry-run and the command arguments inform you it will write files?

joelddiaz:
@dgoodwin #304 finally merged, so this is good for a review as it is next in line

PR #307, opened 2021-03-09
Ccoctl docs

joelddiaz:
update ccoctl docs

this PR builds on top of #306


joelddiaz:
/hold
wait for #306

joelddiaz:
/unhold
@dgoodwin #306 merged, so this is next in line (just documentation)

PR #309, opened 2021-03-17
Bug 1940142: prepare correct clouds.yaml data

Fedosin:
In OCP versions <4.6 we included the original host cacert path in clouds.yaml. It is wrong because internally we always use
"/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem".

The issue was fixed in the installer [1] in 4.6 but we still have to replace incorrect values in the generated clouds.yaml for the
previous versions.

This commit checks if the value is not correct and replaces it with the canonical one on the fly.

[1] openshift/installer#4227

### DIFF
--- a/pkg/openstack/actuator.go
+++ b/pkg/openstack/actuator.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"regexp"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -39,6 +40,7 @@ import (
 
 const (
 	rootOpenStackCredsSecretKey = "clouds.yaml"
+	caCertFile                  = "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem"
 )
 
 type OpenStackActuator struct {
@@ -110,6 +112,8 @@ func (a *OpenStackActuator) sync(ctx context.Context, cr *minterv1.CredentialsRe
 		return err
 	}
 
+	clouds = a.fixInvalidCACertFile(clouds)
+
 	logger.Debugf("provisioning secret")
 	existingSecret, err := a.loadExistingSecret(cr)
 	if err != nil {
@@ -127,6 +131,12 @@ func (a *OpenStackActuator) sync(ctx context.Context, cr *minterv1.CredentialsRe
 	}
 
 	return nil
+}
+
+// fixInvalidCACertFile insures that clouds.yaml has the right CACertFile value
+// For more information: https://bugzilla.redhat.com/show_bug.cgi?id=1940142
+func (a *OpenStackActuator) fixInvalidCACertFile(content string) string {
+	return regexp.MustCompile(`cacert: .*`).ReplaceAllString(content, "cacert: "+caCertFile)
 }
 
 func (a *OpenStackActuator) syncCredentialSecret(cr *minterv1.CredentialsRequest, clouds string, existingSecret *corev1.Secret, userPolicy string, logger log.FieldLogger) error {
--- a/pkg/openstack/actuator_test.go
+++ b/pkg/openstack/actuator_test.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2021 The OpenShift Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openstack
+
+import (
+	"regexp"
+	"testing"
+)
+
+func TestOpenStackActuatorFixInvalidCACertFile(t *testing.T) {
+	const noCACert = `
+clouds:
+  openstack:
+    auth:
+      auth_url: http://1.2.3.4:5000
+      password: password
+      project_domain_name: Default
+      project_name: openshift
+      user_domain_name: Default
+      username: openshift
+    identity_api_version: "3"
+    region_name: regionOne
+    verify: true
+`
+
+	const incorrectCACert = `
+clouds:
+  openstack:
+    auth:
+      auth_url: http://1.2.3.4:5000
+      password: password
+      project_domain_name: Default
+      project_name: openshift
+      user_domain_name: Default
+      username: openshift
+    cacert: /incorrect/path/to/ca-bundle.pem
+    identity_api_version: "3"
+    region_name: regionOne
+    verify: true
+`
+
+	const correctCACert = `
+clouds:
+  openstack:
+    auth:
+      auth_url: http://1.2.3.4:5000
+      password: password
+      project_domain_name: Default
+      project_name: openshift
+      user_domain_name: Default
+      username: openshift
+    cacert: /etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem
+    identity_api_version: "3"
+    region_name: regionOne
+    verify: true
+`
+
+	tests := []struct {
+		name    string
+		arg     string
+		want    string
+		wantErr bool
+	}{
+		{"noCACert", noCACert, "", false},
+		{"incorrectCACert", incorrectCACert, "cacert: " + caCertFile, false},
+		{"correctCACert", incorrectCACert, "cacert: " + caCertFile, false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			a := &OpenStackActuator{}
+			got := a.fixInvalidCACertFile(tt.arg)
+			caCertFile := regexp.MustCompile(`cacert: .*`).FindString(got)
+			if caCertFile != tt.want {
+				t.Errorf("OpenStackActuator.fixInvalidCACertFile() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}


pkg/openstack/actuator.go
```diff
@@ -31,14 +31,20 @@ import (
...
 	"github.com/openshift/cloud-credential-operator/pkg/operator/constants"
 	actuatoriface "github.com/openshift/cloud-credential-operator/pkg/operator/credentialsrequest/actuator"
 	"github.com/openshift/cloud-credential-operator/pkg/operator/utils"
+
+	yaml "gopkg.in/yaml.v2"
 )
 
 const (
 	rootOpenStackCredsSecretKey = "clouds.yaml"
+	rootOpenStackCloudName      = "openstack"
```
mandre, line 46: We can't assume the cloud entry is named `openstack` - we need to sanitize every cloud entries in the `clouds.yaml` file.

mandre (changes requested):

pkg/openstack/actuator.go
Fedosin: In the installer this value is hardcoded, so we always have "openstack" cloud there: https://github.com/openshift/installer/blob/d237b9dee575c6e9ceeba80fd3e137231d907cab/pkg/asset/machines/openstack/machines.go#L29-L30

pkg/openstack/actuator.go
```diff
@@ -129,6 +136,12 @@ func (a *OpenStackActuator) sync(ctx context.Context, cr *minterv1.CredentialsRe
 	return nil
 }
 
+func (a *OpenStackActuator) prepareCloudsData(clouds string) (string, error) {
+	cloud, err := gophercloudutils.
```
mdbooth: Looks like a syntax error? Probably shouldn't commit this, even if it is fixed in the next commit.

mdbooth: /hold for this

pkg/openstack/actuator.go
```diff
@@ -136,10 +140,30 @@ func (a *OpenStackActuator) sync(ctx context.Context, cr *minterv1.CredentialsRe
...
-	cloud, err := gophercloudutils.
+func (a *OpenStackActuator) prepareCloudsData(content string) (string, error) {
+	var clouds clientconfig.Clouds
 
-	return clouds, nil
+	err := yaml.Unmarshal([]byte(content), &clouds)
+	if err != nil {
+		return "", err
+	}
+
```
mdbooth, line 150: Please can you add a comment linking to the bug here? Any future maintainer looking at this will see that we never write an incorrect value and may assume that this can be safely removed. Finding an original git commit message is not always easy.

Related: perhaps call the method something explicit like fixInvalidCACertFile?

pkg/openstack/actuator.go
mandre: > In the installer this value is hardcoded, so we always have "openstack" cloud there: https://github.com/openshift/installer/blob/d237b9dee575c6e9ceeba80fd3e137231d907cab/pkg/asset/machines/openstack/machines.go#L29-L30

What happens if someone changes that in the installer code?
Sanitizing all entries is not more work and we don't make assumption about the cloud name.

pkg/openstack/actuator.go
```diff
@@ -37,10 +37,14 @@ import (
...
 	actuatoriface "github.com/openshift/cloud-credential-operator/pkg/operator/credentialsrequest/actuator"
 	"github.com/openshift/cloud-credential-operator/pkg/operator/utils"
+
+	yaml "gopkg.in/yaml.v2"
 )
 
 const (
 	rootOpenStackCredsSecretKey = "clouds.yaml"
+	rootOpenStackCloudName      = "openstack"
+	caCertFile                  = "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem"
```
mdbooth, line 47: I wondered about hard-coding this path here, and if other code pulls it from somewhere. However, a quick look in library-go reveals it's hard-coded in several places there too, so I'm assuming this is safe.

pkg/openstack/actuator.go
```diff
@@ -136,10 +140,30 @@ func (a *OpenStackActuator) sync(ctx context.Context, cr *minterv1.CredentialsRe
...
+
+	if clouds.Clouds[rootOpenStackCloudName].CACertFile == caCertFile || clouds.Clouds[rootOpenStackCloudName].CACertFile == "" {
+		// Don't have to modify anything
+		return content, nil
+	}
+
+	// CACertFile has an incorrect value, we need to replace it with the canonical one
+	newCloud := clouds.Clouds[rootOpenStackCloudName]
+	newCloud.CACertFile = caCertFile
+	clouds.Clouds[rootOpenStackCloudName] = newCloud
```
mdbooth, line 159:
```
clouds.Clouds[rootOpenStackCloudName].CACertFile = caCertFile
```

mdbooth (changes requested):
We really should have a unit test for this.

pkg/openstack/actuator.go
Fedosin: @mandre first, why would you want to change the cloud name from "openstack" to something else in the installer code? second, a lot of components assume that the entry is called "openstack" in the clouds.yaml.

CIRO: https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/swift/swift.go#L113
Manila CSI driver operator: https://github.com/openshift/csi-driver-manila-operator/blob/master/pkg/controllers/secret/secretsync.go#L36-L37
Cinder CSI driver operator: https://github.com/openshift/openstack-cinder-csi-driver-operator/blob/master/assets/configmap.yaml#L7
CCMO: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/openstack/assets/config.yaml#L7

pkg/openstack/actuator.go
Fedosin: Yeah, it is hardcoded in multiple places, including our installer, image registry and csi drivers...

pkg/openstack/actuator.go
Fedosin: thank you for the suggestion but unfortunately your code won't compile with the next error:

> cannot assign to struct field clouds.Clouds[rootOpenStackCloudName].CACertFile in mapcompilerUnaddressableFieldAssign

pkg/openstack/actuator.go
Fedosin: done

pkg/openstack/actuator.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2019 The OpenShift Authors.
-
```
mdbooth, line 3: Accidental damage to the copyright notice?

mdbooth (changes requested):
The copyright notice damage is trivial, but we ought to fix it.

However, I'm still uncomfortable with the lack of a unit test.

pkg/openstack/actuator.go
Fedosin: Thanks! I have no idea how this could happen...

pkg/openstack/actuator.go
mandre: My point being, hardcoding the cloud name is a code smell.

It doesn't cost us anything to sanitize all entries of the clouds.yaml and it makes the code more robust.

mdbooth:
I pushed a unit test here: https://github.com/mdbooth/cloud-credential-operator/commit/fa9618f5af4a285f1952dbcb762a63326c3a3f83

pkg/openstack/actuator.go
```diff
@@ -129,6 +140,34 @@ func (a *OpenStackActuator) sync(ctx context.Context, cr *minterv1.CredentialsRe
 	return nil
 }
 
+// fixInvalidCACertFile insures that clouds.yaml has the right CACertFile value
+// For more information: https://bugzilla.redhat.com/show_bug.cgi?id=1940142
+func (a *OpenStackActuator) fixInvalidCACertFile(content string) (string, error) {
+	var clouds clientconfig.Clouds
```
wking, line 146: Not my code, but `gophercloud` seems like a heavy dependency for massaging some YAML.  Is the format of this content so complicated that it's worth adding the dependency instead of inlining enough knowledge locally to make the required pivot?

mdbooth: That's a good point. I've pushed a modified version of this patch here which doesn't use gophercloud: https://github.com/mdbooth/cloud-credential-operator/tree/correct_cacert

I'll discuss with @Fedosin if he wants to use it.

joelddiaz (commented):
Is it unreasonable to modify the kube-system/openstack-credentials Secret directly? Then we know that a cluster that upgrades through 4.8 has a "fixed" Secret, and we can remove this one-time code for 4.9?

I'm a little uncomfortable with setting some precedent where CCO is taught about good/bad cloud secret contents as CCO is really just passing these contents along unchanged (at least for platforms like OpenStack). So I'd feel better with a special-case for one release, and then CCO can return to being blind to the contents of these Secrets.

I'd also be interested in an approach that doesn't bring in the gophercloud dependency just for this change.

mdbooth:
@joelddiaz I posted https://github.com/mdbooth/cloud-credential-operator/tree/correct_cacert which doesn't have the gophercloud dep, btw. I'll discuss it with @Fedosin when he's back from PTO.

mdbooth:
@joelddiaz As for modifying the kube-system/openstack-credentials secret directly, I actually thought that's what we were doing here (sorry I hadn't looked far enough beyond the immediate context).

Yes, that is definitely what we want to do. Is there anywhere we could move this code such that the operator would modify the 'root' secret automatically?

mdbooth:
@joelddiaz It looks like secretannotator does something similar already: it modifies the 'root' secret, although only to add an annotation. There is currently no OpenStack implementation.

Would it make sense to implement that for OpenStack and mutute the root secret in that controller?

joelddiaz:
The secretannotator controller does seem like a good place to put this kind of operation. @dgoodwin WDYT?

dgoodwin:
sgtm

mdbooth:
@dgoodwin @joelddiaz Looking at that controller, presumably we're (OpenStack) going to want to implement that at some point soon anyway when we add application credential support.

The existing behaviour is to run the AWS controller on an OpenStack cluster. This looks like a noop, because we'll ignore all events. So...

I propose for the smallest initial implementation to add a controller which doesn't add any secret annotations and just fixes this bug if it finds it. We'll then build on this later to validate credentials and annotate them appropriately. Sound ok?

joelddiaz:
Yes. No need to be annotating the secret since for OpenStack it is always assumed to be in "passthrough".

joelddiaz:
@mdbooth i am curious about this adding of "application credential support". What is planned? (I ask because we're in the middle of rethinking/plannning the long-term approach for what the cloud-credential-operator will be responsible for).

mdbooth:
@joelddiaz https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/14/html/users_and_identity_management_guide/application_credentials

This could support the current tenant credentials model, a cluster-specific application secret mode, and also the 'mint' mode. The feature isn't planned yet, though, so now's the time to let us know if it's changing!

joelddiaz:
@mdbooth this "application credentials" sounds a lot like implementing CCO "mint" mode for OpenStack. Please do reach out to us before anyone takes on this task, as our approach for this may be changing substantially..

mdbooth:
@joelddiaz It would allow us to enable mint mode, but even using application credentials with passthrough would be useful.

Where's the best place to discuss?

joelddiaz:
> @joelddiaz It would allow us to enable mint mode, but even using application credentials with passthrough would be useful.
> 
> Where's the best place to discuss?

@akhil-rane is working on an enhanecment proposal discussing this. I've asked him to post his branch where those docs are coming together.

akhil-rane:
@mdbooth [here](https://github.com/akhil-rane/enhancements/blob/cco_manual_mode_is_default/enhancements/installer/credentials-management-outside-openshift-cluster.md) is the intial draft of the proposal

Fedosin:
Closed in favor of https://github.com/openshift/cloud-credential-operator/pull/314

PR #310, opened 2021-03-17
Bug 1927017: Context cancel on stopped leading

suhanime:
Post proxy CA change, pod was not relinquishing its leader election
quickly due to os.exit. Changed it to run cancel context to propagate
the exit cleanly

### DIFF
--- a/pkg/operator/configmap/configmap_controller.go
+++ b/pkg/operator/configmap/configmap_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"crypto/md5"
 	"fmt"
-	"os"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -60,10 +59,12 @@ func NewReconciler(mgr manager.Manager) *ReconcileConfigMap {
 		// Controller only watches this one single configmap.
 		"configmap": fmt.Sprintf("%s/%s", ccoNamespace, configMapName),
 	}
+	context, cancel := context.WithCancel(context.TODO())
 	return &ReconcileConfigMap{
-		Client: mgr.GetClient(),
-		logger: log.WithFields(logFields),
-		exit:   exitFunc,
+		Client:  mgr.GetClient(),
+		logger:  log.WithFields(logFields),
+		context: context,
+		exit:    cancel,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


joelddiaz (commented):
So does this cancel() get to be called even when the os.Exit() is called from https://github.com/openshift/cloud-credential-operator/blob/master/pkg/operator/configmap/configmap_controller.go#L146 ?

suhanime:
> So does this cancel() get to be called even when the os.Exit() is called from https://github.com/openshift/cloud-credential-operator/blob/master/pkg/operator/configmap/configmap_controller.go#L146 ?

No... The current plan (which I am working on) is to pass a context to the configmap controller and cancel the context when proxy CA change is detected. However I will keep these changes as well as os.exit is not a great idea regardless.

joelddiaz (commented):
/lgtm

pkg/operator/configmap/configmap_controller.go
```diff
@@ -60,10 +59,12 @@ func NewReconciler(mgr manager.Manager) *ReconcileConfigMap {
 		// Controller only watches this one single configmap.
 		"configmap": fmt.Sprintf("%s/%s", ccoNamespace, configMapName),
 	}
+	context, cancel := context.WithCancel(context.TODO())
```
abhinavdahiya, line 62: canceling this context doesn't do anything to the context of the leaderelection code. how this this signalling to the leaderelection code to stop running and go on with release on cancel code path?

pkg/operator/configmap/configmap_controller.go
```diff
@@ -108,6 +109,7 @@ type ReconcileConfigMap struct {
 	client.Client
 	logger            log.FieldLogger
 	configMapDataHash string
+	context           context.Context
```
akhil-rane, line 112: @suhanime it is advised not to store a context inside a struct. Any specific reason why we are doing it here?

https://github.com/golang/go/blob/2887ef499a0e01a3b6fe92bf874843f9f2d05167/src/context/context.go#L29-L36

joelddiaz: For the case (not implemented in this PR) where we want to pass the context generated up in https://github.com/openshift/cloud-credential-operator/blob/master/pkg/cmd/operator/cmd.go down into a runtime controller, where would we otherwise store it?

PR #318, opened 2021-04-07
WIP: Make ccoctl do additive updates to iam role policies

akhil-rane:
For upgrades we should ensure that old permissions are not removed
while updating the iam roles.

x-ref: https://issues.redhat.com/browse/CCO-84


akhil-rane:
This is no longer relevant, the intended work is already taken care of as part of this [PR](https://github.com/openshift/cloud-credential-operator/pull/364).

/close

PR #320, opened 2021-04-08
Bug 1927017: Use file observer for Proxy CA changes

suhanime:
Config map controller observed changes for Proxy CA and restarted the
pod when it did. Whenever it saw a change in Proxy CA, it restarted the
pod without releasing the lease, so the new pod took about 8 minutes to
acquire the new lease.

In order to support a graceful shutdown, this commit also aims at
dropping the config map controller and instead observing the changes via
a file observer. Whenever there is a change in Proxy CA, the observer
will call cancel context of leader election kick-starting a graceful
shutdown, and restart the pod.

/assign @joelddiaz 
/cc @abhinavdahiya


pkg/cmd/operator/cmd.go
```diff
@@ -113,6 +120,11 @@ func NewOperator() *cobra.Command {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
+			caConfigMapPath := filepath.Join(caConfigMapMountPath, caConfigMapName)
+			if _, err := os.Stat(caConfigMapPath); err != nil {
+				go terminateWhenProxyChanges(caConfigMapPath, cancel)
+			}
```
suhanime, line 126: I think the place where I am calling the goroutine is not correct - any pointers?

pkg/cmd/operator/cmd.go
```diff
@@ -113,6 +120,11 @@ func NewOperator() *cobra.Command {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
+			caConfigMapPath := filepath.Join(caConfigMapMountPath, caConfigMapName)
+			if _, err := os.Stat(caConfigMapPath); err == nil {
+				go terminateWhenProxyChanges(caConfigMapPath, cancel)
```
abhinavdahiya, line 125: we already start the fileobserver in a go routine https://github.com/openshift/cloud-credential-operator/pull/320/files#diff-2f87b1f2687fa01b2bd650174a4d17234de24ce7061f469e90c3ea56d72a77ddR232-R236, so why start this function also in go routine?

pkg/cmd/operator/cmd.go
```diff
@@ -113,6 +120,11 @@ func NewOperator() *cobra.Command {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
```
abhinavdahiya, line 122: i would personally would have it seen like,

// create the file observer

// add reactor for proxy files

// run the file observer

pkg/cmd/operator/cmd.go
```diff
@@ -113,6 +120,11 @@ func NewOperator() *cobra.Command {
...
 
+			caConfigMapPath := filepath.Join(caConfigMapMountPath, caConfigMapName)
+			if _, err := os.Stat(caConfigMapPath); err == nil {
```
abhinavdahiya, line 124: also i would like us to comment why we are checking for file to exist to start the reactor?

pkg/cmd/operator/cmd.go
suhanime: there wasn't any specific need - changed

pkg/cmd/operator/cmd.go
```diff
@@ -196,3 +208,30 @@ func (writer glogWriter) Write(data []byte) (n int, err error) {
...
+			log.Info("Proxy CA configmap change detected, restarting pod")
+			cancel()
+			return nil
+		},
+		fileContents,
+	)
+	stop := make(chan struct{})
+	go func() {
+		log.WithField("file", path).Info("running file observer")
+		obs.Run(stop)
```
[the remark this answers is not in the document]
suhanime: nice catch - changed

joelddiaz (commented):
/lgtm

PR #321, opened 2021-04-08
Bug 1948628: Add aws subcommand to ccoctl and additional UX changes

akhil-rane:
- groups all the creation/deletion commands for AWS cloud into a separate subcommand
- name-prefix option is changed to name
- All the created resources will have additional Name tag
- iam-create-roles subcommand will not make any updates to existing roles. Only missing roles will be created

x-ref: https://issues.redhat.com/browse/CCO-89


pkg/cmd/provisioning/aws/create-iam-roles.go
```diff
@@ -1,8 +1,9 @@
-package provisioning
+package aws
 
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/cloud-credential-operator/pkg/cmd/provisioning"
```
joelddiaz, line 6: can we move this import into its own grouping separate from the standard import packages?

pkg/cmd/provisioning/aws/create_identity_provider.go
```diff
@@ -447,9 +457,9 @@ spec:
 	return nil
 }
 
-func identityProviderCmd(cmd *cobra.Command, args []string) {
+func CreateIdentityProviderCmd(cmd *cobra.Command, args []string) {
```
joelddiaz, line 460: why exporting this function outside the aws package?

pkg/cmd/provisioning/aws/create_all.go
```diff
@@ -0,0 +1,116 @@
...
+	if publicKeyPath == "" {
+		publicKeyPath = path.Join(CreateAllOpts.TargetDir, publicKeyFile)
+	}
+
+	if err := createKeys(CreateAllOpts.TargetDir); err != nil {
+		log.Fatalf("Failed to create public/private key pair: %s", err)
+	}
+
+	identityProviderARN, err := createIdentityProvider(awsClient, CreateAllOpts.Name, CreateAllOpts.Region, publicKeyPath, CreateAllOpts.TargetDir, false)
+	fmt.Printf("ARN: %+v\n", identityProviderARN)
```
joelddiaz, line 49: Should this be a log output?

akhil-rane: This is not needed as we already print the created ARN inside `createIdentityProvider` funciton

sample: `2021/04/09 12:18:30 Identity Provider created with ARN: <arn>`

akhil-rane:
@joelddiaz addressed your feedback

dgoodwin:
Taking a look as well.

docs/ccoctl.md
```diff
@@ -42,20 +44,34 @@ $ oc adm release extract --credentials-requests --cloud=aws --to=./credrequests
...
-## Deleting resources
+### Creating all the required resources together
+
+To create all the above mentioned resources in one go, run
+
+```bash
+$ oc adm release extract --credentials-requests --cloud=aws --to=./credrequests quay.io/path/to/openshift-release:version
+```
+
+Then you can use `ccoctl` to process each CredentialsRequest object in the `./credrequests` directory (from the example above)
```
dgoodwin, line 62:
```suggestion
Then you can use `ccoctl` to process all CredentialsRequest objects in the `./credrequests` directory (from the example above)
```

Nit: each made me think I need to run the command once per CredRequest.

dgoodwin (commented):
Just one nit.

akhil-rane:
@dgoodwin made the doc update

pkg/cmd/provisioning/aws/create-iam-roles.go
```diff
@@ -413,6 +415,7 @@ func NewCreateIAMRolesCmd() *cobra.Command {
 	createIAMRolesCmd.MarkPersistentFlagRequired("credentials-requests-dir")
 	createIAMRolesCmd.PersistentFlags().StringVar(&CreateIAMRolesOpts.IdentityProviderARN, "identity-provider-arn", "", "ARN of IAM Identity provider for IAM Role trust relationship (can be created with the 'create identity-provider' sub-command)")
 	createIAMRolesCmd.MarkPersistentFlagRequired("identity-provider-arn")
+	createIAMRolesCmd.PersistentFlags().StringVar(&CreateIAMRolesOpts.Region, "region", "", "AWS region endpoint only required for GovCloud")
```
joelddiaz, line 418: Should we just make it required no matter what, that way AWS, GovCloud, and China should "just work"?

akhil-rane: there is possibility that someone will ask why we need it for non GovCloud

akhil-rane: Best thing in my opinion will be to document it and say it is required only for GovCloud (we need to test it first)

PR #325, opened 2021-04-13
Bug 1948628: Fix ccoctl create aws iam-roles command and docs

akhil-rane:
- fixes wrongly generated trust relationship document for iam roles
- fixes public-key-file parameter typo in the documentation


joelddiaz (commented):
/lgtm

PR #327, opened 2021-04-19
Bug 1949866: use initialization structure for ID provider

joelddiaz:
fix param init for create-identity-provider
fix param init for create-iam-roles

### DIFF
--- a/pkg/cmd/provisioning/aws/create_identity_provider.go
+++ b/pkg/cmd/provisioning/aws/create_identity_provider.go
@@ -11,13 +11,14 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
-	"github.com/openshift/cloud-credential-operator/pkg/cmd/provisioning"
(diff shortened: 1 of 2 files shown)


pkg/cmd/provisioning/aws/create_identity_provider.go
```diff
@@ -11,14 +11,15 @@ import (
...
-	"github.com/openshift/cloud-credential-operator/pkg/cmd/provisioning"
 	"io/ioutil"
 	"log"
 	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/openshift/cloud-credential-operator/pkg/cmd/provisioning"
+
```
akhil-rane, line 22: this import should come last, along with other openshift import

joelddiaz: ahh...VS Code...i almost with it wouldn't auto-import things

PR #328, opened 2021-04-19
Bug 1949862: avoid referencing object after error

joelddiaz:
After a GetRole() has failed, do not assume the returned value is
non-nil.


akhil-rane (commented):
/lgtm

joelddiaz:
@akhil-rane got some weird double-import error that is now fixed

PR #330, opened 2021-04-27
Bug 1953077: rework GCP passthrough permissions checking

joelddiaz:
In GCP Passthrough mode, the permissions are dynamically checked before
copying the secret contents over. This is done by performing a
TestIamPermissions() call against the current GCP project.

There are certain permissions that make no sense to check against the
project level. Periodically fetch a list of testable permissions using the
QueryTestablePermissions() API call.

Before making a TestIamPermissions() call, filter the permissions through the
allowed list to avoid needlessly receiving errors when calling TestIamPermissions()
against a GCP Project resource.

### DIFF
--- a/pkg/operator/utils/gcp/utils.go
+++ b/pkg/operator/utils/gcp/utils.go
@@ -17,20 +17,27 @@ limitations under the License.
 package gcp
 
 import (
+	"context"
 	"fmt"
 	"regexp"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 
-	minterv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
-	ccgcp "github.com/openshift/cloud-credential-operator/pkg/gcp"
-
 	cloudresourcemanager "google.golang.org/api/cloudresourcemanager/v1"
+	iamadminpb "google.golang.org/genproto/googleapis/iam/admin/v1"
 
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/sets"
+
+	minterv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+	ccgcp "github.com/openshift/cloud-credential-operator/pkg/gcp"
 )
+
+type testablePermissions struct {
+	permSet     sets.String
+	lastUpdated time.Time
+}
 
 var (
 	// CredMintingPermissions is a list of GCP permissions needed to run in the mode where the
@@ -64,19 +71,8 @@ var (
 	}
 
 	credentailRequestScheme = runtime.NewScheme()
-	credentialRequestCodec  = serializer.NewCodecFactory(credentailRequestScheme)
 
-	invalidPermsForProjectScopedPermissionsTesting = []string{
-		// checking whether we have permissions resourcemanager.projects.list returns an error
-		// (as it is invalid when checking against a project resource),
-		// but listing projects is always allowed (it simply returns a list of projects you
-		// have projects.get permissions for)
-		"resourcemanager.projects.list",
-	}
-)
-
-const (
-	infrastructureConfigName = "cluster"
+	testablePerms = testablePermissions{}
 )
 
 var (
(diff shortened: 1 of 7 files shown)


joelddiaz:
/hold
we should discuss what this means for CredentialsRequests we don't control. A user can cause CCO to perform many, many API calls as it cycles through TestIamPermissions() to find 1 rejection, remove the rejected permission, and run TestIamPermissions() again, and again for each rejected permission.

pkg/operator/utils/gcp/utils.go
```diff
@@ -164,22 +123,57 @@ func checkServicesAndPermissions(gcpClient ccgcp.Client, permissionsList []strin
...
-func filterOutPermissions(permList, filterOutList []string) []string {
-	filteredPerms := []string{}
-	for _, perm := range permList {
-		filterOut := false
-		for _, filterPerm := range filterOutList {
-			if perm == filterPerm {
-				filterOut = true
+func refreshTestablePermissions(gcpClient ccgcp.Client, projectName string, logger log.FieldLogger) error {
+	// refresh if last update was over 1 hour ago
+	if testablePerms.lastUpdated.Add(time.Hour).Before(time.Now()) {
```
abhinavdahiya, line 128: early return maybe?

pkg/operator/utils/gcp/utils.go
```diff
@@ -164,22 +123,57 @@ func checkServicesAndPermissions(gcpClient ccgcp.Client, permissionsList []strin
...
+
+		ctx := context.TODO()
+		nextPageToken := ""
+		permList := []string{}
+
+		for {
+			projectNamePath := fmt.Sprintf(`//cloudresourcemanager.googleapis.com/projects/%s`, projectName)
```
abhinavdahiya, line 135: this can be outside the loop

pkg/operator/utils/gcp/utils.go
```diff
@@ -164,22 +123,57 @@ func checkServicesAndPermissions(gcpClient ccgcp.Client, permissionsList []strin
...
+			req := &iamadminpb.QueryTestablePermissionsRequest{
+				FullResourceName: projectNamePath,
+				PageToken:        nextPageToken,
+			}
```
abhinavdahiya, line 139: same let's create the req body once before loop, and only update the PageToken

pkg/operator/utils/gcp/utils.go
```diff
@@ -164,22 +123,57 @@ func checkServicesAndPermissions(gcpClient ccgcp.Client, permissionsList []strin
...
+			resp, err := gcpClient.QueryTestablePermissions(ctx, req)
+			if err != nil {
+				logger.WithError(err).Error("failed to gather list of testable permissions")
+				return err
+			}
+
+			for _, perm := range resp.Permissions {
+				permList = append(permList, perm.Name)
```
abhinavdahiya, line 147: just directly add to the new local copy of set we can generate.

pkg/operator/utils/gcp/utils.go
```diff
@@ -164,22 +123,57 @@ func checkServicesAndPermissions(gcpClient ccgcp.Client, permissionsList []strin
...
+	}
+
+	return nil
+}
+
+// filterOutPermissions will take a list of GCP permissions to test against and return a list of permissions that
+// are valid to test against at the project level
+func filterOutPermissions(gcpClient ccgcp.Client, projectName string, permList []string, logger log.FieldLogger) []string {
+
+	refreshTestablePermissions(gcpClient, projectName, logger)
```
abhinavdahiya, line 168: handle error.

joelddiaz: last second refactoring (－‸ლ)

joelddiaz:
@abhinavdahiya this should be good for another review

pkg/operator/utils/gcp/utils.go
```diff
@@ -183,6 +227,51 @@ func CheckPermissionsAgainstPermissionList(gcpClient ccgcp.Client, permList []st
 	return disallowedPerms.Len() == 0, nil
 }
 
+/*
```
abhinavdahiya, line 230: can we drop this commented code?

joelddiaz: Good catch. Done.

PR #332, opened 2021-04-30
Bug 1952891: Fix OpenStack clouds.yaml updating in secretannotator

mdbooth:
There were 2 bugs:

* https://bugzilla.redhat.com/show_bug.cgi?id=1952891 would be a serious regression affecting anybody upgrading from OCP 4.2.
* Additionally, we would never complete the reconcile loop if clouds.yaml contained a cacert.

Also adds logging to make it easier to tell when secretannotator has or hasn't updated clouds.yaml.


pkg/operator/secretannotator/openstack/reconciler_test.go
```diff
@@ -220,85 +224,65 @@ func TestReconcileCloudCredSecret_Reconcile(t *testing.T) {
...
-					require.NoError(t, err, "Unexpected error parsing original clouds.yaml")
-					origClouds.Clouds.OpenStack.CACert = tc.expectedCACert
+				if tc.wantErr {
+					require.Error(t, reconcileErr, "ReconcileCloudCredSecret.Reconcile() did not return expected error")
 
-					reconciledClouds, err := parseClouds(reconciledSecret)
-					require.NoError(t, err, "Unexpected error parsing updated clouds.yaml")
-					assert.Equal(t, origClouds, reconciledClouds, "Secret was not updated as expected")
+					// clouds.yaml should have been modified if there was an error
+					assert.Equal(t, tc.cloudsYAML, reconciledCloudsYAML)
```
joelddiaz, line 292: What is this assertion asserting precisely (maybe put some text in the final parameter to the Equal() call explaining what was expected)?

mdbooth: Eurgh! I noted that I hadn't done that yet when I wrote it, and then forgot 🤦‍♂️

joelddiaz (commented):
/lgtm

PR #334, opened 2021-05-03
Bug 1954865: add priorityClass to pod-identity Deployment

joelddiaz:
Since the pod-identity-webhook exists for user workloads, mark the
priority class as "system-cluster-critical" (not be preempted by user
workloads).


akhil-rane (commented):
/lgtm

PR #336, opened 2021-05-05
Bug 1957446: v1beta to v1 for the CredentialsRequest CRD

joelddiaz:
migrate to v1 CRDs in anticipation of kube 1.22 dropping support for
v1beta1.

Process used was to first generate v1 CRD into a temporary location:
`_output/tools/bin/controller-gen paths=./pkg/apis/...  crd:crdVersions=v1 output:crd:artifacts:config=tmpcrds`

Next copy the CRD to manifests/00-crd.yaml, and compare for any
unexpected changes.

Add back the `include.release.openshift.io` annotations.

Finally, copy that CRD to
`bindata/boostrap/cloudcredential_v1_credentialsrequest_crd.yaml`.

`make update` will now take any new API changes and update the generated
CRDs appropriately (keeping the v1 CRD).

Also remove broken `go generate` which wasn't actually creating the
deepcopy functions anymore.

xref: https://issues.redhat.com/browse/CCO-65

### DIFF
--- a/bindata/bootstrap/cloudcredential_v1_credentialsrequest_crd.yaml
+++ b/bindata/bootstrap/cloudcredential_v1_credentialsrequest_crd.yaml
@@ -1,6 +1,9 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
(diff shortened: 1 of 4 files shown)


joelddiaz:
/assign @dgoodwin @abhinavdahiya 
Looks like we want this in for 4.8 to smooth out the upgrade to 4.9 where v1beta1 really goes away.

PR #340, opened 2021-05-11
Bug 1958861: back to no CSR permissions for pod-id-webhook

joelddiaz:
the Pod is creating CSRs which don't get approved, and they pile up on
the cluster.

Return to not granting CSR permissions via the clusterrole. The Pod will
have errors about the CSR issues, but the webhook still works to
properly mutate Pods (to insert the env vars and add the projected
volume mounts).


dgoodwin:
what does it need those csrs for?

joelddiaz:
> what does it need those csrs for?

IIUC, this is for non-OpenShift environments where the service-signing-certs are not available. Since we have the service-singing-cert annotation on our Service (https://github.com/openshift/cloud-credential-operator/blob/master/bindata/v4.1.0/aws-pod-identity-webhook/svc.yaml#L10 ), then our Secret does get populated with TLS certificates, and it works itself out. So a failure to create a CSR isn't critical since other processes are handling the certificates for our pod-identity-webhook Service.

PR #341, opened 2021-05-17
Bug 1864116: Fix "unsupported platform type" log spamming on BareMetal.

dgoodwin:
We were treating our not having a Spec implementation the clusters
Infrastructure platform as an error, but this is normal in BareMetal
clusters which have no CredentialsRequests.

Stop treating as an error and just return that the CredRequest is for
another cloud, which is true, and continue.

/assign @akhil-rane 
/cc @joelddiaz


joelddiaz (commented):
/lgtm

PR #343, opened 2021-05-20
Bug 1962602: Remove log statement in frequently called function.

dgoodwin:
This function is called often when watching all Secrets in the cluster
to determine if we should re-reconcile if the cluster platform is none.
As such it should not be logging anything.


joelddiaz (commented):
/lgtm

PR #326, opened 2021-04-15
Documentation to add a new cloud provider

akhil-rane:
x-ref: https://issues.redhat.com/browse/CCO-82

PR #344, opened 2021-05-20
Add IBMCloud Passthrough

BobbyRadford:
Support the IBMCloud platform in Passthrough mode for IPI/UPI.

Related to: https://github.com/openshift/enhancements/pull/773


joelddiaz:
@dgoodwin are we still accepting new platform enablement of this style (where CCO is doing in-cluster processing of CredentialsRequests)?

dgoodwin:
@BobbyRadford per Joel's comment above, we are wondering if you would be willing to be an early adopter of our new approach to new platforms in CCO. 

We have some docs on how we want this to work going forward that will merge once the window opens for next release: https://github.com/openshift/cloud-credential-operator/pull/326

Essentially over the past few years of openshift 4 we have learned that customers prefer (a) fine grained least privileged for each component, and (b) the cluster not having admin level credentials. We're now moving away from mint and passthough mode (especially for new platforms) instead opting for manual mode, with ccoctl utility for admins to manually mint their credentials. 

For this PR we'd love to have IBM cloud be a fully supported best practice implementation within CCO which would mean:

  * Add support for specifying precise permissions each OpenShift component needs in IBM Cloud.
  * Drop passthrough support and instead just support manual.
  * Implement IBM Cloud minting in the ccoctl tool discussed in the above doc PR.

Lets us know what you think here, or find Joel or I on slack.

jeffnowicki:
@dgoodwin - thank you for the feedback and nudge to look at the new CCO approach. We are in 100% agreement on the preference for fine-grain component-level privilege control and removal of admin level credentials from the cluster.  OCP 4.9 will be the first release for [IBM Cloud platform support](https://github.com/openshift/enhancements/pull/773).  The timeline we have is extremely tight and we have significant amount of work to complete. Decisions have been made that would keep us on this timeline. One of those decisions (discussed with Katherine Dube and RH team) was to support “passthrough” only in our genesis effort. We would be in a better position to look at this support as an update on a 4.10 timeline.

joelddiaz:
FWIW, using the Manual mode approach doesn't mean you *must* use fine-grained permissions. Manual mode just means that there is no in-cluster handling of cloud credentials data.

It is entirely possible to just put the same credentials secret data into each generated Secret manifests when using Manual mode (which is precisely how things work when doing an install on AWS in an environment where the IAM API endpoints are not available, and the cloud-cred-operator cannot do it's normal mode of operation). But extending `ccoctl` for this kind of copy-the-same-secret-data-over-and-over-again operation does seem a bit pointless.

BobbyRadford:
Closing this in light of the new PR for Manual mode: #356

PR #346, opened 2021-05-23
Add permissions boundary support to ccoctl when creating AWS IAM Roles

bellpr:


joelddiaz (commented):
This looks fine to me.
@dgoodwin I think the only question I have is whether we want to expose this functionality for the `create-all` subcommand.

dgoodwin:
It's on AWS specific sub-commands, it would be weird if you could only use it if you did step by step and not create-all. Feels like create-all might need some command flag inheritance someday, but as is it looks ok to me. Any reason we'd not want it on create-all I'm not aware of?

akhil-rane:
looks okay to me to have this subcommand on `create-all`

pkg/cmd/provisioning/aws/create-iam-roles.go
```diff
@@ -31,6 +31,7 @@ const (
 	roleTemplate = `{
 	"RoleName": "%s",
 	"AssumeRolePolicyDocument": "%s",
+	"PermissionsBoundary": "%s",
```
joelddiaz, line 34: The template needs to be conditional to account for someone not using/specifying a permissions boundary. The generated JSON files when not providing a permissions boundary fail when trying to apply them.

```
./ccoctl aws create-iam-roles --credentials-requests-dir ~/crtemp/ --dry-run --output-dir ./tdir/ --identity-provider-arn arn:aws:iam::123456789012:oidc-provider/jdiaz-pbound-oidc.s3.us-east-1.amazonaws.com --name jdiaz-pbound

aws iam create-role --cli-input-json file://tdir/05-0-jdiaz-pbound-openshift-machine-api-aws-cloud-credentials-role.json

Parameter validation failed:
Invalid length for parameter PermissionsBoundary, value: 0, valid min length: 20
```

joelddiaz (changes requested):

pkg/cmd/provisioning/aws/create-iam-roles.go
bellpr: @joelddiaz Done, and added a comment to ensure the use case for the JSON files is clear.

PR #349, opened 2021-06-02
Bug 1960176: Make sure credentials have newly generated azure client secret

akhil-rane:
x-ref: https://issues.redhat.com/browse/OCPBUGSM-29141

### DIFF
--- a/pkg/azure/actuator.go
+++ b/pkg/azure/actuator.go
@@ -504,8 +504,12 @@ func (a *Actuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequest
 		return err
 	}
 
-	if clientSecret == "" {
+	if newClientSecret != "" {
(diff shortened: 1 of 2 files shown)


pkg/azure/actuator.go
```diff
@@ -504,7 +504,7 @@ func (a *Actuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequest
 		return err
 	}
 
-	if clientSecret == "" {
+	if newClientSecret != "" {
```
joelddiaz, line 507: Took me a while to follow the logic here, but this looks like it would work. :+1: 

Should we error after this if block if clientSecret happens to still be "" ? (just out an abundance of caution even though the current code flow should never allow for it)

akhil-rane: yes, sounds good!

PR #351, opened 2021-06-07
Bug 1967531: Fix ccoctl delete not removing iam roles

akhil-rane:
iam.ListRoles results are paginated to 100 items by default. If
an account had more then 100 roles, delete command did not work in
some cases. This commit fixes it by checking if iam.ListRoles
result is truncated, fetching next batch of results and deleting
the roles.

x-ref: https://issues.redhat.com/browse/OCPBUGSM-29969


joelddiaz (commented):
/lgtm

PR #353, opened 2021-06-10
Bug 1970062: delete client should also use shared session function

joelddiaz:
this fixes delete client not reading from ~/.aws/config


akhil-rane (commented):
/lgtm

PR #356, opened 2021-06-15
Add IBMCloud manual mode

BobbyRadford:
Support the IBMCloud platform in Manual mode for IPI/UPI.

Related to #344 and https://github.com/openshift/enhancements/pull/773


pkg/apis/cloudcredential/v1/types_ibmcloud.go
```diff
@@ -0,0 +1,29 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TODO: these types should eventually be broken out, along with the actuator, to a separate repo.
+
+// IBMCloudProviderSpec the specification of the credentials request in IBM Cloud.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type IBMCloudProviderSpec struct {
+	metav1.TypeMeta `json:",inline"`
+}
```
dgoodwin, line 29: Lets get a Status in here even if it's empty now.

pkg/apis/cloudcredential/v1/register.go
```diff
@@ -57,6 +57,7 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 		&AWSProviderStatus{}, &AWSProviderSpec{},
 		&AzureProviderStatus{}, &AzureProviderSpec{},
 		&GCPProviderStatus{}, &GCPProviderSpec{},
+		&IBMCloudProviderSpec{},
```
dgoodwin, line 60: Status here as well.

pkg/apis/cloudcredential/v1/types_ibmcloud.go
```diff
@@ -0,0 +1,29 @@
...
+
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TODO: these types should eventually be broken out, along with the actuator, to a separate repo.
+
+// IBMCloudProviderSpec the specification of the credentials request in IBM Cloud.
```
dgoodwin, line 25:
```suggestion
// IBMCloudProviderSpec is the specification of the credentials request in IBM Cloud.
```

dgoodwin (commented):
Looks pretty good to me for a first cut until we can get defined permissions in there. Just a couple minor requests.

BobbyRadford:
@dgoodwin thanks for the review. I pushed up a couple of commits to fix things up.

pkg/cmd/provisioning/ibmcloud/create.go
```diff
@@ -0,0 +1,177 @@
...
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+
+	credreqv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+	"github.com/openshift/cloud-credential-operator/pkg/cmd/provisioning"
```
akhil-rane, line 12: To keep consistency, these imports should come in a separate block at the end

pkg/cmd/provisioning/ibmcloud/create.go
```diff
@@ -0,0 +1,177 @@
...
+	// objects.
+	CreateOpts = options{
+		TargetDir: "",
+	}
+)
+
+// NewCreateCmd implements the "create" command for the credentials provisioning
+func NewCreateCmd() *cobra.Command {
+	createCmd := &cobra.Command{
+		Use:              "create",
```
akhil-rane, line 42: by command conventions that we follow, this should be `create-secrets` I feel. @dgoodwin WDYT?

pkg/cmd/provisioning/ibmcloud/create_test.go
```diff
@@ -0,0 +1,103 @@
...
+			verify: func(t *testing.T, targetDir string) {
+				files, err := ioutil.ReadDir(targetDir)
+				require.NoError(t, err, "Unexpected error listing files in targetDir")
+
+				assert.Equal(t, 1, len(files), "Should be exactly 1 Secret generated for 1 CredentialsRequest")
+			},
+			cleanup: func(t *testing.T) {
+				return
+			},
+			expectError: false,
```
akhil-rane, line 48: Can we have a test that does setting IC_API_KEY in setup and verify if generated secret actually has that key? Also, one negative test where env var is not set and error is thrown.

akhil-rane (commented):
few suggestions

pkg/cmd/provisioning/ibmcloud/create.go
BobbyRadford: That's no problem. Changed to `create-secrets`

pkg/cmd/provisioning/ibmcloud/create.go
BobbyRadford: Done. Hopefully I understood you correctly.

akhil-rane (commented):
/lgtm

BobbyRadford:
Ran into this on a very recent Installer PR, but seems like AWS e2e tests are failing due to some throttling issues. We ended up getting GCP tests to run instead. Is that something we can do here to progress this PR? @akhil-rane @dgoodwin https://github.com/openshift/installer/pull/4923#issuecomment-862500285

pkg/cmd/provisioning/ibmcloud/create_secrets.go
```diff
@@ -0,0 +1,182 @@
...
+	// objects.
+	CreateOpts = options{
+		TargetDir: "",
+	}
+)
+
+// NewCreateSecretsCmd implements the "create-secrets" command for the credentials provisioning
+func NewCreateSecretsCmd() *cobra.Command {
+	createSecretsCmd := &cobra.Command{
+		Use:              "create-secrets",
```
joelddiaz, line 46: @BobbyRadford I was discussing this with the team and I was wondering if a name like `create-shared-secrets` would be better.
I'm comparing this to the existing aws support and there is no equivalent `create-secrets` subcommand (today) for AWS (which isn't surprising b/c of lack of support for passthrough-equivalent mode). So this subcommand feels specific to implementing Passthrough mode and the name of the command should reflect that.

Could we rename this to something like `create-shared-secrets` and have the long description make it clear that this subcommand just takes the API key from an env var and uses it to satisfy all the CredentialsRequests?

BobbyRadford: Hey @joelddiaz no problem. Will make that happen.

PR #358, opened 2021-06-22
Add code coverage script & make target

2uasimojo:
This commit adds
- a `coverage` make rule that runs
- a new codecov.sh script adapted from [the one OSD operators use](https://github.com/openshift/boilerplate/blob/f3c8c8d393f69f83ce9dbdbc0aa72d2e5b9e5d96/boilerplate/openshift/golang-osd-operator/codecov.sh);
and
- a codecov configuration file, stolen directly from [the one OSD operators use](https://github.com/openshift/boilerplate/blob/f3c8c8d393f69f83ce9dbdbc0aa72d2e5b9e5d96/boilerplate/openshift/golang-osd-operator/.codecov.yml).

This is designed to be run in three modes: local, presubmit, and postsubmit.

- In all modes, it runs the go cover tool, scrubs out the stats for generated files, and writes the raw report to `coverage.out`.
- In local mode, that's it.
- Presubmit and postsubmit modes are intended for use in CI jobs, where they save codecov.io's reporting tool as a build artifact, and run it.
  - Presubmit mode writes a hypothetical ("if this were to merge now") coverage report to the PR.
  - Postsubmit mode saves the coverage report to codecov.io where it becomes permanently available in the historical record for this repository.

### DIFF
--- a/hack/codecov.sh
+++ b/hack/codecov.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+CI_SERVER_URL=https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test
+COVER_PROFILE=${COVER_PROFILE:-coverage.out}
+JOB_TYPE=${JOB_TYPE:-"local"}
+
+# Default concurrency to four threads. By default it's the number of procs,
+# which seems to be 16 in the CI env. Some consumers' coverage jobs were
+# regularly getting OOM-killed; so do this rather than boost the pod resources
+# unreasonably.
+COV_THREAD_COUNT=${COV_THREAD_COUNT:-4}
+make -C "${REPO_ROOT}" test-unit GO_TEST_FLAGS="-coverprofile=${COVER_PROFILE}.tmp -covermode=atomic -coverpkg=./... -p ${COV_THREAD_COUNT}"
+
+# Remove generated files from coverage profile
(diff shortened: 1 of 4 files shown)


hack/codecov.sh
```diff
@@ -0,0 +1,54 @@
...
+export CI_BUILD_URL="${JOB_LINK}"
+export CI_BUILD_ID="${JOB_NAME}"
+export CI_JOB_ID="${BUILD_ID}"
+
+if [[ "${JOB_TYPE}" != "local" ]]; then
+       if [[ -z "${ARTIFACT_DIR:-}" ]] || [[ ! -d "${ARTIFACT_DIR}" ]] || [[ ! -w "${ARTIFACT_DIR}" ]]; then
+              echo '${ARTIFACT_DIR} must be set for non-local jobs, and must point to a writable directory' >&2
+              exit 1
+       fi
+       curl -sS https://codecov.io/bash -o "${ARTIFACT_DIR}/codecov.sh"
```
joelddiaz, line 50: Are we cool with dynamically pulling in code to execute off the interwebs? (I ask this as someone who hasn't dealt with codecov previously...)

2uasimojo: Totally legit question. There was a breach just a couple of months ago on this. Our [SOP has TODOs](https://github.com/openshift/hive-sops/blob/master/sop/Coverage.md#todo) to make it moar secure.

A couple of things came out of that incident, though, that make us reasonably confident that this specific setup is okay. The main thing is the fact that we're running in an isolated container in CI, so the attack surface is limited to that container. The worst thing that happened in the incident was that the codecov tokens themselves were leaked, meaning the worst thing the attackers could have done was to upload bogus coverage reports.

We rotated the tokens; and also started saving the script as a build artifact for auditing.

PR #359, opened 2021-06-29
Support gcp workload identity federation

akhil-rane:
To enable short-lived credentials in gcp cluster using workload identity
federation, we need to make following set of changes

- update golang.org/x/oauth2 to a version that supports workload identity
- add kubernetes service account names to gcp credentials request manifest
  which can be used by ccoctl tool to create gcp service accounts with 
  limited access

ref: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity

x-ref: https://issues.redhat.com/browse/CCO-105


akhil-rane:
@joelddiaz this looks good for a quick review

manifests/05-gcp-ro-credentialsrequest.yaml
```diff
@@ -17,3 +17,6 @@ spec:
   secretRef:
     name: cloud-credential-operator-gcp-ro-creds
     namespace: openshift-cloud-credential-operator
+  serviceAccountNames:
+  - cloud-credential-operator
```
joelddiaz, line 21: I'm assuming this is more of an example of the kinds of changes that need to be made across the various repos, and CCO - as presently written - will not actually make use of these creds when in Manual mode. Correct?

akhil-rane: thats right

PR #365, opened 2021-07-26
Bug 1990975: Enhance mechanism of reading ibm cloud apikey

mkumatag:
Have seen users to use following set of environments to pass the APIKEY, so this PR is to support all those possible environment varaibles.

```
"IC_API_KEY", "IBMCLOUD_API_KEY", "BM_API_KEY", "BLUEMIX_API_KEY"
```

### DIFF
--- a/pkg/cmd/provisioning/ibmcloud/create_shared_secrets.go
+++ b/pkg/cmd/provisioning/ibmcloud/create_shared_secrets.go
@@ -29,8 +29,8 @@ type: Opaque`
 	manifestsDirName = "manifests"
 )
 
-// APIKeyEnvVar is the environment variable name containing an IBM Cloud API key
-const APIKeyEnvVar = "IC_API_KEY"
+// APIKeyEnvVars is a list of environment variable names containing an IBM Cloud API key
+var APIKeyEnvVars = []string{"IC_API_KEY", "IBMCLOUD_API_KEY", "BM_API_KEY", "BLUEMIX_API_KEY"}
 
 var (
 	// CreateOpts captures the options that affect creation of the generated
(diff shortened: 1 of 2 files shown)


pkg/cmd/provisioning/ibmcloud/create_shared_secrets.go
```diff
@@ -29,8 +29,8 @@ type: Opaque`
 	manifestsDirName = "manifests"
 )
 
-// APIKeyEnvVar is the environment variable name containing an IBM Cloud API key
-const APIKeyEnvVar = "IC_API_KEY"
+// APIKeyEnvVars is the environment variable names containing an IBM Cloud API key
```
akhil-rane, line 32: // APIKeyEnvVars is a list of environment variable names containing an IBM Cloud API key

pkg/cmd/provisioning/ibmcloud/create_shared_secrets.go
```diff
@@ -29,8 +29,8 @@ type: Opaque`
...
+var APIKeyEnvVars = []string{"IC_API_KEY", "IBMCLOUD_API_KEY", "BM_API_KEY", "BLUEMIX_API_KEY"}
```
akhil-rane, line 33: Is there any kind of precedence criteria here? How ibm cloud cli handles it when all of these environment variables are set?

mkumatag: go-client follows a same approach for reading the token here - 
https://github.com/IBM-Cloud/bluemix-go/blob/56edc36e39e52b243c4fb23a2cc63d8296d7d796/session/session.go#L41

pkg/cmd/provisioning/ibmcloud/create_shared_secrets.go
mkumatag: fixed it.

akhil-rane (commented):
/lgtm

pkg/cmd/provisioning/ibmcloud/create_shared_secrets.go
clnperez: @mkumatag lmk what you come up with here. i'm guessing we should be consistent with the installer.

mkumatag: if it is missing, we can enhance the code in the installer as well,

mkumatag:
@akhil-rane @joelddiaz it says missing bz label, not sure if there is a way to get rid of this condition!

akhil-rane:
> @akhil-rane @joelddiaz it says missing bz label, not sure if there is a way to get rid of this condition!

I am creating a BZ and linking this PR

PR #367, opened 2021-07-27
Add ccoctl create key pair command for google cloud

akhil-rane:
- [x] Add command to create keypair for google cloud
- [x] Refactor key pair generation code and make it common for all cloud providers
- [x] Create a common constants file for pkg/cmd/provisioning package

x-ref: https://issues.redhat.com/browse/CCO-118


cmd/ccoctl/main.go
```diff
@@ -17,6 +18,7 @@ func main() {
 
 	rootCmd.AddCommand(aws.NewAWSCmd())
 	rootCmd.AddCommand(ibmcloud.NewIBMCloudCmd())
+	rootCmd.AddCommand(gcp.NewGCPCmd())
```
joelddiaz, line 21: alphabetize
```suggestion
	rootCmd.AddCommand(gcp.NewGCPCmd())
	rootCmd.AddCommand(ibmcloud.NewIBMCloudCmd())
```

pkg/cmd/provisioning/aws/constants.go
```diff
@@ -1,33 +0,0 @@
-package aws
```
joelddiaz, line 1: for git history purposes, can this file show up as a rename to pkg/cmd/provisioning/constants.go ?

joelddiaz (commented):
:+1: 
minor nits

cmd/ccoctl/main.go
akhil-rane: fixed

pkg/cmd/provisioning/aws/constants.go
akhil-rane: I tried but as there are changes in that file git does not show it as rename

joelddiaz: wierd b/c it seems to have worked for pkg/cmd/provisioning/aws/create_key_pair.go → pkg/cmd/provisioning/create_key_pair.go , oh well

pkg/cmd/provisioning/gcp/gcp.go
```diff
@@ -0,0 +1,29 @@
...
+	Region         string
+	Name           string
+	CredRequestDir string
+	DryRun         bool
+}
+
+// NewGCPCmd implements the "gcloud" subcommand for the credentials provisioning
+func NewGCPCmd() *cobra.Command {
+	createCmd := &cobra.Command{
+		Use:   "gcloud",
```
joelddiaz, line 21: So are we finally going with `gcp` as the subcommand? (cc @dgoodwin )

pkg/cmd/provisioning/aws/constants.go
akhil-rane: true, no clue what is happening for constants.go

akhil-rane:
@joelddiaz this good for another round of review. Changed sub-command name from gcloud to gcp

joelddiaz (commented):
/lgtm

pkg/cmd/provisioning/constants.go
```diff
@@ -0,0 +1,33 @@
...
+	// NameTagKey is the key of the "Name" tag applied to the AWS resources created by ccoctl
+	NameTagKey = "Name"
+	// Generated identity provider files
+	OidcBucketFilename          = "01-oidc-bucket.json"
+	OidcConfigurationFilename   = "02-openid-configuration"
+	OidcKeysFilename            = "03-keys.json"
+	IAMIdentityProviderFilename = "04-iam-identity-provider.json"
+	// Generated role files
+	RoleFilenameFormat       = "05-%d-%s-role.json"
+	RolePolicyFilenameFormat = "06-%d-%s-policy.json"
```
abhinavdahiya, line 32: nit: no reason to keep aws specific constants in a global package. keep them private and close to the usage in aws implementation.

akhil-rane: These are the constants which are shared among different files in cloud provider packages. There are two options here
1) to create one constants file for entire provisioning package
2) to create separate constants file for each cloud provider

I prefer option (1) as it is easy to manage

joelddiaz: if these are truly global constants, then the var name `CcoctlAWSResourceTagKeyPrefix` shouldn't say `AWS` in the var name. Same for `OwnedCcoctlAWSResourceTagValue`.
And the comments for those two vars and `NameTagKey` shouldn't mention AWS.

akhil-rane: `CcoctlAWSResourceTagKeyPrefix` and `OwnedCcoctlAWSResourceTagValue` are constants specific to aws , but they are shared between different files in provisioning/aws package. I think I should move them somewhere inside the package.

akhil-rane:
addressed PR feedback. PTAL

joelddiaz (commented):
:+1: i'll let @abhinavdahiya give a look since he also had some feedback

abhinavdahiya:
this is looking great! 
/lgtm

PR #368, opened 2021-07-29
Add ccoctl command to create workload identity pool in GCP

akhil-rane:
- [x] vendor google.golang.org/api/iam/v1, this has sdk support for workload identity
- [x] Fix tests in pkg/operator/utils/gcp/utils_test.go due gomock version update
- [x] Move common code in pkg/cmd/provisioning package to utils.go
- [x] Add ccoctl gcp command to create workload identity pool

x-ref: https://issues.redhat.com/browse/CCO-119


akhil-rane:
/hold

wait for #367

pkg/operator/utils/gcp/utils_test.go
```diff
@@ -139,7 +139,7 @@ func TestCheckPermissionsAgainstPermissionListChunking(t *testing.T) {
 
 				chunkCopy := chunk
 
-				mockCall := mockClient.EXPECT().TestIamPermissions("fake", gomock.Any()).DoAndReturn(
+				mockCall := mockClient.EXPECT().TestIamPermissions("fake", gomock.Any()).AnyTimes().DoAndReturn(
```
joelddiaz, line 142: Should these really be `AnyTimes()` or could they be `Times(1)` (just to keep the test case expectations tighter)?

pkg/gcp/client.go
```diff
@@ -149,6 +152,10 @@ func (c *gcpClient) QueryTestablePermissions(ctx context.Context, request *iamad
 	return c.iamClient.QueryTestablePermissions(ctx, request)
 }
 
+func (c *gcpClient) CreateWorkloadIdentityPool(ctx context.Context, parent, poolID string, pool *iam.WorkloadIdentityPool) (*iam.Operation, error) {
+	return c.iamService.Projects.Locations.WorkloadIdentityPools.Create(parent, pool).WorkloadIdentityPoolId(poolID).Context(ctx).Do()
```
joelddiaz, line 156: lets wrap this with a timeout like the other wrapper functions do
```suggestion
	ctx, cancel := contextWithTimeout(ctx)
	defer cancel()
	return c.iamService.Projects.Locations.WorkloadIdentityPools.Create(parent, pool).WorkloadIdentityPoolId(poolID).Context(ctx).Do()
```

pkg/cmd/provisioning/gcp/create_workload_identity_pool.go
```diff
@@ -0,0 +1,75 @@
...
+
+// NewCreateWorkloadIdentityPool provides the "create-workload-identity-pool" subcommand
+func NewCreateWorkloadIdentityPool() *cobra.Command {
+	createIAMRolesCmd := &cobra.Command{
+		Use:   "create-workload-identity-pool",
+		Short: "Create workload identity pool",
+		Run:   createWorkloadIdentityPoolCmd,
+	}
+
+	createIAMRolesCmd.PersistentFlags().StringVar(&CreateWorkloadIdentityPoolOpts.Name, "name", "", "User-define name for all created Google cloud resources (can be separate from the cluster's infra-id)")
```
joelddiaz, line 69:
```suggestion
	createIAMRolesCmd.PersistentFlags().StringVar(&CreateWorkloadIdentityPoolOpts.Name, "name", "", "User-defined name for all created Google cloud resources (can be separate from the cluster's infra-id)")
```

pkg/cmd/provisioning/gcp/create_workload_identity_pool.go
```diff
@@ -0,0 +1,75 @@
...
+		Project: "",
+	}
+)
+
+func createWorkloadIdentityPoolCmd(cmd *cobra.Command, args []string) {
+	ctx := context.Background()
+
+	gcpClient, err := gcp.NewClient(CreateWorkloadIdentityPoolOpts.Project, nil)
+	if err != nil {
+		log.Fatal(err)
```
joelddiaz, line 27:
```suggestion
		log.Fatalf("Failed to setup GCP client: %s", err)
```

pkg/cmd/provisioning/gcp/create_workload_identity_pool.go
```diff
@@ -0,0 +1,75 @@
...
+
+	pool := &iam.WorkloadIdentityPool{
+		Name:        name,
+		DisplayName: shortenedName,
+		Description: "A workload identity pool for an OpenShift cluster",
+		State:       "ACTIVE",
+		Disabled:    false,
+	}
+
+	_, err := client.CreateWorkloadIdentityPool(ctx, fmt.Sprintf("projects/%s/locations/global", project), shortenedName, pool)
```
joelddiaz, line 53: can we make a simple string template to store the magic "projects/%s/locations/global" string (with a hopefully descriptive variable name)?

pkg/cmd/provisioning/gcp/create_workload_identity_pool_test.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+		})
+	}
+}
+
+func mockCreateWorkloadIdentityPoolSuccess(mockGCPClient *mockgcp.MockClient) {
+	mockGCPClient.EXPECT().CreateWorkloadIdentityPool(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(
+		&iam.Operation{
+			Done: true,
+		}, nil).AnyTimes()
```
joelddiaz, line 69: Can we tighten this expect to be more strict than `AnyTimes()`?

pkg/cmd/provisioning/gcp/gcp.go
```diff
@@ -0,0 +1,31 @@
...
+	Name           string
+	CredRequestDir string
+	DryRun         bool
+	Project        string
+}
+
+// NewGCPCmd implements the "gcloud" subcommand for the credentials provisioning
+func NewGCPCmd() *cobra.Command {
+	createCmd := &cobra.Command{
+		Use:   "gcloud",
```
joelddiaz, line 22: I don't have a strong opinion here, but I expected it to be `gcp` not `gcloud`. What do others think?

dgoodwin: +1 for gcp

pkg/gcp/client.go
```diff
@@ -198,9 +205,16 @@ func fixupServiceMap(serviceMap map[string]bool) {
...
 func NewClient(projectName string, authJSON []byte) (Client, error) {
 	ctx := context.TODO()
-
+	var creds *google.Credentials
+	var err error
 	// since we're using a single creds var, we should specify all the required scopes when initializing
-	creds, err := google.CredentialsFromJSON(context.TODO(), authJSON, "https://www.googleapis.com/auth/cloud-platform")
+	scope := "https://www.googleapis.com/auth/cloud-platform"
+
+	if authJSON != nil {
```
abhinavdahiya, line 213: check for len of slices instead of nil value.

pkg/operator/utils/gcp/utils_test.go
akhil-rane: This was helpful. It caught some failing tests. This might need more careful review.

pkg/cmd/provisioning/gcp/gcp.go
akhil-rane: gcp it is

pkg/operator/utils/gcp/utils_test.go
```diff
@@ -171,39 +182,20 @@ func TestIgnoreInvalidProjectPermissions(t *testing.T) {
...
-						return nil, &googleapi.Error{
-							Code:    400,
-							Message: "Permission invalidPerm is not valid for this resource.",
-						}
-					})
-
-				// now allow the perm list w/o invalidPerm
-				secondMock := mockClient.EXPECT().TestIamPermissions(projectName, gomock.Any()).DoAndReturn(
+				// allow the perm list w/o invalidPerm
+				mockCall := mockClient.EXPECT().TestIamPermissions(projectName, gomock.Any()).Times(1).DoAndReturn(
```
akhil-rane, line 196: There is only one call to `TestIamPermissions` inside `CheckPermissionsAgainstPermissionList`. It takes all permissions and returns valid permissions. I do not think other mocks are required here.

pkg/cmd/provisioning/gcp/create_workload_identity_pool.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+func createWorkloadIdentityPool(ctx context.Context, client gcp.Client, name, project, targetDir string, generateOnly bool) error {
+	if generateOnly {
+		createWorkloadIdentityPoolScript := provisioning.CreateShellScript([]string{
+			fmt.Sprintf(createIdentityPoolCmd, name, name),
+		})
+		createIdentityPoolScripFullPath := filepath.Join(targetDir, provisioning.CreateIdentityPoolScriptName)
+		if err := provisioning.SaveToFile(fmt.Sprintf("script to create workload identity pool %s", name), createIdentityPoolScripFullPath, []byte(createWorkloadIdentityPoolScript)); err != nil {
```
joelddiaz, line 78:
```suggestion
		createIdentityPoolScriptFullPath := filepath.Join(targetDir, provisioning.CreateIdentityPoolScriptName)
		if err := provisioning.SaveToFile(fmt.Sprintf("script to create workload identity pool %s", name), createIdentityPoolScriptFullPath, []byte(createWorkloadIdentityPoolScript)); err != nil {
```

pkg/cmd/provisioning/gcp/create_workload_identity_pool_test.go
```diff
@@ -0,0 +1,130 @@
+package gcp
+
+import (
+	"context"
+	"github.com/openshift/cloud-credential-operator/pkg/cmd/provisioning"
```
joelddiaz, line 5: move this import down with the mockgcp import

pkg/cmd/provisioning/utils.go
```diff
@@ -23,3 +27,34 @@ func EnsureDir(path string) error {
...
+	_, err = f.Write(data)
+	f.Close()
+	if err != nil {
+		return errors.Wrapf(err, "failed to write %s to file %s", filePurpose, filePath)
+	}
+	return nil
+}
+
+// CountNonDirectoryFiles counts files which are not a directory
+func CountNonDirectoryFiles(files []os.FileInfo) int {
```
joelddiaz, line 52: Is this used outside of the test files?

pkg/cmd/provisioning/utils.go
```diff
@@ -23,3 +27,34 @@ func EnsureDir(path string) error {
...
 	return nil
 }
+
+// CreateShellScript creates a shell script given commands to execute
+func CreateShellScript(commands []string) string {
+	return fmt.Sprintf("#!/bin/sh\n%s", strings.Join(commands, "\n"))
+}
+
+// SaveToFile saves the given data to a given file
+func SaveToFile(filePurpose, filePath string, data []byte) error {
```
joelddiaz, line 37: Any particular reason to not simply use ioutil.WriteFIle()?

pkg/operator/utils/gcp/utils_test.go
```diff
@@ -107,6 +107,14 @@ func perms(low, hi int) []string {
 	return result
 }
 
+func createStringSet(stringArray []string) sets.String {
+	stringSet := sets.NewString()
+	for _, str := range stringArray {
+		stringSet.Insert(str)
+	}
+	return stringSet
```
joelddiaz, line 115: I think you can pass in your initial list of strings into the NewString() call.
```suggestion
        return sets.NewString(stringArray...)
```

joelddiaz (commented):
One overarching comment on saving the gcloud commands to a script is that this really makes our approach a Mac/Linux solution (unless we consider WSL acceptable for Windows users). 
We don't presently produce ccoctl for anything except Linux today (due to where in the build pipeline this repo is processed), but it has been discussed previously to build ccoctl for the same target set as the installer is built for.

Thoughts?

cmd/ccoctl/main.go
```diff
@@ -19,6 +19,7 @@ func main() {
 	rootCmd.AddCommand(aws.NewAWSCmd())
 	rootCmd.AddCommand(gcp.NewGCPCmd())
 	rootCmd.AddCommand(ibmcloud.NewIBMCloudCmd())
+	rootCmd.AddCommand(gcp.NewGCPCmd())
```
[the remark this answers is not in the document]
akhil-rane: yes, will fix it

pkg/cmd/provisioning/utils.go
akhil-rane: to be honest I just moved it to utils.go since it was a common code. Let me explore if we can use ioutil.WriteFile() here.

pkg/cmd/provisioning/utils.go
joelddiaz: that's fair. let each test file at least share this functionality :+1:

akhil-rane: fixed

pkg/cmd/provisioning/utils.go
akhil-rane: done

akhil-rane:
@joelddiaz this is ready for another round of review

/assign @jeana-redhat @lwan-wanglin @sferich888

pkg/operator/utils/gcp/utils_test.go
joelddiaz: Is this being missed because of a missing `defer Finish()` on the generated mockcontroller down by line 255?

akhil-rane: I am not really sure why we are expecting 2 calls to `TestIamPermissions` here. From what I understand (from the code snippet below of `CheckPermissionsAgainstPermissionList` function), we should have 1 call to `TestIamPermissions` for 100 permissions (i.e. chunksize) and since we have only 2 permissions there should be only 1 call. 

```
	chunkSize := 100
	permLen := len(filteredPermList)
	for i := 0; i < permLen; i += chunkSize {
		end := i + chunkSize
		if end > permLen {
			end = permLen
		}
		req := &cloudresourcemanager.TestIamPermissionsRequest{Permissions: filteredPermList[i:end]}
		resp, err := gcpClient.TestIamPermissions(projectName, req)
		if err != nil {
			return false, fmt.Errorf("error testing permissions: %v", err)
		}
		allowedPerms.Insert(resp.Permissions...)
	}
```

Correct me if I am missing something here :(

pkg/cmd/provisioning/constants.go
```diff
@@ -13,4 +13,6 @@ const (
 	ManifestsDirName = "manifests"
 	// TLSDirName is the name of the directory to save bound service account signing key created by ccoctl
 	TLSDirName = "tls"
+	// FileModeCcoctlDryRun represents a mode and permission bits of the files created by ccoctl in dry run
+	FileModeCcoctlDryRun = 0744
```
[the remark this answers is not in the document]
akhil-rane: you are right! I will keep separate constants for aws and gcp

pkg/operator/utils/gcp/utils_test.go
joelddiaz: Maybe I'm missing something in this discussion. I only see one EXPECT() for TestIamPermissions()...

lwan-wanglin:
@akhil-rane  here is the test case for the PR https://url.corp.redhat.com/9aab015, could you help to review?

akhil-rane:
> @akhil-rane here is the test case for the PR https://url.corp.redhat.com/9aab015, could you help to review?

@lwan-wanglin I think there should be one more step for `--dry-run`. After checking the contents of the shell script (step 6), we should run the shell script and check if workload identity pool is getting created in the gcp. wdyt?

pkg/operator/utils/gcp/utils_test.go
joelddiaz: Ok, we got this figured out after talking.

pkg/cmd/provisioning/constants.go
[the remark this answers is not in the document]
akhil-rane: @joelddiaz this is fixed.

joelddiaz (commented):
/lgtm

joelddiaz:
/assign @jeana-redhat @lwan-wanglin @sferich888 
As far as docs (and px), there is a standalone PR in flight with the documentation that should cover this whole series of PRs adding support for GCP workload-identity in #376

lwan-wanglin:
> I think there should be one more step for `--dry-run`. After checking the contents of the shell script (step 6), we should run the shell script and check if workload identity pool is getting created in the gcp. wdyt?

@akhil-rane thanks for your advice, I have updated the test case and added steps for running shell script, please check again https://url.corp.redhat.com/9aab015 .

And I have a small question,  openshift-install binary can read gcp credentials from /home/<user>/.gcp/osServiceAccount.json by default, is it possible for ccoctl to do so?  now ccoctl only respect GOOGLE_APPLICATION_CREDENTIALS env variable.

akhil-rane:
> > I think there should be one more step for `--dry-run`. After checking the contents of the shell script (step 6), we should run the shell script and check if workload identity pool is getting created in the gcp. wdyt?
> 
> @akhil-rane thanks for your advice, I have updated the test case and added steps for running shell script, please check again https://url.corp.redhat.com/9aab015 .
> 
> And I have a small question, openshift-install binary can read gcp credentials from /home//.gcp/osServiceAccount.json by default, is it possible for ccoctl to do so? now ccoctl only respect GOOGLE_APPLICATION_CREDENTIALS env variable.

interesting find about default location for credentials. I think we should include it. But it looks unrelated in this PR. Lets open a jira issue and handle it in a separate PR. Does that sound good to you?

lwan-wanglin:
soulds ok to me.
/label qe-approved

PR #370, opened 2021-08-07
Replace stringData with data

mkumatag:
This change is just to keep sync with other providers secret, all are feeding in `data` instead of `stringdata`

### DIFF
--- a/pkg/cmd/provisioning/ibmcloud/create_shared_secrets.go
+++ b/pkg/cmd/provisioning/ibmcloud/create_shared_secrets.go
@@ -1,6 +1,7 @@
 package ibmcloud
 
 import (
+	b64 "encoding/base64"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


joelddiaz (commented):
This looks fine, but I'm curious as to the rationale here. AFAICT, the AWS codepath that creates Secrets uses `stringData` https://github.com/openshift/cloud-credential-operator/blob/master/pkg/cmd/provisioning/aws/create-iam-roles.go#L29-L40 .

What provider is this PR attempting to mimic?

mkumatag:
> This looks fine, but I'm curious as to the rationale here. AFAICT, the AWS codepath that creates Secrets uses `stringData` https://github.com/openshift/cloud-credential-operator/blob/master/pkg/cmd/provisioning/aws/create-iam-roles.go#L29-L40 .
> 
> What provider is this PR attempting to mimic?

Ah.. referring to the specs mentioned in the readme file..

mkumatag:
> @mkumatag do you want to see this change in 4.9? Does not look like a super urgent to create BZ

Can wait for next release

mkumatag:
This is not needed anymore because the entire layout of the secret is getting changed via https://github.com/openshift/cloud-credential-operator/pull/374, if needed will accommodate the changes in the same PR, hence closing this PR as of now. -Thanks.

PR #371, opened 2021-08-16
Add ccoctl gcp command to create service accounts

akhil-rane:
- [x] Add command to create gcp service accounts along with policy bindings for workload identity to work
- [x] Add code for adding policy bindings to service account. We currently have code to add bindings to project.
- [x] Move common code to fetch credentials request manifests from aws package to utils.go

x-ref: https://issues.redhat.com/browse/CCO-122

PR #372, opened 2021-08-17
Add ccoctl gcp command to delete credentials infra

akhil-rane:
x-ref: https://issues.redhat.com/browse/CCO-121


joelddiaz (commented):
:+1:

akhil-rane:
/hold 

wait for #371

cmd/ccoctl/main.go
```diff
@@ -17,6 +17,7 @@ func main() {
 	}
 
 	rootCmd.AddCommand(aws.NewAWSCmd())
+	rootCmd.AddCommand(gcp.NewGCPCmd())
```
joelddiaz, line 20: duplicate of the line two below

pkg/cmd/provisioning/gcp/delete.go
```diff
@@ -0,0 +1,145 @@
...
+		Long:  "Deleting objects related to cloud credentials",
+		Run:   deleteCmd,
+	}
+
+	deleteCmd.PersistentFlags().StringVar(&DeleteOpts.Name, "name", "", "User-defined name for all created google cloud resources (can be separate from the cluster's infra-id)")
+	deleteCmd.MarkPersistentFlagRequired("name")
+	deleteCmd.PersistentFlags().StringVar(&DeleteOpts.Project, "project", "", "ID of the google cloud project")
+	deleteCmd.MarkPersistentFlagRequired("project")
+	deleteCmd.PersistentFlags().StringVar(&DeleteOpts.Region, "region", "", "Google cloud region where the resources were created")
+	deleteCmd.MarkPersistentFlagRequired("region")
```
joelddiaz, line 142: Where is this used?

akhil-rane: `region` is not required here. I should not make it a required field while creating the oidc bucket also. The default is us multi-region, which users can change if they want. sounds ok to you?

joelddiaz: Sounds reasonable to me.

PR #373, opened 2021-08-20
Update CI build root image to golang-1.16

akhil-rane:
Currently we are using golang-1.15 image in CI, but go.mod is set to go1.16. This is causing verify-deps test to fail.


2uasimojo:
All true.

/lgtm

akhil-rane:
@joelddiaz approve is missing here

PR #374, opened 2021-08-23
Enhance ccoctl ibmcloud command to create policies

mkumatag:
This enhancement will accept the `CredentialsRequest` in the following format.

```yaml
kind: CredentialsRequest
metadata:
  name: cr-openshift-machine-api-operator
spec:
  providerSpec:
    kind: IBMCloudProviderSpec
    policies:
      - roles:
          - "crn:v1:bluemix:public:iam::::serviceRole:Manager"
          - "crn:v1:bluemix:public:iam::::role:Editor"
          - "crn:v1:bluemix:public:iam::::role:Viewer"
        attributes:
          - name: "serviceName"
            value: "is"
      - roles:
          - "crn:v1:bluemix:public:iam::::role:Viewer"
        attributes:
          - name: "serviceName"
            value: "cloud-object-storage"
      - roles:
          - "crn:v1:bluemix:public:iam::::role:Viewer"
        attributes:
          - name: "serviceName"
            value: "power-iaas"
  secretRef:
    namespace: openshift-machine-api-operator
    name: ibmcloud-credential
```

Here is the sequence of flow:
1. Creates the [Service ID](https://cloud.ibm.com/docs/account?topic=account-serviceids&interface=ui) for every `CredentialsRequest`
2. Create the [policies](https://cloud.ibm.com/apidocs/iam-policy-management#list-policies) for the above-created Service ID
3. Create [API Key](https://cloud.ibm.com/docs/account?topic=account-serviceidapikeys&interface=ui#create_service_key) for the created Service ID


pkg/cmd/provisioning/ibmcloud/create_shared_secrets.go
```diff
@@ -67,80 +59,184 @@ func getEnv(envs []string) string {
...
 }
 
+// NewCreateSharedSecretsCmd provides the "create-shared-secrets" subcommand
+func NewCreateSharedSecretsCmd() *cobra.Command {
+	createSharedSecretsCmd := &cobra.Command{
+		Use:              "create-shared-secrets",
+		Short:            "Create shared secrets",
+		RunE:             createSharedSecretsCmd,
+		PersistentPreRun: initEnvForCreateIAMPoliciesCmd,
+	}
```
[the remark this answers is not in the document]
mkumatag: @akhil-rane you are right, `create-service-id` will be a term here.

pkg/cmd/provisioning/ibmcloud/create_shared_secrets.go
```diff
@@ -67,80 +59,184 @@ func getEnv(envs []string) string {
...
+
+		serviceIDOptions := &iamidentityv1.CreateServiceIDOptions{
+			AccountID: apiKeyDetails.AccountID,
+			Name:      &serviceIDName,
+		}
+		serviceID, err := client.CreateServiceID(serviceIDOptions)
+		if err != nil {
+			return errors.Wrap(err, "Failed to create Service ID")
+		}
+		log.Printf("Created IAM Service, Name: %s, ID: %s", *serviceID.Name, *serviceID.ID)
```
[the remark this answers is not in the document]
mkumatag: This is how the whole struct looks like:

```json
    {
        "id": "ServiceId-960fedcb-9f49-434f-9067-17728e004ddd",
        "iam_id": "iam-ServiceId-960fedcb-9f49-434f-9067-17728e004ddd",
        "crn": "crn:v1:bluemix:public:iam-identity::a/65b64c1f1c29460e8c2e4bbfbd893c2c::serviceid:ServiceId-960fedcb-9f49-434f-9067-17728e004ddd",
        "locked": false,
        "account_id": "65b64c1f1c29460e8c2e4bbfbd893c2c",
        "name": "test1-openshift-machine-api-operator-ibmcloud-credential",
        "description": "",
        "entity_tag": "1-8def162c025bf4a0a540db98808141d0",
        "created_at": "2021-08-21T05:30+0000",
        "modified_at": "2021-08-21T05:30+0000",
        "unique_instance_crns": []
    }
```

I don't see any relationship between the name and the id gets created, the name we supply will be fed to the `name` to the Service ID and you are right, 2 entries can have the same name, only id is unique what you get during creation.

mkumatag:
> Is there any way we can keep track of resources created in IBM cloud, something similar to aws tags or gcp labels? This would help in clearing out resources with ccoctl once cluster is uninstalled.

We do have the concept of tags in the IBM cloud but unfortunately, not all the resource got this luxury, and service ID is one of them and we can't really tag with it.

pkg/ibmcloud/client.go
```diff
@@ -0,0 +1,104 @@
...
+	clientID       string
+	identityClient *identityv1.IamIdentityV1
+}
+
+func (i *ibmcloudClient) CreateAPIKey(options *identityv1.CreateAPIKeyOptions) (*identityv1.APIKey, error) {
+	apikey, _, err := i.identityClient.CreateAPIKey(options)
+	if err != nil {
+		return nil, err
+	}
+	return apikey, err
```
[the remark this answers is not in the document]
mkumatag: make sense, lemme make those changes.

mkumatag:
@akhil-rane @joelddiaz fixed the review comments, ptal at the recent commit

pkg/cmd/provisioning/ibmcloud/create_service_id.go
```diff
@@ -59,26 +58,26 @@ func getEnv(envs []string) string {
...
 }
 
-// NewCreateSharedSecretsCmd provides the "create-shared-secrets" subcommand
-func NewCreateSharedSecretsCmd() *cobra.Command {
-	createSharedSecretsCmd := &cobra.Command{
-		Use:              "create-shared-secrets",
-		Short:            "Create shared secrets",
-		RunE:             createSharedSecretsCmd,
-		PersistentPreRun: initEnvForCreateIAMPoliciesCmd,
+// NewCreateServiceIDCmd provides the "create-shared-secrets" subcommand
```
akhil-rane, line 61:
```suggestion
// NewCreateServiceIDCmd provides the "create-service-id" subcommand
```

pkg/cmd/provisioning/ibmcloud/create_service_id.go
```diff
@@ -93,18 +92,24 @@ func createSharedSecretsCmd(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func createSharedSecrets(client ibmcloud.Client,
-	credReqDir, resourceGroupID, targetDir string) error {
+func createSharedSecrets(client ibmcloud.Client, accountID, resourceGroupID *string,
```
akhil-rane, line 111: function name should be changed to `createServiceIDs`

pkg/cmd/provisioning/ibmcloud/create_service_id.go
```diff
@@ -152,10 +151,10 @@ func processCredentialsRequests(client ibmcloud.Client,
 		serviceIDName := fmt.Sprintf("%s-%s-%s", CreateOpts.Name, cr.Spec.SecretRef.Namespace, cr.Spec.SecretRef.Name)
```
akhil-rane, line 151: For now, I think the only way of identifying the service ids created is by looking for this specific name format. Do we want to block the creation of a new service id if one with the name already exists?

akhil-rane (commented):
Looks good. few minor suggestionns

akhil-rane:
> > Is there information about what the plans are for when IBMCloud implements support for pod-identity/workload-identity? I would hope that understanding that, and how the IBMCloud Golang packages will work in that environment will help design a Secret format that can capture both posibilities.
> 
> I haven't started exploring that as of now, for auth [here](https://github.com/IBM/go-sdk-core/blob/main/Authentication.md) are the different possible types for the IBMCloud Golang package and I don't see anything like reading config file and make decision based on the content, all they have is different structs for different types and use them.

I am curious about [this](https://github.com/IBM/go-sdk-core/blob/e16080ef86a4e87959392a8020f6404b0eceb5ab/v5/core/iam_authenticator.go#L114-L131) function though. This comes close to reading from properties/config file

pkg/cmd/provisioning/ibmcloud/create_service_id.go
mkumatag: well, make sense, but what if someone creates service accounts with the same name after running this tool?! any either way we may end up deleting all those service IDs following this naming convention.

But this is going to be a little tricky because this tool runs on a directory looping through all the files and create service IDs, so will have to think about how to handle if failes: 

1. fail and keep the created service IDs?
2. fail and delete all the created service IDs?
3. run a loop through all the files and first check for the service ID existence and fail if exists

mkumatag: 3rd solution looks to be clean, lemme know if you have any other thought

akhil-rane: I agree, I am fine with the 3rd option

mkumatag:
> I am curious about [this](https://github.com/IBM/go-sdk-core/blob/e16080ef86a4e87959392a8020f6404b0eceb5ab/v5/core/iam_authenticator.go#L114-L131) function though. This comes close to reading from properties/config file. You cannot directly pass the filepath. Maybe reading properties file should be done outside the function. I wonder if they might use it to support pod-identity/workload-identity in the future.

Over the last few days, I was exploring how to achieve service account-based auth and found this [App ID](https://www.ibm.com/in-en/cloud/app-id) with custom identity integration, looking at APIs, we may need to write our own authenticator. [container_authenticator](https://github.com/IBM/go-sdk-core/blob/e16080ef86a4e87959392a8020f6404b0eceb5ab/v5/core/container_authenticator.go) is close to what we are looking for one used for loading the token from the file but parameters like grant type, url etc not present in this authenticator. We will have more clarity once we explore more about the App ID integration and a flow

[this](https://github.com/IBM/go-sdk-core/blob/e16080ef86a4e87959392a8020f6404b0eceb5ab/v5/core/iam_authenticator.go#L114-L131) is mainly used for the different services to load the file use it, but considering our use case we wouldn't know what all services requester is going to connect, and environment file differs from requester to requester e.g: machine-api-operator requires vpc access, ingress needs cis, vpc etc.. And also we don't really know if all the sdks are really making use of [this function](https://github.com/IBM/go-sdk-core/blob/e16080ef86a4e87959392a8020f6404b0eceb5ab/v5/core/authenticator_factory.go#L24) to authenticate.

mkumatag:
@akhil-rane refactored the code with a better validation and cleanup logic

mkumatag:
Will fix the UT later

joelddiaz:
After reviewing some of the discussion here about the Secret format and what the future with pod/workload identity might look like here's what I have in mind.

I feel like we could use the function at https://github.com/IBM/go-sdk-core/blob/main/v5/core/authenticator_factory.go#L24 as a common credentials/config loading option.
What I see is a Secret format that looks like:
```yaml
apiVersion: v1
stringData:
  ibm-credentials.env: |
    IBMCLOUD_APIKEY=xxxxx
    IBMCLOUD_AUTHTYPE=iam
kind: Secret
metadata:
  name: ibmcloud-credential
  namespace: openshift-machine-api-operator
type: Opaque
```

Then the in-cluster components can read in the Secret (can't mount it in at the Pod Spec level b/c we need to ensure the Pod runs no matter which cloud/platform), set the IBM_CREDENTIALS_FILE env var to point to the file holding the Secret contents.
At this point we should be able to call the v5/core/GetAuthenticatorFromEnvironment() with "IBMCLOUD" as the `credentialsKey` parameter.

If I'm reading this correctly, the code path will see that we have AUTHTYPE_IAM and read in the API key (and any other settable env vars) as it sets up the `Authenticator` object.

(Now here's where I start to guess...) If all this is doable, then supporting pod/workload identity would follow a similar pattern to the "container" auth type. So the Secret format for IBMCloud when using pod/workload identity might look like:

```yaml
apiVersion: v1
stringData:
  ibm-credentials.env: |
    IBMCLOUD_CR_TOKEN_FILENAME=/var/run/secrets/openshift/serviceaccount/token ### this is where we always project the signed ServiceAccount token
    IBMCLOUD_IAM_PROFILE_NAME=name-of-ibmcloud-iam-role/profile
    IBMCLOUD_AUTHTYPE=podidentity  ### or "container" if it is extended to work as-is for pod/workload identity
kind: Secret
metadata:
  name: ibmcloud-credential
  namespace: openshift-machine-api-operator
type: Opaque
```

Does what I describe seem reasonable to allow us to not have to inject authentication-type-specific logic into each in-cluster component?

joelddiaz:
Now that I have an IBMCloud account I was able to try this out for myself. One thing I ran into was that I passed the resource group name as the `--resource-group-id` parameter. I didn't get any errors out of ccoctl, but then I couldn't even read the objects from the `ibmcloud` cli.

```
[jdiaz@minigoomba ~]$ ibmcloud iam service-policies jdiaztest-openshift-image-registry-installer-cloud-credentials
Getting all policies of service ID jdiaztest-openshift-image-registry-installer-cloud-credentials under 3cdf7bdcbafa4dcc944eb3bdd3f8bef5 as jdiaz@redhat.com...
OK
FAILED
The following error(s) occurred:
Invalid Id: jdiaztestrg

Trace ID: 36009402075
```

If the IBMCloud sdk isn't going to error on providing a non-existent resource-group-id, can we just fetch the list of resource groups and make sure that the provided resource-group-id exists?
And even better, can we support `--resource-group-name` and have ccoctl do the resource-group-name---->resource-group-id translation for us?

mkumatag:
> Then the in-cluster components can read in the Secret (can't mount it in at the Pod Spec level b/c we need to ensure the Pod runs no matter which cloud/platform), set the IBM_CREDENTIALS_FILE env var to point to the file holding the Secret contents.
> At this point we should be able to call the v5/core/GetAuthenticatorFromEnvironment() with "IBMCLOUD" as the `credentialsKey` parameter.

Seems reasonable, I will include this logic and refactor the code.

mkumatag:
> And even better, can we support `--resource-group-name` and have ccoctl do the resource-group-name---->resource-group-id translation for us?

even I have seen this issue, createPolicy didn't throw any error with invalid resource-group-id, better I will switch to name and have validation logic for checking it.

mkumatag:
@joelddiaz Included the changes in the latest commit, ptal
- Modified the code to generate the creds in the `ibm-credentials.env`
- Use resource-group-name instead of resource-group-id

joelddiaz (commented):
This looks good to me. One thing is that if we already have in-cluster components expecting the old Secret format (with the `ibmcloud_api_key` field), we should probably temporarily keep populating that field in the Secret. Update the in-cluster components to read/handle the new `ibm-credentials.env` field, then once all the PRs land we can do an additional PR for ccoctl to stop populating `ibmcloud_api_key`.

mkumatag:
> This looks good to me. One thing is that if we already have in-cluster components expecting the old Secret format (with the `ibmcloud_api_key` field), we should probably temporarily keep populating that field in the Secret. Update the in-cluster components to read/handle the new `ibm-credentials.env` field, then once all the PRs land we can do an additional PR for ccoctl to stop populating `ibmcloud_api_key`.

I fixed the following things in the latest commit:
- Included ibmcloud_api_key on a temporary basis
- As resource group is an optional param while creating the policies, so accommodated those changes
- Fixed all the UT and added more test cases

pkg/cmd/provisioning/ibmcloud/service_id.go
```diff
@@ -234,8 +236,13 @@ func (s *ServiceID) UnDo(targetDir string) error {
 	}
 
 	secretFileName := filepath.Join(targetDir, manifestsDirName, fmt.Sprintf(secretFileNamePattern, s.cr.Spec.SecretRef.Namespace, s.cr.Spec.SecretRef.Name))
+	fmt.Printf("secretFileName: %s\n", secretFileName)
```
joelddiaz, line 239: should this be a log message, or just removed entirely?

joelddiaz (commented):
this looks good. what ever came of the discussion of implementing a 'delete' subcommand that can automate cleaning up these Service IDs?

pkg/cmd/provisioning/ibmcloud/service_id.go
mkumatag: Will remove this line..

mkumatag:
> this looks good. what ever came of the discussion of implementing a 'delete' subcommand that can automate cleaning up these Service IDs?

we can have a follow-up PR for deleting the serviceIDs.

joelddiaz (commented):
ok. this looks good. can we squash the commits down to 1 (preferably 2 commits: one for the vendoring and one for the ccoctl code, but i can live with one commit)?

mkumatag:
> ok. this looks good. can we squash the commits down to 1 (preferably 2 commits: one for the vendoring and one for the ccoctl code, but i can live with one commit)?

done, you have 2 commits(1 for vendor and another for actual code)

joelddiaz (commented):
/lgtm

akhil-rane:
@joelddiaz do we need `dry-run` here option here?

joelddiaz:
> @joelddiaz do we need `dry-run` here option here?

We can add that work in with the followup delete PR. But I agree it would be nice to have a --dry-run option where we can feed some files and/or scripts through the ibmcloud CLI

mkumatag:
> > @joelddiaz do we need `dry-run` here option here?
> 
> We can add that work in with the followup delete PR. But I agree it would be nice to have a --dry-run option where we can feed some files and/or scripts through the ibmcloud CLI

will submit a separate PR for the `--dry-run`

PR #379, opened 2021-09-03
Bug 1991770: The logLevel and operatorLogLevel values do not work with Cloud Credential Operator

MichaelWasher:
Add loglevelcontroller to reconcile on changes to CloudCredential object
Change the loglevel on CR updates
Add openshift/library-go vendor modules (operator/loglevel)

### DIFF
--- a/pkg/operator/loglevel/controller.go
+++ b/pkg/operator/loglevel/controller.go
@@ -1,0 +1,153 @@
+/*
+Copyright 2020 The OpenShift Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package loglevel
+
+import (
+	"fmt"
+
+	log "github.com/sirupsen/logrus"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	logLevelUtils "github.com/openshift/library-go/pkg/operator/loglevel"
+
+	"github.com/openshift/cloud-credential-operator/pkg/operator/status"
+	"github.com/openshift/cloud-credential-operator/pkg/operator/utils"
+)
+
+const (
+	controllerName = "loglevel"
+)
+
+// Add creates a new ConfigController and adds it to the Manager.
+func Add(mgr manager.Manager, kubeConfig string) error {
+	r := newReconciler(mgr)
+
+	// Create a new controller
+	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to CloudCredential object and reconcile the loglevel changes
+	err = c.Watch(
+		&source.Kind{Type: &operatorv1.CloudCredential{}},
+		&handler.EnqueueRequestForObject{})
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	r := &ReconcileCloudCredentialConfig{
+		Client: mgr.GetClient(),
+	}
+	status.AddHandler(controllerName, r)
+
+	return r
+}
+
+var _ reconcile.Reconciler = &ReconcileCloudCredentialConfig{}
+
+// ReconcileCloudCredentialConfig reconciles a CredentialConfig object
+type ReconcileCloudCredentialConfig struct {
+	client.Client
+}
+
+// Reconcile reads the state of the cluster for a CredentialConfig object and
+// makes changes based on the state read and what is in the CredentialConfig.LogLevel
+// Automatically generate RBAC rules to allow the Controller to read and write required types.
+// +kubebuilder:rbac:groups=operator.openshift.io,resources=cloudcredential/spec,verbs=get
+func (r *ReconcileCloudCredentialConfig) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	logger := log.WithFields(log.Fields{
+		"controller": controllerName,
+		"cr":         fmt.Sprintf("%s/%s", request.NamespacedName.Namespace, request.NamespacedName.Name),
+	})
+
+	// Check OperatorLogLevel and LogLevel
+	log.Debug("Checking log level for Reconcile request")
+	desiredOperatorLogLevel, err := utils.GetOperatorLogLevel(r, logger)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if !logLevelUtils.ValidLogLevel(desiredOperatorLogLevel) {
+		return reconcile.Result{}, fmt.Errorf("the operatorLogLevel defined for the cloudcredential object is invalid")
+	}
+
+	desiredLogLevel, err := utils.GetLogLevel(r, logger)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if !logLevelUtils.ValidLogLevel(desiredLogLevel) {
+		return reconcile.Result{}, fmt.Errorf("the logLevel defined for the cloudcredential object is invalid")
+	}
+
+	// Pick the logLevel with the highest verbosity level.
+	maxLogLevel := desiredOperatorLogLevel
+	if logLevelUtils.LogLevelToVerbosity(desiredLogLevel) > logLevelUtils.LogLevelToVerbosity(desiredOperatorLogLevel) {
+		maxLogLevel = desiredLogLevel
+	}
+
+	currentLogLevel, errBool := logLevelUtils.GetLogLevel()
+	if errBool {
+		log.Debug("Unable to get the current loglevel. Defaulting the current level to 'Normal'")
+	}
+
+	if currentLogLevel == maxLogLevel {
+		return reconcile.Result{}, nil
+	}
+
+	// Set klog logLevel to the desired level
+	err = logLevelUtils.SetLogLEvel(maxLogLevel)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	log.Infof("klog level changed to %v", maxLogLevel)
+
+	// Set lorus logLevel to the desired level
+	logrusLevel := crLogLevelToLogrusLevel(maxLogLevel)
+	log.SetLevel(logrusLevel)
+	log.Infof("logrus level changed to %s", logrusLevel)
+
+	return reconcile.Result{}, nil
+}
+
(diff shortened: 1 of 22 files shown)


pkg/operator/loglevel/status.go
```diff
@@ -0,0 +1,22 @@
+package loglevel
+
+import (
+	log "github.com/sirupsen/logrus"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cloud-credential-operator/pkg/operator/status"
```
akhil-rane, line 7: keep openshift imports in a separate block below

pkg/operator/loglevel/controller.go
```diff
@@ -0,0 +1,113 @@
...
+	if errBool {
+		log.Debug("Unable to get the current loglevel. Defaulting to 'Normal'")
+	}
+
+	if currentLogLevel == desiredLogLevel {
+		return reconcile.Result{Requeue: false}, nil
+	}
+
+	// Set logLevel to the new level
+	logLevelUtils.SetLogLEvel(desiredLogLevel)
```
akhil-rane, line 110:
```suggestion
	logLevelUtils.SetLogLevel(desiredLogLevel)
```

akhil-rane: This function returns an error which needs to be handled

pkg/operator/utils/utils.go
```diff
@@ -206,6 +206,16 @@ func GetOperatorConfiguration(kubeClient client.Client, logger log.FieldLogger)
...
 
+func GetOperatorLogLevel(kubeClient client.Client, logger log.FieldLogger) (operatorv1.LogLevel, error) {
+
+	conf, err := getOperatorConfiguration(kubeClient, logger)
+	if err != nil {
+		return "", err
+	}
+	// TODO Configuration conflicts
+	return conf.Spec.LogLevel, nil
+}
```
akhil-rane, line 217: `GetOperatorLogLevel` should return `conf.Spec.OperatorLogLevel` I guess

pkg/operator/loglevel/controller.go
akhil-rane, line 110: I think we need code to set both `logLevel` and `operatorLogLevel`

MichaelWasher: Yeah, I agree, but I wasn't sure what to set when logLevel is changed, as this (AFAIK) Operator doesn't have managed components but instead has the actuators in the Operator.

Would you prefer if both operatorLogLevel and logLevel define the Operator loglevel and the highest verbosity wins?

pkg/operator/utils/utils.go
MichaelWasher: You are correct, that is my mistake. I'll fix that up.

pkg/operator/loglevel/controller.go
MichaelWasher: I've updated this to catch the error, but I have not changed the function name. There's a typo in the library-go version used.
https://github.com/openshift/cloud-credential-operator/pull/379/files#diff-e5b905c5074fe810ed7617bc85a7128c651866094cb4934536459b932ce09236R65

I didn't want my first PR to contain updates to the library versions, but if you would prefer I can update the version of library-go.

The PR this was changed is here: 
https://github.com/openshift/library-go/commit/15d11d2f6bfcbe15679acd184ac69c77aa2e65bc

pkg/operator/loglevel/controller.go
```diff
@@ -0,0 +1,128 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/openshift/cloud-credential-operator/pkg/operator/status"
+	"github.com/openshift/cloud-credential-operator/pkg/operator/utils"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	logLevelUtils "github.com/openshift/library-go/pkg/operator/loglevel"
```
akhil-rane, line 35: We prefer to keep cco imports at the end

```suggestion
	operatorv1 "github.com/openshift/api/operator/v1"
	logLevelUtils "github.com/openshift/library-go/pkg/operator/loglevel"
	
	"github.com/openshift/cloud-credential-operator/pkg/operator/status"
	"github.com/openshift/cloud-credential-operator/pkg/operator/utils"
```

pkg/operator/loglevel/controller.go
```diff
@@ -0,0 +1,128 @@
...
+		"controller": controllerName,
+		"cr":         fmt.Sprintf("%s/%s", request.NamespacedName.Namespace, request.NamespacedName.Name),
+	})
+
+	// Check OperatorLogLevel and LogLevel
+	log.Debug("Checking log level for Reconcile request")
+	desiredOperatorLogLevel, err := utils.GetOperatorLogLevel(r, logger)
+	if err != nil || !logLevelUtils.ValidLogLevel(desiredOperatorLogLevel) {
+		return reconcile.Result{Requeue: false}, fmt.Errorf("the operatorLogLevel defined for the cloudcredential object is invalid")
+	}
```
akhil-rane, line 96: this can error out for two reasons
1) error fetching the desired log level on CredentialConfig
2) desired log level is invalid

I would go with the following
```suggestion
	desiredOperatorLogLevel, err := utils.GetOperatorLogLevel(r, logger)
	if err != nil {
		return reconcile.Result{}, err
	}
	if !logLevelUtils.ValidLogLevel(desiredOperatorLogLevel) {
		return reconcile.Result{}, fmt.Errorf("the operatorLogLevel defined for the cloudcredential object is invalid")
	}
```

pkg/operator/loglevel/controller.go
```diff
@@ -0,0 +1,128 @@
...
+		maxLogLevel = desiredLogLevel
+	}
+
+	currentLogLevel, errBool := logLevelUtils.GetLogLevel()
+	if errBool {
+		log.Debug("Unable to get the current loglevel. Defaulting the current level to 'Normal'")
+	}
+
+	if currentLogLevel == maxLogLevel {
+		return reconcile.Result{Requeue: false}, nil
```
akhil-rane, line 115: nit: If I am not wrong requeue is false by default so `return reconcile.Result{}, nil` should work the same

pkg/operator/loglevel/controller.go
akhil-rane: I am not sure if this deserves a library-go version update. I am leaning towards no. @joelddiaz wdyt?

akhil-rane: > Would you prefer if both operatorLogLevel and logLevel define the Operator loglevel and the highest verbosity wins?

Yes, that sounds good to me

pkg/operator/loglevel/controller.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Set logLevel to the new level
+	err = logLevelUtils.SetLogLEvel(maxLogLevel)
+	if err != nil {
+		log.Errorf("Unable to set loglevel. %v", err)
+		return reconcile.Result{Requeue: false}, nil
+	}
+
+	log.Infof("Log level changed to %v", maxLogLevel)
+
```
joelddiaz, line 126: This a little tricky for our operator. We use logrus throughout, and the logLevelUtils.GetLogLevel() returns the level for klog.
I would propose dropping this detection logic and just blindly assigning/re-assigning the log level based on a change the CloudCredentials CR.

And because we use logrus, we need to handle setting the log level for that package as well.

Something like this:

```suggestion
	// Set klog logLevel to the desired level
	err = logLevelUtils.SetLogLEvel(maxLogLevel)
	if err != nil {
		log.Errorf("Unable to set loglevel. %v", err)
		return reconcile.Result{Requeue: false}, nil
	}

	log.Infof("klog level changed to %v", maxLogLevel)
	
	var logrusLevel log.Level
	switch maxLogLevel {
	case operatorv1.Debug:
	        logrusLevel = log.DebugLevel
	case operatorv1.Trace, operatorv1.TraceAll:
	        logrusLevel = log.TraceLevel
	case operatorv1.Normal:
	       logrusLevel = log.InfoLevel
	 default:
	         logrusLevel = log.InfoLevel
	  }
	  
	  log.SetLevel(logrusLevel)
	  log.Infof("logrus level changed to %s", logrusLevel)

```

joelddiaz (commented):
Seems reasonable. Just one bit of feedback about handling log level for both klog and logrus.

pkg/operator/loglevel/controller.go
MichaelWasher: Sorry, I misunderstood what you were referencing previously. I have resolved this now.

pkg/operator/loglevel/controller.go
MichaelWasher: I have split both compound err checks to follow the recommended pattern.

pkg/operator/loglevel/controller.go
MichaelWasher: I have changed this for all of the `reconcile.Result{Requeue: false}, nil`  returns

pkg/operator/loglevel/controller.go
MichaelWasher: I have made these changes to include the logrus level changes.
https://github.com/openshift/cloud-credential-operator/pull/379/files#diff-e983e87892f7cdbd2b02917acb48aa087f3a4955ddf91b27354a2ce0dcb0dcbfR125-R153

I did keep the early exit clause:
```
	if currentLogLevel == maxLogLevel {
		return reconcile.Result{Requeue: false}, nil
	}
```

If you think this is wrong, let me know and I can fix it up :)

pkg/operator/loglevel/controller_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+			_, err := rccc.Reconcile(reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Name:      constants.CloudCredOperatorConfig,
+					Namespace: testNamespace,
+				},
+			})
+			if err != nil {
+				require.NoError(t, err, "Unexpected error: %v", err)
+			}
```
joelddiaz, line 88: Don't really need to check whether err != nil, just require that it not be an error unconditionally
```suggestion
			require.NoError(t, err, "Unexpected error: %v", err)
```

joelddiaz (commented):
One small nit, otherwise :+1:

pkg/operator/loglevel/controller_test.go
MichaelWasher: Thanks for the advice, super new to unit-testing in Go. Youtube++ 

All fixed up now. How do you normally merge here? (automatically squash commits or should I manually squash ?)

MichaelWasher:
I hate to be the 'new guy' and blame the CI but is the CI a little flakey?
It's just that `47765db3f59fb5b1275d546c1c608479e55f7267` was green but `7c98ea907843bb299995e9222fd74f6043ae8484` red-built on ci/prow/e2e-aws?

pkg/operator/loglevel/controller_test.go
joelddiaz: Go ahead an squash on your end, and then we can lgtm it and put it in the merge queue

joelddiaz:
> I hate to be the 'new guy' and blame the CI but is the CI a little flakey?
> It's just that [47765db](https://github.com/openshift/cloud-credential-operator/commit/47765db3f59fb5b1275d546c1c608479e55f7267) was green but [7c98ea9](https://github.com/openshift/cloud-credential-operator/commit/7c98ea907843bb299995e9222fd74f6043ae8484) red-built on ci/prow/e2e-aws with warnings that seemed unrelated.

Yes, flakes are our present burden :/ there is ongoing work to help the situation. I wouldn't not look at the CI failures until you've seen it fail for more than 1 day straight. Once you get an lgtm, it will retry automatically for you.

pkg/operator/loglevel/controller_test.go
MichaelWasher: Cool, I've squashed into a single commit as this seems to be what most PRs are. If you want it split into multiple, I can roll it back. Otherwise, hopefully this'll get me my first OpenShift lgtm 😄

joelddiaz (commented):
/lgtm

PR #383, opened 2021-09-15
Bug 2002152: Stop putting CCO in degraded state when stale credentials are found

akhil-rane:
Currently, we put CCO in the degraded state when stale credentials are found. When in manual mode, stale credentials are not removed by the cleanup controller and thus upgrade is blocked. This commit stops it and instead we raise an alert warning that cluster admin can see and take necessary action.

### DIFF
--- a/manifests/0000_90_cloud-credential-operator_04_alertrules.yaml
+++ b/manifests/0000_90_cloud-credential-operator_04_alertrules.yaml
@@ -42,3 +42,11 @@ spec:
       for: 5m
       labels:
         severity: warning
+    - alert: CloudCredentialOperatorStaleCredentials
+      annotations:
+        message: 1 or more credentials requests are stale and should be deleted. Check the status.conditions on CredentialsRequest CRs to identify the stale one(s).
+      expr: cco_credentials_requests_conditions{condition="StaleCredentials"}
+        > 0
+      for: 5m
+      labels:
+        severity: warning
--- a/pkg/operator/metrics/metrics.go
+++ b/pkg/operator/metrics/metrics.go
@@ -231,6 +231,7 @@ func newAccumulator(client client.Client, logger log.FieldLogger) *credRequestAc
 	for _, c := range credreqv1.FailureConditionTypes {
 		acc.crConditions[c] = 0
 	}
+	acc.crConditions[credreqv1.StaleCredentials] = 0
 
 	return acc
 }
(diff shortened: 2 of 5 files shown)


akhil-rane:
yet to be tested

manifests/0000_90_cloud-credential-operator_04_alertrules.yaml
```diff
@@ -42,3 +42,11 @@ spec:
       for: 5m
       labels:
         severity: warning
+    - alert: CloudCredentialOperatorStaleCredentials
+      annotations:
+        message: 1 or more credentials requests are stale and should be deleted
```
joelddiaz, line 47:
```suggestion
        message: 1 or more credentials requests are stale and should be deleted. Check the status.conditions on CredentialsRequest CRs to identify the stale one(s).
```

joelddiaz (commented):
i thought this was going to be more involved to move away from degraded. nice!

joelddiaz:
> @joelddiaz when CCO is in manual mode, we skip reporting conditions and alerts are not raised. Should if we make exception for stale credentials or just let the user handle it and not raise an alert?

Yeah, I suppose it makes sense to collect and report the stale-conditions count up into metrics even when in manual mode so that the alert can fire. We already update the status on the CredReq objects when in Manual mode...
Otherwise we have no communication mechanism to the cluster admin(s).

joelddiaz (commented):
/lgtm

pkg/operator/metrics/metrics.go
```diff
@@ -252,10 +253,19 @@ func (a *credRequestAccumulator) processCR(cr *credreqv1.CredentialsRequest, cco
...
+			if cond.Status == corev1.ConditionTrue && cond.Type != credreqv1.StaleCredentials {
+				a.crConditions[cond.Type]++
+			}
+		}
+	} else {
+		for _, cond := range cr.Status.Conditions {
+			if cond.Status == corev1.ConditionTrue && cond.Type == credreqv1.StaleCredentials {
 				a.crConditions[cond.Type]++
 			}
 		}
```
akhil-rane, line 271: This code can be refactored as follows but I felt the existing one is more readable

```suggestion
       for _, cond := range cr.Status.Conditions {
		if cond.Status == corev1.ConditionTrue && ((cond.Type != credreqv1.StaleCredentials) != ccoDisabled) {
			a.crConditions[cond.Type]++
		}
	}
```

akhil-rane (commented):
tested and ready to go

joelddiaz (commented):
/lgtm

akhil-rane:
@sdodson what is the best way to get `bugzilla/valid-bug` label here? Do I need to remove the BZ `nec` group?

sdodson:
> @sdodson what is the best way to get `bugzilla/valid-bug` label here? Do I need to remove the BZ `nec` group?

I removed the extra private group from the bug while leaving `redhat` private group on the bug.

joelddiaz:
@lwan-wanglin it looks like we need your qe-approved label according to https://bugzilla.redhat.com/show_bug.cgi?id=2002152#c16

PR #385, opened 2021-09-20
Bug 2006364: fix(ibmcloud): Set policy attribute name properly for RG access

hasueki:
This change allows `ccoctl` to set access policy attributes properly for a specific resource group access. In cases where we want to assign access to just the resource group, the policy attributes should look like the following:
```
{
  "attributes":[
    {
      "name":"accountId",
      "value":"my-account-id",
    },
    {
      "name":"resourceType",
      "value":"resource-group",
    },
    {
      "name":"resource",
      "value":"my-resource-group-id"
    }
  ],
}
```

To compare, the attributes generated without this fix will result in..
```
{
  "attributes":[
    {
      "name":"accountId",
      "value":"my-account-id",
    },
    {
      "name":"resourceType",
      "value":"resource-group",
    },
    {
      "name":"resourceGroupId",
      "value":"my-resource-group-id"
    }
  ],
}
```

.. which does not grant access to the resource group as expected.

This fix will determine if `resourceType == "resource-group"`, then set the additional attribute of `resourceGroupId || resource` appropriately.

### DIFF
--- a/pkg/cmd/provisioning/ibmcloud/service_id.go
+++ b/pkg/cmd/provisioning/ibmcloud/service_id.go
@@ -172,8 +172,15 @@ func (s *ServiceID) createPolicy(policy *credreqv1.AccessPolicy) error {
 
 	// Append the resource group attribute if specified as a command line argument.
 	if s.resourceGroupID != "" {
+		resourceGroupAttrName := "resourceGroupId"
+		for _, attr := range resourceAttributes {
+			if *attr.Name == "resourceType" && *attr.Value == "resource-group" {
+				resourceGroupAttrName = "resource"
+				break
+			}
+		}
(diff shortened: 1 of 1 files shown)


joelddiaz:
With our new no-feature-freeze development style (just started a week ago), we need a valid BZ for this. Otherwise, we need to get review from docs, qe, and product experience (it's easier to just open a BZ ;) ).

mkumatag (approved):
@hasueki thanks for the fix, this fix is for assigning proper resource-group-id for the resource type `resource-group`(they have a different field with compare to other resources)

cjschaef:
Here's the BZ, the `Target Release` needs to be set to successfully target it in this PR
https://bugzilla.redhat.com/show_bug.cgi?id=2006364

PR #386, opened 2021-09-23
require spec.serviceAccountNames

joelddiaz:
No longer treat CredentialsRequests without spec.serviceAccountNames
filled out as valid. This creates IAM Roles that limit the Role by
audience which means any ServiceAccount in the cluster could potentially
assume the IAM Role (because they all share the same audience).

All CredentialsRequests now populate the ServiceAccountNames field, so
we can remove this transitional support.

xref: https://issues.redhat.com/browse/CCO-104

### DIFF
--- a/pkg/cmd/provisioning/aws/create-iam-roles.go
+++ b/pkg/cmd/provisioning/aws/create-iam-roles.go
@@ -112,7 +112,7 @@ func createRole(awsClient aws.Client, name string, credReq *credreqv1.Credential
 
 	rolePolicyDocument, err := createRolePolicyDocument(oidcProviderARN, issuerURL, credReq.Spec.SecretRef.Namespace, credReq.Spec.ServiceAccountNames)
 	if err != nil {
-		return "", errors.Wrap(err, "Error while create Role policy document")
+		return "", errors.Wrapf(err, "error while creating Role policy document for %s", credReq.Name)
 	}
 
 	roleDescription := fmt.Sprintf("OpenShift role for %s/%s", credReq.Spec.SecretRef.Namespace, credReq.Spec.SecretRef.Name)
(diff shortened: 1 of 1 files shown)


pkg/cmd/provisioning/aws/create-iam-roles.go
```diff
@@ -256,7 +256,11 @@ func createRolePolicyDocument(oidcProviderARN, issuerURL, namespace string, serv
 		conditionString = fmt.Sprintf(` { "StringEquals": { "%s:sub": %s } }`, issuerURL, serviceAccountListString)
 	} else {
 		// TODO: maybe return an error once all CredentialsRequest start including ServiceAccountNames
-		conditionString = fmt.Sprintf(`{ "StringEquals": { "%s:aud": "openshift" } }`, issuerURL)
+		// We used to support leaving the list of ServiceAccounts blank in the CredentialsRequest while
+		// we transitioned all the existing CredReqs. That work is complete, and we should not
+		// create Role policies that limit the Role by audience any more.
+		// Return an error indicating that ccoctl requires the ServiceAccount list to be populated.
+		return "", fmt.Errorf("must provide ServieAccounts to bind the Role policy to")
```
akhil-rane, line 263:
```suggestion
		return "", fmt.Errorf("CredentialsRequest must provide ServieAccounts to bind the Role policy to")
```

joelddiaz: :+1: should be in the latest push

joelddiaz:
/assign @lwan-wanglin @jeana-redhat @sferich888 
The TL;DR is: when implementing the first version of `ccoctl` we allowed not specifying the ServiceAccount list (as not all repos had added that info yet). Now all in-cluster operators have updated their CredentialsRequest manifests to include the ServiceAccounts. Creating IAM Roles that bind to a specific ServiceAccount instead of to the cluster-wide `audience` field is more secure, and we will now raise an error if someone tries to use `ccoctl` to create an IAM Role where the CredentialsRequest has not filled out the ServiceAccount names list.

jeana-redhat:
/label docs-apporoved
thanks for the summary, Joel!

jeana-redhat:
thanks for the summary, Joel!

/label docs-approved

lwan-wanglin:
/label qe-approved

test pass:
prepare a CredentialsRequest without  .spec.serviceAccountNames, run ccoctl to create policies, it will hit below error:
`
2021/09/26 15:50:51 Failed to process IAM Roles: Failed while processing each CredentialsRequest: error while creating Role policy document for openshift-machine-api-aws: CredentialsRequest must provide ServieAccounts to bind the Role policy to
`
the test case: https://url.corp.redhat.com/bbfbefd

PR #387, opened 2021-09-23
Bug 1976674: make azure and ovirt consistent when in manual mode

joelddiaz:
We started to require the upgradeable-to annotation when in manual mode.
Make sure Azure and oVirt behave like the other platforms.

### DIFF
--- a/pkg/azure/actuator.go
+++ b/pkg/azure/actuator.go
@@ -23,8 +23,7 @@ import (
 	"reflect"
 	"time"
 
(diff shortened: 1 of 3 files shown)


akhil-rane (commented):
Looks good! should we also do the same for KubeVirt?

joelddiaz:
> Looks good! should we also do the same for KubeVirt?

:+1: Seems like a good idea. Pushed kubevirt changes with latest commit.

akhil-rane (commented):
/lgtm

PR #388, opened 2021-09-24
Bug 1977319: cleanup orphaned Service 'controller-manager-service'

joelddiaz:
Back in 4.1 we installed a Service named 'controller-manager-service'.
That was subsequently dropped, but any cluster that has upgraded from
4.1 to present keeps the needless Service arround. Re-create the Service
  with the delete annotation so that the CVO can properly clean this up.

### DIFF
--- a/manifests/01-service-delete.yaml
+++ b/manifests/01-service-delete.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    release.openshift.io/delete: "true"
+  labels:
+    control-plane: controller-manager
+    controller-tools.k8s.io: "1.0"
+  name: controller-manager-service
+  namespace: openshift-cloud-credential-operator
+spec:
+  ports:
+  - port: 443
+  selector:
+    control-plane: controller-manager
+    controller-tools.k8s.io: "1.0"


akhil-rane (commented):
/lgtm

gone are the days of cleanup controller

manifests/01-service-delete.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    release.openshift.io/delete: "true"
```
wking, line 5: The CVO will ignore this unless you set [profile annotations][1].  Checking your other manifests:

```console
$ git grep -oh 'include.release.openshift.io/.*' manifests | sort | uniq -c
     14 include.release.openshift.io/ibm-cloud-managed: "true"
     20 include.release.openshift.io/self-managed-high-availability: "true"
```

[1]: https://github.com/openshift/enhancements/blob/ce4d303db807622687159eb9d3248285a003fabb/enhancements/update/cluster-profiles.md

joelddiaz: Wow, our manifest is so old it predates all that stuff. Thanks for the quick catch. You saved us some QE time :1st_place_medal:

lwan-wanglin:
/label qe-approved

test pass:
1. installed a 4.9 cluster , and create a service controller-manager-service
 `$ cat 01-service.yaml 
apiVersion: v1
kind: Service
metadata:
  labels:
    control-plane: controller-manager
    controller-tools.k8s.io: "1.0"
  name: controller-manager-service
spec:
  ports:
  - port: 443
  selector:
    control-plane: controller-manager
    controller-tools.k8s.io: "1.0" `
2. Check clusterversion
`
$ oc get clusterversion
NAME      VERSION      AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.9.0-rc.3   True        False         7h34m   Cluster version is 4.9.0-rc.3
`
3. Check service
`
$ oc get service -n openshift-cloud-credential-operator
NAME                         TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
cco-metrics                  ClusterIP   172.30.50.225   <none>        8443/TCP   7h55m
controller-manager-service   ClusterIP   172.30.67.51    <none>        443/TCP    18m
pod-identity-webhook         ClusterIP   172.30.37.210   <none>        443/TCP    7h47m
`
4. upgrade cluster to a version with this fix
`
oc adm upgrade --to-image registry.build01.ci.openshift.org/ci-ln-tnfrvhb/release:latest --allow-explicit-upgrade --force
warning: Using by-tag pull specs is dangerous, and while we still allow it in combination with --force for backward compatibility, it would be much safer to pass a by-digest pull spec instead
warning: The requested upgrade image is not one of the available updates.  You have used --allow-explicit-upgrade to the update to proceed anyway
warning: --force overrides cluster verification of your supplied release image and waives any update precondition failures.
`
5. After upgrade, check service/controller-manager-service is deleted
`
$ oc get service -n openshift-cloud-credential-operator
NAME                   TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
cco-metrics            ClusterIP   172.30.50.225   <none>        8443/TCP   8h
pod-identity-webhook   ClusterIP   172.30.37.210   <none>        443/TCP    8h
`

PR #390, opened 2021-09-28
Add UT for the delete-service-id command

mkumatag:
Add the Unit testscases for the `ccoctl ibmcloud delete-service-id` command

### DIFF
--- a/pkg/cmd/provisioning/ibmcloud/delete_service_id_test.go
+++ b/pkg/cmd/provisioning/ibmcloud/delete_service_id_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/IBM/go-sdk-core/v5/core"
@@ -23,19 +24,22 @@ func Test_deleteServiceIDCmd(t *testing.T) {
 		args []string
 	}
 	tests := []struct {
-		name    string
-		args    args
-		wantErr bool
+		name        string
+		args        args
+		expectError string
 	}{
 		{
-			name:    "CreateSharedSecretsCmd with unset API key environment variable should fail",
-			wantErr: true,
+			name:        "CreateSharedSecretsCmd with unset API key environment variable should fail",
+			expectError: "environment variable not set",
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if err := deleteServiceIDCmd(tt.args.cmd, tt.args.args); (err != nil) != tt.wantErr {
-				t.Errorf("deleteServiceIDCmd() error = %v, wantErr %v", err, tt.wantErr)
+			err := deleteServiceIDCmd(tt.args.cmd, tt.args.args)
+			if tt.expectError == "" {
+				assert.NoError(t, err)
+			} else {
+				assert.Containsf(t, err.Error(), tt.expectError, "expected error containing %q, got %s", tt.expectError, err)
 			}
 		})
 	}
@@ -47,7 +51,7 @@ func Test_deleteServiceIDs(t *testing.T) {
 		mockIBMCloudClient func(mockCtrl *gomock.Controller) *mockibmcloud.MockClient
 		setup              func(*testing.T) string
 		force              bool
-		wantErr            bool
+		expectError        string
 	}{
 		{
 			// Tests the delete-service-id command without any CredReqs present in the credentials-requests-dir directory.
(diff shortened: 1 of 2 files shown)


pkg/cmd/provisioning/ibmcloud/delete_service_id_test.go
```diff
@@ -0,0 +1,213 @@
...
+		{
+			name:    "CreateSharedSecretsCmd with unset API key environment variable should fail",
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := deleteServiceIDCmd(tt.args.cmd, tt.args.args); (err != nil) != tt.wantErr {
+				t.Errorf("deleteServiceIDCmd() error = %v, wantErr %v", err, tt.wantErr)
+			}
```
joelddiaz, line 39: can we reflow this logic so that the asserts kind of self-document what is happening/expected?
```suggestion
			err := deleteServiceIDCmd(tt.args.cmd, tt.args.args)
			if tt.wantErr {
			    assert.Error(err, ...)
			} else {
			    assert.NoError(err, ...)
			}
```

pkg/cmd/provisioning/ibmcloud/delete_service_id_test.go
```diff
@@ -0,0 +1,213 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			defer mockCtrl.Finish()
+
+			mockIBMCloudClient := tt.mockIBMCloudClient(mockCtrl)
+
+			credReqDir := tt.setup(t)
+			defer os.RemoveAll(credReqDir)
+
+			if err := deleteServiceIDs(mockIBMCloudClient, "1234", "name", credReqDir, tt.force); (err != nil) != tt.wantErr {
```
joelddiaz, line 190: Similar comment to the other file. Can this flow be written so that the assert.Error()/NoError() functions are used to show exactly what we're testing.

pkg/cmd/provisioning/ibmcloud/delete_service_id_test.go
```diff
@@ -0,0 +1,213 @@
...
+				t.Errorf("deleteServiceIDs() error = %v, wantErr %v", err, tt.wantErr)
+			}
+		})
+	}
+}
+
+func mockListServiceIDWithSameName(client *mockibmcloud.MockClient, name string, count int, fail bool) {
+	var err error
+	if fail {
+		err = fmt.Errorf("failed to get ListServiceID")
```
joelddiaz, line 200: Can the mocked error more closely mimic the actual error types returned by IBMCloud, or is this generic error a fair representation of the kind of error response we would get?

joelddiaz (commented):
Looks good. Just some minor feedback :+1:

pkg/cmd/provisioning/ibmcloud/delete_service_id_test.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+}
+
+func Test_deleteServiceIDs(t *testing.T) {
+	tests := []struct {
+		name               string
+		mockIBMCloudClient func(mockCtrl *gomock.Controller) *mockibmcloud.MockClient
+		setup              func(*testing.T) string
+		force              bool
+		wantErr            bool
```
akhil-rane, line 50: nit: I would name this variable as `expectErr` or `expectError`, consistent with the naming elsewhere in the CCO codebase

pkg/cmd/provisioning/ibmcloud/delete_service_id_test.go
akhil-rane: agreed, this is more readable

akhil-rane (commented):
super minor nit. looks good otherwise

pkg/cmd/provisioning/ibmcloud/delete_service_id_test.go
mkumatag: All credit goes to goland autogenerate feature :) I understand the expectation, I will enhance the code accordingly.

pkg/cmd/provisioning/ibmcloud/delete_service_id_test.go
mkumatag: It won't make a lot of difference because we aren't comparing any errors and failing if any non-nil error returns, having that said, I will reuse some error from the sdk.

mkumatag:
@joelddiaz @akhil-rane fixed the review comments, ptal latest commit

joelddiaz (commented):
/lgtm

joelddiaz:
just adding test cases for already committed code, no effect for qe, docs, or px
/label docs-approved
/label qe-approved
/label px-approved

PR #392, opened 2021-10-01
Defensive nil check in gcp actuator

2uasimojo:
A nil check was redundant (there was no way for the var to be nil) but done after the pointer in question would have already been dereferenced. Move it to right after the assignment, to guard against a future where it *is* possible to be nil.

[CCO-95](https://issues.redhat.com/browse/CCO-95)


akhil-rane (commented):
Nice!

/lgtm

akhil-rane:
@joelddiaz needs approve label from you

PR #398, opened 2021-10-08
Bug 2039085: Exclude operator config from ibm-cloud-managed profile

csrwng:
The cloud cred operator does not run for ibm-cloud-managed or hypershift. This removes the annotation that includes the config CR in that profile.

### DIFF
--- a/manifests/01-operator-config.yaml
+++ b/manifests/01-operator-config.yaml
@@ -3,7 +3,6 @@ kind: CloudCredential
 metadata:
   name: cluster
   annotations:
-    include.release.openshift.io/ibm-cloud-managed: "true"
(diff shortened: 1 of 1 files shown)


joelddiaz (commented):
/lgtm

joelddiaz:
@csrwng do you have a Jira or BZ for this? This repo is under this new "no feature freeze" process, and it requires either a BZ with QE signoff or for non-BZs signoff from QE, Docs, and PX.

joelddiaz:
@csrwng do we have a Jira to link this to so that docs / qe / product experience can all review and sign off? (the CCO repo needs all these sign-offs before something can merge)

PR #403, opened 2021-10-18
unit test cases for the refresh-keys

mkumatag:
Unit testcases for the refresh-keys subcommand

### DIFF
--- a/pkg/cmd/provisioning/ibmcloud/create_service_id_test.go
+++ b/pkg/cmd/provisioning/ibmcloud/create_service_id_test.go
@@ -644,7 +644,7 @@ func mockListResourceGroups(client *mockibmcloud.MockClient, resourceGroupExist,
 	client.EXPECT().ListResourceGroups(gomock.Any()).Return(list, nil, err).Times(1)
 }
 
-func mockListServiceID(client *mockibmcloud.MockClient, namePrefix string, count int, fail bool) {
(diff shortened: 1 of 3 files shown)


pkg/cmd/provisioning/ibmcloud/create_service_id_test.go
```diff
@@ -644,7 +644,7 @@ func mockListResourceGroups(client *mockibmcloud.MockClient, resourceGroupExist,
 	client.EXPECT().ListResourceGroups(gomock.Any()).Return(list, nil, err).Times(1)
 }
 
-func mockListServiceID(client *mockibmcloud.MockClient, namePrefix string, count int, fail bool) {
+func mockListServiceIDTimes(client *mockibmcloud.MockClient, namePrefix string, count int, fail bool, times int) {
```
joelddiaz, line 647: The var names 'count' and 'times' are a bit close in meaning. Can we rename 'count' to something like 'numSvcIDs' (or just something that makes the difference between 'count' and 'times' more immediately clear?

mkumatag: done, renamed to `numSvcIDs`

akhil-rane:
> @joelddiaz @akhil-rane wondering if we can add all the remaining tags because this is just a ut and doesnt impact any functionality..

Sure!

/docs-approved
/qe-approved
/px-approved

PR #407, opened 2021-10-29
internalcontroller not needed since controller-runtime 0.6.2

joelddiaz:
Ever since the upgrade to controller-runtime 0.6.2 in
https://github.com/openshift/cloud-credential-operator/pull/232, we have
had the ability to slip in our own rateLimitter when building the
controller w/o needing to resort to our custom/copied internalcontroller.

Make use of this new ability of controller-runtime and retire our copy
of the internal controller implementation from controller-runtime.

This should make upgrading to future versions of controller-runtime a
bit easier.

xref: https://issues.redhat.com/browse/CCO-150


2uasimojo:
Oh, how lovely.

Just needs a `go mod tidy` I believe.

joelddiaz:
/assign @lwan-wanglin @jeana-redhat @sferich888 
This should be a no-op in terms of features, functionality, UX as we are just starting to use some pre-existing features of our vendored version of controller-runtime so that we can drop some unnecessary copied code from older versions of controller-runtime.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -112,27 +112,18 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-		Config:                  mgr.GetConfig(),
-		Scheme:                  mgr.GetScheme(),
-		Client:                  mgr.GetClient(),
-		Recorder:                mgr.GetEventRecorderFor(name),
-		Queue:                   workqueue.NewNamedRateLimitingQueue(rateLimiter, name),
-		MaxConcurrentReconciles: 1,
-		Name:                    name,
-	}
-
-	if err := mgr.Add(c); err != nil {
```
joelddiaz, line 128: Needed to remove this mgr.Add() as it is already called in the controller.New() codepath right above. Without this CCO errors early on startup with `time="2021-11-01T16:28:09Z" level=fatal msg="unable to run the manager" error="controller was started more than once. This is likely to be caused by being added to a manager multiple times"`

2uasimojo:
/retest

timeout

/lgtm

still

lwan-wanglin:
do some tests, no regression. https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/Runner/209292/console

/label qe-approved

joelddiaz:
@2uasimojo PTAL, had to rebase due to other code changes. no functional changes from last time

2uasimojo:
Will restamp once CI is clear.

joelddiaz:
Just internal bookkeeping changes with no change in product behaivor.
/label px-approved

PR #411, opened 2021-11-03
Alibaba: add Alibaba Cloud manual mode

dongchen126:
Support the Alibaba Cloud platform in manual mode
phase 1: Add 'AlibabaCloudProviderSpec' and 'AlibabaCloudProviderStatus' 
phase 2: 'ccoctl' support（ongoing)


pkg/apis/cloudcredential/v1/types_alibabacloud.go
```diff
@@ -0,0 +1,56 @@
...
+	Condition Condition `json:"condition,omitempty"`
+}
+
+// Condition - map of condition types, with associated key - value mapping
+// +k8s:deepcopy-gen=false
+type Condition map[string]ConditionKeyValue
+
+// ConditionKeyValue - mapping of values for the chosen type
+// +k8s:deepcopy-gen=false
+type ConditionKeyValue map[string]interface{}
```
joelddiaz, line 50: Using type interface{} complicates generating deep-copy functions. See https://github.com/openshift/cloud-credential-operator/blob/master/pkg/apis/cloudcredential/v1/aws_manual.deepcopy.go and https://github.com/openshift/api/pull/683#pullrequestreview-447058271 .

Just because we made this mistake for the AWS cloud, does not mean we should repeat it for Alibaba Cloud. Please specify the exact values this field can hold, or just remove the ability to specify conditions as part of each `Entry`.

If we do in fact expect to have components specify static Conditions as part of their permissions lists, I'd like to see some examples.

bd233: Thanks for your review.
At present, Condition is not needed in Alibaba Cloud's credsRequest, so I removed it.

dongchen126:
Closed this PR due to will use new PR #412

PR #415, opened 2021-11-10
Fix small log typo for upgradable check

jharrington22:


akhil-rane (commented):
Just a typo fix in the log message.

/label px-approved
/label docs-approved
/label qe-approved

PR #417, opened 2021-11-15
Bug 2023832: pkg/operator/status: Only bump lastTransitionTime on status changes

wking:
From [the property docs][1]:

> lastTransitionTime is the time of the last update to the current status property.

Restricting the timestamp bumps to the status transitions allows consumers to answer "how long have I been `Degraded=True`?" and similar, which is often more useful than being able to answer "how long have I been `Degraded=True` with exactly this `reason` and exactly this `message`?".  Messages are [only for human consumption][2] anyway, so there should be very few users that care about "... and exactly this `message`".  For folks who do care about reasons, [the `cluster_operator_conditions` metric][3] provides access with Prometheus-scrape precision.

[1]: https://github.com/openshift/api/blob/be1be0e89115702f8b508d351c4f5c9a16e5ae95/config/v1/types_cluster_operator.go#L126
[2]: https://github.com/openshift/api/blob/be1be0e89115702f8b508d351c4f5c9a16e5ae95/config/v1/types_cluster_operator.go#L136
[3]: https://github.com/openshift/cluster-version-operator/blob/887aa47093d740ce827e3046848d4f0c1aca374f/pkg/cvo/metrics.go#L84-L86

### DIFF
--- a/pkg/operator/status/status_controller.go
+++ b/pkg/operator/status/status_controller.go
@@ -344,23 +344,12 @@ func setLastTransitionTime(oldConditions []configv1.ClusterOperatorStatusConditi
 	for i := range newConditions {
 		newCondition := &newConditions[i]
 		oldCondition, _ := findClusterOperatorCondition(oldConditions, newCondition.Type)
-		if oldCondition == nil || !conditionEqual(*oldCondition, *newCondition) {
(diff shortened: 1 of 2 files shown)


joelddiaz (commented):
/lgtm

PR #420, opened 2021-11-17
Generate ccoctl binary for all the supported platforms

mkumatag:
This PR will allow docker image to ship the ccoctl client binary for all the supported platform, post this user can download the relevant binary and use.


joelddiaz:
After some internal discussions, it was decided that the right approach is to have the MacOS binaries be signed and delivered alongside things like the `oc` and `openshift-install` binaries. The first step in that process is following along with preexisting patterns (including building for Windows in addition to MacOS). That first step can be seen/followed with openshift/cloud-credential-operator#431 .

PR #436, opened 2021-12-21
Bug 2035328: Make ccoctl ignore cred requests marked for deletion

akhil-rane:
Manifests with annotation "release.openshift.io/delete" set to "true"
are marked for in-cluster deletion. Ccoctl should ignore these while
creating cloud infra. We also log a message so that user is aware
why a CredentialsRequest is being ignored.

x-ref: https://issues.redhat.com/browse/CCO-175


pkg/cmd/provisioning/aws/create-iam-roles_test.go
```diff
@@ -118,6 +158,40 @@ func TestIAMRoles(t *testing.T) {
 				assert.Equal(t, 1, provisioning.CountNonDirectoryFiles(files), "Should be exactly 1 secret in manifestsDir for one CredReq")
 			},
 		},
+		{
+			name:         "Ignore CredReq marked for deletion",
```
akhil-rane, line 162: I have included UT for only aws, but it covers the code path so we should be good.

joelddiaz (commented):
/lgtm

PR #439, opened 2021-12-24
Bug 2037903: Alibaba Cloud not required credReqDir in command delete-ram-users

### LINKED ISSUE #437: Alibaba Cloud: delete-ram-user requires the credentials-requests
Summary:

When running the removal of ram users the `ccoctl` command requires the original `--credentials-requests` directory. This might be an issue when customers remove them after a cluster install. Having to preserve the original release payload or preserving the credentials requests becomes cumbersome.

Example:
```
~/go/src/github.com/openshift/cloud-credential-operator/ccoctl alibabacloud  delete-ram-users --region us-east-1 --name test-nsrlt   --credentials-requests-dir ~/tmp/alibaba/crs
```

Would it be possible to update the `delete-ram-users` command to use only the `--name <cluster_id>` parameter data to remove credentials? This would greatly simplify the process of removing credentials from a customer's standpoint and for the CI.

cc @DahuK

DahuK:
Alibaba Cloud: support not required input credentials-requests-dir in command delete-ram-users
Fixes #437

### DIFF
--- a/pkg/cmd/provisioning/alibabacloud/create-ram-users.go
+++ b/pkg/cmd/provisioning/alibabacloud/create-ram-users.go
@@ -37,6 +37,8 @@ type: Opaque`
 	ramPolicyType = "Custom"
 	// autoRotateStrategy will delete the earliest inactive policy version
 	autoRotateStrategy = "DeleteOldestNonDefaultVersionWhenLimitExceeded"
+	// ccoctlResourceTagKeyPrefix is the prefix of the tag key applied to the Alibaba Cloud ram user created by ccoctl
+	ccoctlResourceTagKeyPrefix = "openshift.io/ccoctl"
 )
 
 var (
@@ -174,7 +176,7 @@ func createUser(client alibabacloud.Client, name string, credReq *credreqv1.Cred
 	userReq := ram.CreateCreateUserRequest()
 	userReq.UserName = shortName
 	userReq.DisplayName = displayName
-
+	userReq.Comments = fmt.Sprintf("%s/%s", ccoctlResourceTagKeyPrefix, name)
 	user, err := client.CreateUser(userReq)
 	if err != nil {
 		aErr, ok := err.(*alibabaerrors.ServerError)
--- a/pkg/cmd/provisioning/alibabacloud/create-ram-users.go
+++ b/pkg/cmd/provisioning/alibabacloud/create-ram-users.go
@@ -37,8 +37,8 @@ type: Opaque`
 	ramPolicyType = "Custom"
 	// autoRotateStrategy will delete the earliest inactive policy version
 	autoRotateStrategy = "DeleteOldestNonDefaultVersionWhenLimitExceeded"
-	// ccoctlResourceTagKeyPrefix is the prefix of the tag key applied to the Alibaba Cloud ram user created by ccoctl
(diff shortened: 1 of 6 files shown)


DahuK:
@kwoodson please help to review this, thanks a lot!

pkg/cmd/provisioning/alibabacloud/create-ram-users.go
```diff
@@ -37,6 +37,8 @@ type: Opaque`
 	ramPolicyType = "Custom"
 	// autoRotateStrategy will delete the earliest inactive policy version
 	autoRotateStrategy = "DeleteOldestNonDefaultVersionWhenLimitExceeded"
+	// ccoctlResourceTagKeyPrefix is the prefix of the tag key applied to the AWS resources created/shared by ccoctl
+	ccoctlResourceTagKeyPrefix = "openshift.io/ccoctl"
```
joelddiaz, line 41: The variable name indicates that this is related to tagging a resource. From looking at the code, this string is simply being used to place a comment/note on the RAM User details. Also, the comment for this variable talks about AWS.

pkg/cmd/provisioning/alibabacloud/delete-ram-users.go
```diff
@@ -89,15 +89,42 @@ func deleteComponentUser(client alibabacloud.Client, userName string) error {
 	return err
 }
 
+func getUsersToDelete(client alibabacloud.Client, name, credReqDir string) ([]string, error) {
+	usersToDelete := make([]string, 0)
+	if credReqDir != "" {
+		// Process directory
+		credRequests, err := provisioning.GetListOfCredentialsRequests(credReqDir)
```
joelddiaz, line 96: I thought the purpose of this PR was to not need the `credReqDir` to be provided in the `delete` subcommand?

pkg/cmd/provisioning/alibabacloud/create-ram-users.go
DahuK: fixed the comment, thanks for the finding! and the comment is used in delete command to find out which user should be removed if no credReqDir given

joelddiaz: This still calls the variable \*TagKeyPrefix\*. I think it would be more clear to remove any reference to "Tag" as AliCloud doesn't presently allow tagging RAM Users. Maybe something like `ccoctlResourcePrefix` or `ccoctlResourceCommentPrefix`?

pkg/cmd/provisioning/alibabacloud/delete-ram-users.go
joelddiaz: I'd rather just delete the `credReqDir` since finding the resource(s) by using the `--name` parameter should be sufficient for finding the resources that need to be cleaned up. And the `--name` parameter is already required, so we already have all the information needed to do a proper cleanup.

DahuK: ok, have removed the `credReqDir`

pkg/cmd/provisioning/alibabacloud/create-ram-users.go
DahuK: got it, change the variable to `ccoctlResourcePrefix`

joelddiaz (commented):
/lgtm

PR #441, opened 2022-01-12
Bug 2037331: Make ccoctl gracefully handle existing gcp workload identity resources

akhil-rane:
Currently, ccoctl for GCP errors out if there are existing resources 
for a given infra name. This PR handles this gracefully by making
behavior consistent with AWS. Added unit tests to ensure the right
behavior.

Also, ccoctl now fetches GCP service accounts based on name
and not on account ID. The account IDs have a random string as
suffix which makes it difficult to use them to fetch existing service
accounts.
 
x-ref: https://issues.redhat.com/browse/OCPBUGSM-39019

### DIFF
--- a/pkg/cmd/provisioning/gcp/create_service_accounts.go
+++ b/pkg/cmd/provisioning/gcp/create_service_accounts.go
@@ -13,8 +13,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	iamadminpb "google.golang.org/genproto/googleapis/iam/admin/v1"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 
 	credreqv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
 	"github.com/openshift/cloud-credential-operator/pkg/cmd/provisioning"
@@ -124,7 +122,6 @@ func createServiceAccount(ctx context.Context, client gcp.Client, name string, c
 	if err != nil {
 		return "", errors.Wrap(err, "Error generating service account name")
 	}
-	serviceAccountName = serviceAccountName[:len(serviceAccountName)-6] // chop off the trailing random chars
 
 	// Decode GCPProviderSpec
 	codec, err := credreqv1.NewCodec()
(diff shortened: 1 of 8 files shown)


pkg/cmd/provisioning/gcp/create_service_accounts.go
```diff
@@ -124,7 +122,6 @@ func createServiceAccount(ctx context.Context, client gcp.Client, name string, c
 	if err != nil {
 		return "", errors.Wrap(err, "Error generating service account name")
 	}
-	serviceAccountName = serviceAccountName[:len(serviceAccountName)-6] // chop off the trailing random chars
```
akhil-rane, line 127: This was picked up from the actuator code. The `serviceAccountName` does not have any random characters to chop off IIUC. I would prefer longer names in the display without any truncated characters. Any thoughts?

pkg/cmd/provisioning/gcp/create_workload_identity_pool.go
```diff
@@ -94,20 +96,36 @@ func createWorkloadIdentityPool(ctx context.Context, client gcp.Client, name, pr
...
+				_, err := client.CreateWorkloadIdentityPool(ctx, parentResourceForPool, name, pool)
+				if err != nil {
+					return errors.Wrapf(err, "failed to create workload identity pool %s", name)
+				}
+				log.Printf("Workload identity pool created with name %s", name)
+			} else {
+				return errors.Wrapf(err, "failed to check if there is existing workload identity pool %s", name)
+			}
+		} else {
+			log.Printf("Workload identity pool %s already exists and is owned by the user", name)
```
joelddiaz, line 127: How do we know that this identity pool is really owned by the user?

joelddiaz (commented):
just one small question, but otherwise :+1:

pkg/cmd/provisioning/gcp/create_workload_identity_pool.go
akhil-rane: I think this statement is wrong. This should be `pool is owned by the project` instead.

akhil-rane: Let me just reframe it to `log.Printf("Workload identity pool %s already exists", name)` to avoid confusion

joelddiaz (commented):
/lgtm

PR #442, opened 2022-01-13
add unsupported docs for migrating to STS

joelddiaz:
This is just for developers to (semi-)quickly move from a non-STS
cluster over to STS.

xref: https://issues.redhat.com/browse/CCO-58


docs/sts.md
```diff
@@ -80,6 +80,78 @@ The token is a projected ServiceAccount into the Pod, and is short lived for an
...
+```bash
+$ ./ccoctl aws create-iam-roles --output-dir ./newstscluster/ --name newstscluster --identity-provider-arn ARN_CREATED_FROM_CREATE_IDENTITY_PROVIDER_COMMAND --region us-east-2 --credentials-requests-dir ./cred-reqs/
+```
+
+13. Apply the Secrets generated by the above command:
+```bash
+$ find ./cred-reqs -iname "*yaml" -print0 | xargs -I {} -0 -t oc replace -f {}
+```
+
+14. At this point the cluster is using STS. The previously created IAM Users/credentials can be deleted as they are not being used. The "root" AWS creds Secret can also be removed (`oc delete -n kube-system aws-creds`).
```
akhil-rane, line 154:
```suggestion
14. At this point the cluster is using STS. The previously created IAM Users/credentials can be deleted as they are not being used. The "root" AWS creds Secret can also be removed (`oc delete -n secret kube-system aws-creds`).
```

docs/sts.md
```diff
@@ -80,6 +80,78 @@ The token is a projected ServiceAccount into the Pod, and is short lived for an
...
+$ cat newstscluster/manifests/cluster-authentication-02-config.yaml | awk '/serviceAccountIssuer/ { print $2 }'
+$ oc edit authentication cluster
+```
+
+6. Wait for the kube-apiserver pods to be updated with the new config:
+```bash
+$ oc get pods -n openshift-kube-apiserver | grep kube-apiserver
+```
+
+7. Restart all pods in the cluster (because all ServiceAccounts need to be refreshed after updating the serviceAccountIssuer field):
```
akhil-rane, line 119: Some mention here that this step takes a significant amount of time.

akhil-rane (commented):
Some nits, but this works

akhil-rane (commented):
/lgtm

joelddiaz:
this is explicitly not supported and is just some bare-bones instructions for developers looking for a non-cluster-destructive way to get from notSTS to STS
/label docs-approved
/label qe-approved
/label px-approved

PR #443, opened 2022-01-14
cleanup CCO from ibm-cloud-managed

joelddiaz:
CCO is not used in the ibm-cloud-managed profile. Mark all these
previously installed resources as needing to be removed in the
ibm-cloud-managed profile, and remove the ibm-cloud-managed annotation
so they don't get installed again.

xref: https://issues.redhat.com/browse/CCO-177

### DIFF
--- a/manifests/00-crd.yaml
+++ b/manifests/00-crd.yaml
@@ -2,7 +2,6 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
   name: credentialsrequests.cloudcredential.openshift.io
 spec:
--- a/manifests/ibm-cloud-cleanup.yaml
+++ b/manifests/ibm-cloud-cleanup.yaml
@@ -1,0 +1,116 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:openshift:cloud-credential-operator:cluster-reader
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    release.openshift.io/delete: "true"
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    release.openshift.io/delete: "true"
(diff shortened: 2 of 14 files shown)


joelddiaz:
/hold
wait for openshift/cluster-image-registry-operator#741 , openshift/cluster-ingress-operator#695 , openshift/cluster-network-operator#1280 , and openshift/cluster-storage-operator#253

csrwng:
I see this error in the CVO when running this release:
```
F0117 19:59:18.483477       1 start.go:29] error: the local release contents are invalid - no current version can be determined from disk: Error loading manifests from /var/payload: parse 0000_50_cloud-credential-operator_ibm-cloud-cleanup.yaml: error parsing: error unmarshaling JSON: while decoding JSON: Resource with fields Group: "" Kind: "Service" Name: "" must contain kubernetes required fields kind and name
```

manifests/00-crd.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    include.release.openshift.io/ibm-cloud-managed: "true"
```
csrwng, line 5: @joelddiaz if we're not removing cloud cred requests from all repos before 4.10, then it doesn't make sense to exclude the CRD just yet

manifests/ibm-cloud-cleanup.yaml
```diff
@@ -0,0 +1,116 @@
...
+kind: Role
+metadata:
+  name: cloud-credential-operator-role
+  namespace: openshift-cloud-credential-operator
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    release.openshift.io/delete: "true"
+---
+apiVersion: v1
+kind: Service
```
csrwng, line 94: needs name/namespace

manifests/00-crd.yaml
joelddiaz: Yes, I was thinking we'd block this PR until all the depending PRs merge (which of course means we're looking at this stuff landing for 4.11).

manifests/ibm-cloud-cleanup.yaml
joelddiaz: fixed. also added an additional Service that may have long ago been part of CCO and is presently marked for deletion (named 'controller-manager-service') but cover it under the ibm-cloud-managed list of things to delete in the event that it was ever installed on that profile.

PR #445, opened 2022-01-19
support for appending the user tags provided in the Infrastructure resource

sairameshv:
This piece of code adds the support to append the extra tags provided by the user during the installation time.
Reference: https://github.com/openshift/enhancements/blob/master/enhancements/api-review/custom-tags-aws.md


pkg/aws/actuator/actuator.go
```diff
@@ -840,7 +840,30 @@ func (a *AWSActuator) tagUser(logger log.FieldLogger, awsClient minteraws.Client
...
+		logger.WithError(err).Error("unable to get the infrastructure object")
+	} else {
+		// The ResourceTags field would be migrated from the Status field to the Spec field.
+		// Till the field is deprecated, the ResourceTags present in the Spec field takes the priority
+		var userTags []configv1.AWSResourceTag
+		if len(infra.Spec.PlatformSpec.AWS.ResourceTags) == 0 {
+			userTags = infra.Status.PlatformStatus.AWS.ResourceTags
+		} else {
+			userTags = infra.Spec.PlatformSpec.AWS.ResourceTags
+		}
```
joelddiaz, line 855: Why are we referencing both status and spec? What is the relationship between status and spec. The comment indicates that information would migrate "from the Status field to the Spec field" which sounds like the reverse flow for how information is typically updated in kubernetes objects.

joelddiaz (commented):
The codeflow as presently written means that this tagUser() function only gets called if the user doesn't have the expected tags https://github.com/openshift/cloud-credential-operator/blob/47b9d0af050709f4c9a67b54bc37f3984bf64fc3/pkg/aws/actuator/actuator.go#L548 

In a steady-state cluster where the CCO has already created users and tagged them: if a user then updates infrastructure.Spec.PlatformSpec.AWS.ResourceTags, the CCO will never apply the tags because the userHasExpectedTags() function doesn't reference the contents of the Infrastructure CR to know whether the user has the expected tags applied.

sairameshv:
> The codeflow as presently written means that this tagUser() function only gets called if the user doesn't have the expected tags
> 
> https://github.com/openshift/cloud-credential-operator/blob/47b9d0af050709f4c9a67b54bc37f3984bf64fc3/pkg/aws/actuator/actuator.go#L548
> 
> In a steady-state cluster where the CCO has already created users and tagged them: if a user then updates infrastructure.Spec.PlatformSpec.AWS.ResourceTags, the CCO will never apply the tags because the userHasExpectedTags() function doesn't reference the contents of the Infrastructure CR to know whether the user has the expected tags applied.

Hey @joelddiaz , thanks for the review comments :) , This PR is still a draft version, I'm yet to completely understand the flow and add the required code. I'm trying to deploy the cluster with this PR using cluster bot to understand the behavior. I would like to remind you for the review once my changes are ready.

pkg/aws/actuator/actuator.go
sairameshv: I actually meant to convey the following 

> Saving in the status field means it's not included in Velero backups, which is a crucial feature for customers and Day2.
> Thus the status.resourceTags field should be deprecated in favour of a newly created spec.resourceTags with the same content. The installer should only populate the spec, consumers of the infrastructure CRD must favour the spec over the status definition if both are supplied, otherwise the status should be honored and a warning shall be issued.

which is part of this [RFE](https://issues.redhat.com/browse/RFE-2012).

Modified the code to fetch the info from the Spec.

sairameshv:
/assign @joelddiaz

Hey @joelddiaz , 
Could you review the PR now and provide the review comments?

Thanks,
Ramesh

pkg/aws/actuator/actuator.go
joelddiaz: In what situation would status be populated, but not spec? I guess what I might really be asking is: If this repo today doesn't support reading from status, and spec is the "one true location for tag info", why bother reading status at all?

pkg/aws/actuator/actuator.go
```diff
@@ -651,12 +651,25 @@ func (a *AWSActuator) awsPolicyEqualsDesiredPolicy(desiredUserPolicy string, aws
 	return true, nil
 }
 
-func userHasExpectedTags(logger log.FieldLogger, user *iam.User, infraName, clusterUUID string) bool {
+func userHasExpectedTags(logger log.FieldLogger, user *iam.User, infraName, clusterUUID string, awsClient client.Client) bool {
```
joelddiaz, line 654:
```suggestion
func userHasExpectedTags(logger log.FieldLogger, user *iam.User, infraName, clusterUUID string, kubeClient client.Client) bool {
```

pkg/aws/actuator/actuator.go
joelddiaz: Nevermind, it looks like the latest version of the PR only reads from spec.

pkg/aws/actuator/actuator.go
```diff
@@ -651,12 +651,25 @@ func (a *AWSActuator) awsPolicyEqualsDesiredPolicy(desiredUserPolicy string, aws
...
 		return false
 	}
 
+	// Check if the user has the expected tags from the Infrastructure resource
+	infra, _ := utils.GetInfrastructure(awsClient)
+	if infra != nil {
+		if len(infra.Spec.PlatformSpec.AWS.ResourceTags) != 0 {
+			for _, userTag := range infra.Spec.PlatformSpec.AWS.ResourceTags {
+				if !userHasTag(user, userTag.Key, userTag.Value) {
+					log.Warnf("user missing tag: %s=%s", userTag.Key, userTag.Value)
```
joelddiaz, line 666: Is a missing tag an unexpected situation? If not, then perhaps a log.Info/Debug() would be more appropriate.

pkg/aws/actuator/actuator.go
sairameshv: Noted, Thanks

pkg/aws/actuator/actuator.go
```diff
@@ -840,7 +853,22 @@ func (a *AWSActuator) tagUser(logger log.FieldLogger, awsClient minteraws.Client
 		})
 	}
 
-	_, err := awsClient.TagUser(&iam.TagUserInput{
+	// appending the tags present in the Infrastructure CR
+	infra, err := utils.GetInfrastructure(a.Client)
+	if err != nil {
+		logger.WithError(err).Error("unable to get the infrastructure object")
```
joelddiaz, line 859: Why wouldn't the function just return an error here. Isn't the inability to read the Infrastructure resource a problem if this operator is supposed to be tagging things as specified in the Infrastructure?

pkg/aws/actuator/actuator.go
sairameshv: > In what situation would status be populated, but not spec? I guess what I might really be asking is: If this repo today doesn't support reading from status, and spec is the "one true location for tag info", why bother reading status at all?

As of today, all the other operators supporting this feature of user tags are reading from the Status field of the Infrastructure resource. Initially, I added code to read from any of the field and if both are present, the Spec takes the priority according to the RFE. Anyway, if the dependent PR mentioned above is merged, the Infrastructure resource would have the resourceTags in the spec field.

pkg/aws/actuator/actuator.go
sairameshv: Agreed.

pkg/aws/actuator/actuator.go
sairameshv: I was in a doubt here if this has to return an error or continue tagging the user with the other tags and log the error (as implemented above in an if/else statement).
Would make the required change. Thanks for pointing out.

joelddiaz: There is a bit of a potential problem if unable to retrieve the Infrastructure for multiple/repeated attempts, as the IAM User would never be tagged after creation, and then when someone goes to destroy the cluster, the IAM User will not be found because of the missing infrastructure tag. 
Perhaps way up in the syncMint() function where the LoadInfrastructureName() is called, you can also fetch the full Infrastructure resource and error before any AWS API calls are made https://github.com/openshift/cloud-credential-operator/blob/6ac05b473cdaf2b116a6432055dd4ed22aad6759/pkg/aws/actuator/actuator.go#L471

sairameshv: Addressed and added the required code with a small refactoring :)

pkg/aws/actuator/actuator.go
```diff
@@ -545,12 +547,12 @@ func (a *AWSActuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequ
...
 
 	// Check if the user has the expected tags:
-	if !userHasExpectedTags(logger, userOut, infraName, string(clusterUUID)) {
+	if !userHasExpectedTags(logger, userOut, string(clusterUUID), infra) {
 		if rootAWSClient == nil {
 			return fmt.Errorf("no root AWS client available, cred secret may not exist: %s/%s", constants.CloudCredSecretNamespace, constants.AWSCloudCredSecretName)
 		}
 
-		err = a.tagUser(logger, rootAWSClient, awsStatus.User, infraName, string(clusterUUID))
+		err = a.tagUser(logger, rootAWSClient, awsStatus.User, infra.Status.InfrastructureName, string(clusterUUID))
```
joelddiaz, line 555: pass the full Infrastructure to tagUser() so that tagUser() doesn't have to fetch it and potentially error due to that reason.

sairameshv: Sorry, Missed it. Noted.

joelddiaz (commented):
:+1: go ahead an squash the code changes down to one (leave the vendor as a separate commit)

akhil-rane (commented):
I do not see any watch on Infrastructure resource for credentials_request controller. I think we need to reconcile the credentials requests if changes to the infrastructure resource are detected.

pkg/aws/actuator/actuator.go
```diff
@@ -840,6 +854,16 @@ func (a *AWSActuator) tagUser(logger log.FieldLogger, awsClient minteraws.Client
...
+	// appending the tags present in the Infrastructure CR
+	if len(infra.Spec.PlatformSpec.AWS.ResourceTags) != 0 {
+		logger.WithField("userTags", infra.Spec.PlatformSpec.AWS.ResourceTags).Info("tagging the user with the tags present in the infrastructure object")
+	}
+	for _, userTag := range infra.Spec.PlatformSpec.AWS.ResourceTags {
+		tags = append(tags, &iam.Tag{
+			Key:   aws.String(userTag.Key),
+			Value: aws.String(userTag.Value),
+		})
+	}
```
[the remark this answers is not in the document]
sairameshv: Updated the code

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -131,6 +131,12 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for the changes happening to Infrastructure Resource
+	err = c.Watch(&source.Kind{Type: &configv1.Infrastructure{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
```
akhil-rane, line 138: I think sync should happen only if resource tags in the `configv1.Infrastructure` are present or updated. Do we care about the rest of the object?

joelddiaz: Not just watching Infrastructure, but also queueing up all the CredentialsRequests so that they get re-reconciled.

sairameshv: > I think sync should happen only if resource tags in the `configv1.Infrastructure` are present on create event or modified on update event. Do we care about the rest of the object?

Not required really, I would add some predicates to check on the changes done to the ResourceTags

sairameshv: > Not just watching Infrastructure, but also queueing up all the CredentialsRequests so that they get re-reconciled.

Sure, I would address this

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -227,6 +227,24 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	// predicate functions to filter the events based on the AWS resourceTags presence.
+	infraResourcePredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return isAWSResourceTagsPresent(e.Object)
+		},
+		DeleteFunc: func(event.DeleteEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return isAWSResourceTagsPresent(e.ObjectNew)
```
arjunrn, line 239: This will work but is not an effective predicate. If you should check if the tags have changed between the new and old object.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -336,6 +354,23 @@ func isAdminCredSecret(namespace, secretName string) bool {
 	return false
 }
 
+// isAWSResourceTagsPresent returns true if the AWS resourceTags are present in the Infrastructure resource
+func isAWSResourceTagsPresent(event client.Object) bool {
```
arjunrn, line 358: Just a personal preference but `hasResourceTags` sounds better to me.

sairameshv: Updated

pkg/operator/credentialsrequest/credentialsrequest_controller.go
sairameshv: Addressed

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -227,6 +227,24 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		CreateFunc: func(e event.CreateEvent) bool {
+			return hasResourceTags(e.Object)
+		},
+		DeleteFunc: func(event.DeleteEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return areTagsUpdated(e.ObjectOld, e.ObjectNew)
+		},
+	}
```
joelddiaz, line 241: How often do we expect the Infrastructure resource to be changing? I ask because this is introducing a not-insubstantial amount of AWS-specific code into the generic CredentialsRequest reconciliation file(s). What's the harm in saying "Infrastructure changed...let's requeue everything and make sure any changes to Infrastructure are reflected in the resources this controller is responsible for"?

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -227,6 +227,24 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	// Watch for the changes happening to Infrastructure Resource
+	err = c.Watch(&source.Kind{Type: &configv1.Infrastructure{}}, allCredRequestsMapFn, infraResourcePredicate)
```
joelddiaz, line 243: While requeueing all the CredentialsRequest resources makes sense, from looking at the code flow in this controller, it will end up doing nothing because a change to the Infrastructure resource is not something that would get passed all these checks: https://github.com/openshift/cloud-credential-operator/blob/master/pkg/operator/credentialsrequest/credentialsrequest_controller.go#L529-L540

pkg/operator/credentialsrequest/credentialsrequest_controller.go
sairameshv: Hi @joelddiaz ,
I couldn't find an answer about the frequency of the changes happening to the Infrastructure resource. Hence, added a predicate to make sure the Reconcile function gets triggered only during the change observed in the resource tags and not for any other change in the Infrastructure resource.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
sairameshv: Yes @joelddiaz , I observed this and the tags are getting updated whenever the next sync happens.
I discussed this with my team members and we have come up with two ways of resolving this problem.
1. Addition of a new field "ResourceTags" in the "CredentialsRequest" resource's status field, compare these tags with the Infrastructure's ResourceTags whenever the Reconcile function is called.
2. Addition of a new controller to monitor the Infrastructure resource, corresponding reconcile function consists of listing the CredentialaRequest resources, calling the respective actuators etc. similar to that of happening in the existing Reconcile function [here](https://github.com/openshift/cloud-credential-operator/blob/master/pkg/operator/credentialsrequest/credentialsrequest_controller.go#L358).

Please let me know your thoughts on these and also, do you have any suggestions apart from the above mentioned proposals?

sairameshv: @akhil-rane Could you have a look at this once you are available?
Thanks

akhil-rane: @sairameshv option (1) sounds good to me. Instead of having `ResouceTags` field under `CredenetialsRequest.Status`, we should have it under `CredenetialsRequest.Status.ProviderStatus` as this field is specific to cloud provider AWS. Does that make sense?

sairameshv: @akhil-rane, I think that's a good idea but the only concern is that the code is becoming very much specific to AWS (Decoding the Provider specific status, comparing with provider tags in Infra object, etc.) as pointed out by @joelddiaz here in this [comment](https://github.com/openshift/cloud-credential-operator/pull/445#discussion_r799484205). Also, in future, if we are to update the tags on the Azure resources, it would again require changes specific to Azure in the generic credentialsrequest controller code. Don't you think it's better to define the `ResourceTags` field in the `CredenetialsRequest.Status` itself?

Moreover, just FYI, the [enhancement proposal](https://github.com/openshift/enhancements/pull/1051) is not insisting on any time limit under which the tags must be updated. Hence, the existing code works fine as of now with a maximum delay of 1hr updating the tags on the AWS resources.

akhil-rane: > Moreover, just FYI, the https://github.com/openshift/enhancements/pull/1051 is not insisting on any time limit under which the tags must be updated. Hence, the existing code works fine as of now with a maximum delay of 1hr updating the tags on the AWS resources.

Delaying the reconcile by 1 hour would not be a good user experience so I think we should try and update tags on aws resources immediately.

> I think that's a good idea but the only concern is that the code is becoming very much specific to AWS (Decoding the Provider specific status, comparing with provider tags in Infra object, etc.) as pointed out by @joelddiaz here in this https://github.com/openshift/cloud-credential-operator/pull/445#discussion_r799484205. Also, in future, if we are to update the tags on the Azure resources, it would again require changes specific to Azure in the generic credentialsrequest controller code. Don't you think it's better to define the ResourceTags field in the CredenetialsRequest.Status itself?

I agree with you on this. This would introduce more AWS specific code in credentials_request controller. But again we cannot store Tags on Credentials request status. Tags are very specific to AWS. Azure might have tags but other cloud providers might not. Can we have a field called `LastSyncInfrastructureResourceVersion` (something similar to [this](https://github.com/openshift/cloud-credential-operator/blob/ed0612946f4b77a18aa1e631b83abef9d856697a/pkg/apis/cloudcredential/v1/types_credentialsrequest.go#L82-L87))  in the CredentialsRequest.Status and force reconcile if last synced version != current version for infrastructure resource. What do you think?

sairameshv: Yes @akhil-rane, that looks to be a good idea. I would add the required changes and push the code.

akhil-rane: Thanks

pkg/aws/actuator/actuator.go
```diff
@@ -545,12 +547,12 @@ func (a *AWSActuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequ
...
 
 	// Check if the user has the expected tags:
-	if !userHasExpectedTags(logger, userOut, infraName, string(clusterUUID)) {
+	if !userHasExpectedTags(logger, userOut, string(clusterUUID), infra) {
 		if rootAWSClient == nil {
 			return fmt.Errorf("no root AWS client available, cred secret may not exist: %s/%s", constants.CloudCredSecretNamespace, constants.AWSCloudCredSecretName)
 		}
 
-		err = a.tagUser(logger, rootAWSClient, awsStatus.User, infraName, string(clusterUUID))
+		err = a.tagUser(logger, rootAWSClient, awsStatus.User, string(clusterUUID), infra)
```
akhil-rane, line 555: Will this work when certain resource tags are removed from Infrastructure resource. Do we expect them to be removed from cloud resources too?

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -336,6 +354,34 @@ func isAdminCredSecret(namespace, secretName string) bool {
...
+func hasResourceTags(event client.Object) bool {
+	switch infra := event.(type) {
+	case *configv1.Infrastructure:
+		if infra != nil {
+			if infra.Spec.PlatformSpec.AWS != nil {
+				if len(infra.Spec.PlatformSpec.AWS.ResourceTags) != 0 {
+					return true
+				}
+			}
+		}
```
akhil-rane, line 367:
```suggestion
		if infra != nil && infra.Spec.PlatformSpec.AWS != nil && len(infra.Spec.PlatformSpec.AWS.ResourceTags) != 0 {
					return true
		}
```

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -336,6 +354,34 @@ func isAdminCredSecret(namespace, secretName string) bool {
...
+		return false
+	}
+	return false
+}
+
+// areTagsUpdated validates and returns a true if the resourceTags are updated in the new event
+func areTagsUpdated(oldEvent, newEvent client.Object) bool {
+	if !hasResourceTags(newEvent) {
+		return false
+	}
```
akhil-rane, line 378: If a user removes the tags later as part of the update, do we want to remove them from aws user resource as well?

pkg/aws/actuator/actuator.go
```diff
@@ -651,14 +653,28 @@ func (a *AWSActuator) awsPolicyEqualsDesiredPolicy(desiredUserPolicy string, aws
...
+			if len(infraResource.Spec.PlatformSpec.AWS.ResourceTags) != 0 {
+				for _, userTag := range infraResource.Spec.PlatformSpec.AWS.ResourceTags {
+					if !userHasTag(user, userTag.Key, userTag.Value) {
+						log.Infof("user missing tag: %s=%s", userTag.Key, userTag.Value)
+						return false
+					}
+				}
+			}
+		}
+	}
```
akhil-rane, line 674:
```suggestion
	if infraResource != nil && infraResource.Spec.PlatformSpec.AWS != nil && len(infraResource.Spec.PlatformSpec.AWS.ResourceTags) != 0 {
				for _, userTag := range infraResource.Spec.PlatformSpec.AWS.ResourceTags {
					if !userHasTag(user, userTag.Key, userTag.Value) {
						log.Infof("user missing tag: %s=%s", userTag.Key, userTag.Value)
						return false
					}
				}
	}
```

pkg/aws/actuator/actuator.go
sairameshv: According to this [comment](https://github.com/openshift/enhancements/pull/1051/files#r821548397), the deletion of the tag from the Infra resource doesn't actually need to delete the tag from the existing AWS resource. It remains in the unman-aged state. Moreover, user needs to pass the empty string against any tag key that need to be deleted from the AWS resource.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
sairameshv: Not really required, the tags remain in the unmanaged state.

akhil-rane:
@sairameshv This looks good to me. Could you please resolve the conflicts and then we can merge it.

PR #446, opened 2022-01-21
Leader election migration 1: ConfigMap & Lease

2uasimojo:
We were using ConfigMap-based leader election. We need to migrate to
using Lease instead.

Part 1: Use ConfigMap *and* Lease so we can upgrade smoothly from a
build using just ConfigMap.
Part 2: Remove ConfigMap.

This is part 1.

x-ref: https://issues.redhat.com/browse/CCO-191
epic: https://issues.redhat.com/browse/CCO-182

### DIFF
--- a/manifests/01-role.yaml
+++ b/manifests/01-role.yaml
@@ -30,3 +30,15 @@ rules:
   - rolebindings
   verbs:
   - "*"
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
--- a/pkg/cmd/operator/cmd.go
+++ b/pkg/cmd/operator/cmd.go
@@ -39,7 +39,6 @@ import (
 
 	"github.com/openshift/library-go/pkg/controller/fileobserver"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -52,8 +51,8 @@ import (
 )
 
 const (
-	defaultLogLevel         = "info"
-	leaderElectionConfigMap = "cloud-credential-operator-leader"
+	defaultLogLevel        = "info"
+	leaderElectionLockName = "cloud-credential-operator-leader"
 
 	caConfigMapMountPath = "/var/run/configmaps/trusted-ca-bundle"
 	caConfigMapName      = "tls-ca-bundle.pem"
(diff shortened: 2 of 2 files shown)


2uasimojo:
/assign @joelddiaz 

I tested this live and it worked, BUT I had to apply the RBAC part manually -- I'm not sure I got that right in the PR.

akhil-rane (commented):
Tested on a live cluster and everything looks good to me. Let's keep it on hold till 4.11 branch opens as suggested.

/hold

2uasimojo:
> Let's keep it on hold till 4.11 branch opens as suggested.

As suggested where?

To be clear, I agree: it's easier to have phase 1 in a z=0 than worry about a customer skipping the fixed z when they upgrade. We just need to remember to do phase 2 after that.

manifests/01-role.yaml
```diff
@@ -30,3 +30,15 @@ rules:
...
+  resources:
+  - leases
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
```
akhil-rane, line 44: Do we need all these permissions here?  If yes we can just put `"*"`

2uasimojo: I've been told (though I can't remember where) that best practice is to enumerate all the permissions rather than using `"*"`.

2uasimojo: Jason found this for me

https://source.redhat.com/groups/public/product-security/content/product_security_wiki/operators_secure_deployment_guide#4-1-avoid-wildcards

...though the example makes it somewhat ambiguous whether it's okay to use the wildcard when you _do_ need all the verbs.

akhil-rane: that sounds fair

akhil-rane:
@2uasimojo this looks fine to me. 4.11 cycle has opened so should we remove hold here?

manifests/01-role.yaml
2uasimojo: Also [from Grant](https://coreos.slack.com/archives/C0281LN70AC/p1643838336786099?thread_ts=1643836110.625409&cid=C0281LN70AC): "i think it's because verbs are largely opaque, the verbs that make up  `*` today might not be the same as `*` in the future"

2uasimojo:
> 4.11 cycle has opened so should we remove hold here

sgtm

/hold cancel

jeana-redhat:
/label docs-approved

No doc impact :+1:

lwan-wanglin:
@2uasimojo Here is the test case for the PR https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitems?query=trello%3ACCO%5C-182 , please help to review, thanks!

lwan-wanglin:
> Test case is great @lwan-wanglin, thanks!
Thanks for your review
/label qe-approved

PR #454, opened 2022-02-24
Bug 2058270: Remove Azure mint mode support as Active Directory Graph API will be sunset

2uasimojo:
Manual backport of #453.

---

Migrate away from depending on the Azure Active Directory Graph API since it will be sunset in June 2022.

* [x]  No more mint mode for Azure. So continue supporting passthrough and manual mode as the only supported modes on Azure.
* [x]  The implementation is to always mark the Azure root secret as passthrough, and pivot a live running originally-mint cluster to passthrough mode.
* [x]  Overwrite existing Secrets with the content found in kube-system/azure-credentials (we can assume this as removing the root creds was never supported in Azure).
* [x]  Try to clean up previously minted App Registrations / Service Principals (while the Azure AD Graph API still works), but treat errors as non-fatal (just set a condition on the CredentialsRequest).
* [x]  Update test cases.

PR #455, opened 2022-03-17
Bug 2064428: Remove Azure mint mode support as Active Directory Graph API will be sunset

suhanime:
Manual backport of #454.

---

Migrate away from depending on the Azure Active Directory Graph API since it will be sunset in June 2022.

* [x]  No more mint mode for Azure. So continue supporting passthrough and manual mode as the only supported modes on Azure.
* [x]  The implementation is to always mark the Azure root secret as passthrough, and pivot a live running originally-mint cluster to passthrough mode.
* [x]  Overwrite existing Secrets with the content found in kube-system/azure-credentials (we can assume this as removing the root creds was never supported in Azure).
* [x]  Try to clean up previously minted App Registrations / Service Principals (while the Azure AD Graph API still works), but treat errors as non-fatal (just set a condition on the CredentialsRequest).
* [x]  Update test cases.


akhil-rane (commented):
/lgtm

jianping-shu:
It is weird that "launch 4.7-ci,openshift/cloud-credential-operator#455 azure" didn't work as expected, like the PR was not really merged into the image.
So I used "build openshift/cloud-credential-operator#455" for verification.
Tested with OCP-47143 (with registry.build01.ci.openshift.org/ci-ln-hld048t/release:latest, 4.7 image based on the PR), passed
Tested with OCP-47144 and OCP-47159 (upgrade from 4.7 version whose default mode is "mint" to registry.build01.ci.openshift.org/ci-ln-hld048t/release:latest, 4.7 image based on the PR), passed

PR #484, opened 2022-08-10
Bug 2117474: Fix panic when the Provider spec is empty in credential request

Karthik-K-N:
This PR contains the changes to fix the panic while generating a secret from a credential request when the Provider spec is empty or wrongly indented, Instead of panic the below error is thrown

```
2022/08/10 14:11:35 Failed to validate the serviceID: Spec.ProviderSpec is empty in powervs-block-csi-driver-operator credentials request
```


mkumatag (changes requested):
LGTM, can you please add some UT around this scenarios?

akhil-rane (commented):
/lgtm

PR #494, opened 2022-10-03
Update OWNERS to reflect reality.

abutcher:


suhanime:
/lgtm

Should we be backporting this to every branch that needs the change? I think in this case we should

2uasimojo:
Merging without labels as this does not impact docs, px, etc.

PR #495, opened 2022-10-03
Changes generated from make update-gofmt with go-1.19.

abutcher:
Fixes `verify` and carries the change from #493


wking (approved):
/lgtm

PR #511, opened 2023-02-02
OCPBUGS-6370: Bump k8s dependencies from v0.25.3 to v0.26.1 

abutcher:
This commit updates various k8s.io vendored dependencies to v0.26.1 to
pull in the fixes for https://github.com/advisories/GHSA-xrjj-mj9h-534m.

[OCPBUGS-6370](https://issues.redhat.com/browse/OCPBUGS-6370)


go.mod
```diff
@@ -116,34 +115,34 @@ require (
...
 	go.opentelemetry.io/otel v1.10.0 // indirect
 	go.opentelemetry.io/otel/trace v1.10.0 // indirect
-	golang.org/x/crypto v0.0.0-20220511200225-c6db032c6c88 // indirect
-	golang.org/x/net v0.0.0-20220826154423-83b083e8dc8b // indirect
-	golang.org/x/sys v0.0.0-20220728004956-3c1f35247d10 // indirect
-	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211 // indirect
-	golang.org/x/text v0.3.7 // indirect
-	golang.org/x/tools v0.1.12 // indirect
+	golang.org/x/crypto v0.1.0 // indirect
+	golang.org/x/net v0.5.0 // indirect
```
abutcher, line 127: :heavy_check_mark:

abutcher:
`make update` and `make verify` are flip flopping into a state that fails `make verify` :construction: 
/hold

abutcher:
Added a commit that fixes the CRD schema flip flopping that was occurring, see [50bee56](https://github.com/openshift/cloud-credential-operator/pull/511/commits/50bee569f1c834bc7f5a95d4183000f731d05855)

```
We copy the cloud credential operator config CRD from the
openshift/api repository (via the vendor dir) and since openshift/api
now utilizes a different codegen utility than build-machinery-go,
running the schemapatch code generator against the copied manifest
results in a different formatting. This results in a diff which causes
our verify target to fail so we ensure the copied manifests remain
unchanged by copying the CRDs once more after generating CRDs as a
workaround.
```

2uasimojo (commented):
This looks good as far as it goes. However, I believe dependabot is still going to complain about [these](https://github.com/openshift/cloud-credential-operator/blob/7b0a3577bb125c019c51666b744e243bfdfccaab/go.sum#L838-L890) and [these](https://github.com/openshift/cloud-credential-operator/blob/7b0a3577bb125c019c51666b744e243bfdfccaab/go.sum#L1014-L1023) from go.sum.

If the x/net bump is still relevant to the k8s bump, we can keep it in here, but we should take out the claim that we're going to be resolving those dependabot complaints.

If it's not relevant, perhaps we should pull it out and put together a separate PR that addresses dependabot complaints specifically and completely.

abutcher:
@2uasimojo I think this will resolve the dependabot alerts based on the [PR the dependabot automation opened](https://github.com/openshift/cloud-credential-operator/pull/516) but I'm not 100% sure. I kinda wanna try merging this to see.

> If the x/net bump is still relevant to the k8s bump, we can keep it in here, but we should take out the claim that we're going to be resolving those dependabot complaints.

The golang.org/x/net bump to 0.5.0 is relevant to the k8s bump and is the intended fix for https://github.com/advisories/GHSA-xrjj-mj9h-534m from OCPBUGS-6370. The subsequent bump to 0.7.0 golang.org/x/net bump is for the dependabot alerts opened recently.

2uasimojo:
> I kinda wanna try merging this to see.

Well, okay. I'm basing my assertion on what happened in the hive repo, where at least the x/text flags were being raised based solely on go.sum.

/retest

jianping-shu:
@abutcher Looks like it is only dependency library update, we'll run the existing cco auto cases for regression.
@huangmingxia Pls. take it

2uasimojo (commented):
If CI is happy, I'm happy.

/lgtm

PR #492, opened 2022-09-21
Make pod-identity-webhook deployment HA by default

abutcher:
[CCO-259](https://issues.redhat.com//browse/CCO-259)

This PR increases the replicas of the aws-pod-identity-webhook deployment from 1 to 2 and adds management of a `PodDisruptionBudget` with `minAvailable=1` for the mutating webhook deployment.

PR #512, opened 2023-02-06
OCPBUGS-6977: Set pod identity webhook replicas=1 when infrastructure topology is SingleReplica

abutcher:
Previously, https://github.com/openshift/cloud-credential-operator/pull/492 increased the replicas of the AWS pod identity webhook from 1 to 2. We cannot default deployment replicas=2 on a cluster with a `"SingleReplica"` infrastructure topology so this commit checks the infrastructure object to determine the topology and adjusts the replicas when the topology is `"SingleReplica"`. An error will be returned when the cluster infrastructure object is not found. Default replicas (2) used when no infrastructure topology set and adjust replicas=1 when `"SingleReplica"` topology found.

Additionally, the pod disruption budget will not be deployed on `"SingleReplica"` topology clusters.

Rename the podidentitywebhook controller file to "awspodidentitywebhook_controller.go" and add controller tests to ensure the replicas + pdb match the topology.

[OCPBUGS-6977](https://issues.redhat.com/browse/OCPBUGS-6977)

### DIFF
--- a/pkg/operator/utils/utils.go
+++ b/pkg/operator/utils/utils.go
@@ -61,6 +61,17 @@ func LoadCredsFromSecret(kubeClient client.Client, namespace, secretName string)
 	return accessKeyID, secretAccessKey, nil
 }
 
+// LoadInfrastructureTopology loads the topology from the cluster Infrastructure config.
+func LoadInfrastructureTopology(c client.Client, logger log.FieldLogger) (configv1.TopologyMode, error) {
+	infra, err := GetInfrastructure(c)
+	if err != nil {
+		logger.WithError(err).Error("error loading Infrastructure topology")
+		return "", err
+	}
+	logger.Debugf("Loading infrastructure topology: %s", infra.Status.InfrastructureTopology)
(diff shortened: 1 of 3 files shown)


pkg/operator/utils/utils.go
```diff
@@ -61,6 +61,17 @@ func LoadCredsFromSecret(kubeClient client.Client, namespace, secretName string)
...
 }
 
+// LoadInfrastructureTopology loads the topology from the cluster Infrastructure config.
+func LoadInfrastructureTopology(c client.Client, logger log.FieldLogger) (configv1.TopologyMode, error) {
+	infra, err := GetInfrastructure(c)
+	if err != nil {
+		logger.WithError(err).Error("error loading Infrastructure topology")
+		return "", err
+	}
+	logger.Debugf("Loading infrastructure topology: %s", infra.Status.InfrastructureTopology)
```
2uasimojo, line 71: nit: do we `WithField` around here?

2uasimojo (commented):
Seems straightforward and sensible to me.

/lgtm

PR #514, opened 2023-02-16
OCPBUGS-16181: Write manifests when AWS IAM roles already exist.

abutcher:


abutcher:
Holding for testing & to check if manifest generation for other clouds will have the same issue.
/hold

pkg/cmd/provisioning/aws/create-iam-roles.go
```diff
@@ -227,6 +227,11 @@ func createRole(awsClient aws.Client, name string, credReq *credreqv1.Credential
 		} else {
 			role = outRole.Role
 			log.Printf("Existing role %s found", *role.Arn)
```
2uasimojo, line 229: IIUC we get here if there's a role of the expected name. How confident can we be that it has the right settings? Should we validate them? What would be the right behavior if they don't match?
- Delete the role and recreate it? If this is right, then we could simplify by simply doing that regardless.
- Modify the role to have the right attributes? For some deltas -- like tags -- this might make sense; but for others, it could be Bad™ -- like if it would change the actual security-ness of the role.
- ...so probably the only right answer is to error. Which may be why the code was the way it was in the first place.

abutcher: It looks like we [apply policy](https://github.com/openshift/cloud-credential-operator/blob/a4819474e58d1ab88207dd6decdcbcd5ce9eecad/pkg/cmd/provisioning/aws/create-iam-roles.go#L232-L236) derived from the provided `CredentialsRequests` on roles after this even if the role already existed so we may already be stomping the existing role policy to be what we expect based on the invocation.

2uasimojo:
Okay, I was getting confused about what we were writing here. It's the secret in the cluster, not the policy in the cloud. This makes sense.

Add UT? Like [here](https://github.com/openshift/cloud-credential-operator/blob/a4819474e58d1ab88207dd6decdcbcd5ce9eecad/pkg/cmd/provisioning/aws/create-iam-roles_test.go#L218-L239)?

abutcher:
Yep! This writes the secrets we put into the manifests directory to be read by the installer. There was some more discussion on the card so /holding WIP.

/hold

PR #519, opened 2023-03-06
Upgrade build-machinery-go: `make vulncheck`

2uasimojo:
Bring in the latest BMG so we can now `make vulncheck` to get a preview of what dependabot might complain about.


abutcher:
Adding labels for build-machinery-go vendor update.

/label docs-approved
/label px-approved
/label qe-approved

PR #522, opened 2023-03-14
OCPBUGS-8666: Correct pod-identity-webhook annotations for PreferredDuringScheduling.

abutcher:
The previous change introducing the annotations had a typo: s/annotation/annotations/.

[OCPBUGS-8666](https://issues.redhat.com/browse/OCPBUGS-8666)

PR #524, opened 2023-03-24
[release-4.13] OCPBUGS-10822: Adding missing workload annotations

eggfoobar:
Cherry-picks #520 and #522 to 4.13.

With workload pinning feature introduced in 4.13 we'll need to backport this annotation to 4.13. We've also added tests in origin to help catch these annotations going forward.

/assign @abutcher


eggfoobar:
@abutcher Spelling error is impossible this time! 😅

PR #525, opened 2023-04-04
CCO modified for Token mode, a PoC

bentito:
Based on:
image

Steps for the PoC / Demo:

- Assuming using OpenShift Local (CRC) based on OpenShift 4.12.5
- Follow the steps here to make cluster STS Enabled: https://github.com/openshift/cloud-credential-operator/blob/master/docs/sts.md#steps-to-in-place-migrate-an-openshift-cluster-to-sts
- Note that commit 77385e8 hard codes the platform type to AWS for purposes of the PoC/demo.
- ```kubectl replace -f bindata/bootstrap/cloudcredential_v1_credentialsrequest_crd.yaml```
- ```export IMG=quay.io/btofel/cco:latest```
- ```docker build -t $IMG -f Dockerfile.local . && docker push $IMG```
- ```for file in `ls -1 manifests/`; do kubectl apply -f manifests/$file; done``` (which will deploy with the modified `manifests/03-deployment.yaml`
- ```kubectl apply -f cred-reqs/nginx_credentials-request.yaml -n openshift-cloud-credential-operator``` (which will apply a`CredentialsRequest` with the modified fields.

A Secret will be created in the Default Namespace with the needed token. Operator should be able to take advantage of this to access cloud resources.

### DIFF
--- a/pkg/apis/cloudcredential/v1/types_credentialsrequest.go
+++ b/pkg/apis/cloudcredential/v1/types_credentialsrequest.go
@@ -60,6 +60,12 @@ type CredentialsRequestSpec struct {
 	// credentials flow.
 	// +optional
 	ServiceAccountNames []string `json:"serviceAccountNames,omitempty"`
+	// CloudTokenString (for instance for AWS a Role ARN)
+	// +optional
+	CloudTokenString string `json:"cloudTokenString,omitempty"`
+	// CloudTokenPath (JWT token)
+	// +optional
+	CloudTokenPath string `json:"cloudTokenPath,omitempty"`
 }
 
 // CredentialsRequestStatus defines the observed state of CredentialsRequest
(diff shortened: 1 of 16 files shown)


pkg/apis/cloudcredential/v1/types_credentialsrequest.go
```diff
@@ -60,6 +60,12 @@ type CredentialsRequestSpec struct {
 	// credentials flow.
 	// +optional
 	ServiceAccountNames []string `json:"serviceAccountNames,omitempty"`
+	// CloudTokenString (for instance for AWS a Role ARN)
+	// +optional
+	CloudTokenString string `json:"cloudTokenString,omitempty"`
+	// CloudTokenPath (JWT token)
+	// +optional
+	CloudTokenPath string `json:"cloudTokenPath,omitempty"`
```
jharrington22, line 68: @bentito do we want the path configurable since today its not in CCO

bentito: @jharrington22 I can't quite tell from how you are asking:

Do you think we should make the path configurable? If so, is it enough that it's configurable in the CredentialsRequest? Or did you mean it should just be a well-known-location?

PR #535, opened 2023-04-24
Add a make target and stub for actuator e2e tests

bentito:
This should probably merge before the changes to the release config envisioned in the below linked PR, such that the make target will be available to calling by pj-rehearse:
https://github.com/openshift/release/pull/38624

### DIFF
--- a/pkg/aws/actuator/actutator_e2e_test.go
+++ b/pkg/aws/actuator/actutator_e2e_test.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 )
 
-// Test_CheckSTS runs end-to-end tests to verify the basics of an STS workflow work.
+// Test_CheckSTS runs end-to-end tests to verify that the results of an STS workflow are successful.
 
 func Test_CheckSTS(t *testing.T) {
 	t.Run("test of STS", func(t *testing.T) {
--- a/pkg/aws/actuator/actutator_e2e_test.go
+++ b/pkg/aws/actuator/actutator_e2e_test.go
@@ -1,0 +1,16 @@
+//go:build e2e
+// +build e2e
+
+package actuator
+
+import (
+	"testing"
+)
+
+// Test_CheckSTS runs end-to-end tests to verify the basics of an STS workflow work.
+
+func Test_CheckSTS(t *testing.T) {
+	t.Run("test of STS", func(t *testing.T) {
+		t.Parallel()
+	})
+}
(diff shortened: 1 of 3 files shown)


bentito:
@abutcher I'd like to merge this and then check that the corresponding pj-rehearsal using the STS workflow works. Sound reasonable?

abutcher (commented):
@bentito sounds good to me.

pkg/aws/actuator/actutator_e2e_test.go
```diff
@@ -0,0 +1,16 @@
...
+package actuator
+
+import (
+	"testing"
+)
+
+// Test_CheckSTS runs end-to-end tests to verify that the results of an STS workflow are successful.
+
+func Test_CheckSTS(t *testing.T) {
+	t.Run("test of STS", func(t *testing.T) {
```
abutcher, line 13: I hadn't considered that this test would run during normal unit tests so we should exclude this one in some way, perhaps like [this example in the Hive Makefile](https://github.com/openshift/hive/blob/928bef3a88c299ac96a5c8fd4eba4f1e92ccd2c1/Makefile#LL40C1-L41C53). I think we should also move this test into an e2e directory of some sort again [similarly to the Hive setup](https://github.com/openshift/hive/tree/master/test/e2e) to make it clear that this test isn't meant to be ran during unit testing.

abutcher: I take back the first part :) I see that we have[ that in CCO](https://github.com/openshift/cloud-credential-operator/blob/1261fa91b46943e3d46988bace39b295c09f7bef/Makefile#L46-L47). Maybe we should just move this test into an e2e directory.

bentito: Okay, 2018c178 isolates the e2e tests along the hive model. I'm trying avoid fully following that, with the `hack/*.sh` and all because it seems unneeded so far. I left the e2e tagging, but maybe that's overkill with them moved to a new location. Lemme know what you think.

abutcher: That seems fine, thanks!

bentito:
@abutcher want to re-lgtm this now and then I'll go off and see what needs to happen to get valid-bug (or other more appropriate label) on this?

bparees:
discussed w/ TRT and got approval to override the bug requirement:
https://redhat-internal.slack.com/archives/C01CQA76KMX/p1683641369678309

PR #542, opened 2023-05-19
CCO-366 Add ability to detect AWS STS and behave accordingly

bentito:
Adds capability to detect AWS STS enabled clusters and behave a bit like Mint mode when detected.

There is now a small unit test suite. And a single e2e test that uses an STS [workflow](urlhttps://steps.ci.openshift.org/workflow/openshift-e2e-aws-manual-oidc-sts).

Notes:

- Will need changes to allow for move of the TAT data to AWSProviderSpec;
- Is the e2e really working? Lack of "STS detection" related logging and lack of created Secret in the `must-gather` make me unsure:
```bash
$  yq '.items[].metadata.name' /tmp/artifacts/registry*/namespaces/default/core/secrets.yaml
builder-dockercfg-l6mps
builder-token-6skc2
default-dockercfg-28hgj
default-token-pr62d
deployer-dockercfg-p248v
deployer-token-djlzm
```
Ah, it seems to be working, see inline comment with log lines showing Secret creation/deletion on CredentialsRequest
addition in the e2e test.


pkg/aws/actuator/actuator.go
```diff
@@ -323,50 +323,90 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
...
-
-	switch credentialsRootSecret.Annotations[constants.AnnotationKey] {
-	case constants.InsufficientAnnotation:
-		msg := "cloud credentials insufficient to satisfy credentials request"
-		logger.Error(msg)
-		return &actuatoriface.ActuatorError{
-			ErrReason: minterv1.InsufficientCloudCredentials,
-			Message:   msg,
+	stsDetected, err := utils.IsTimedTokenCluster(a.Client, logger)
+	if stsDetected {
```
bentito, line 330: After this change e2e started working. I am kind of unsure about it though.

test/e2e/aws/sts/actutator_e2e_test.go
```diff
@@ -4,13 +4,132 @@
...
+				t.Fatal(err)
+			}
+			return ctx
+		}).Feature()
+	testenv.Test(t, credReqFeature)
+}
+
+func onAdd(t *testing.T, cfg envconf.Config, ctx context.Context) func(obj interface{}) {
+	time.Sleep(2 * time.Minute)
+	if err := cfg.Client().Resources().Get(ctx, secretName, namespace, secret); err != nil {
```
bentito, line 88: This is bit due to unsure about Secret really existing here, despite this being a callback from the watch on the Secret.

pkg/aws/actuator/actuator.go
bentito: Okay, here is logging from a Clusterbot cluster launched with `workflow-launch openshift-e2e-aws-manual-oidc-sts https://github.com/openshift/cloud-credential-operator/pull/542` and then having run `make test-e2e-sts` manually:

```log
time="2023-06-12T13:40:45Z" level=info msg="reconciling clusteroperator status"
time="2023-06-12T13:40:45Z" level=info msg="operator set to disabled / manual mode" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:40:45Z" level=info msg="syncing credentials request" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:40:45Z" level=info msg="adding finalizer: cloudcredential.openshift.io/deprovision" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req secret=default/test-sts-secret
time="2023-06-12T13:40:45Z" level=info msg="operator set to disabled / manual mode" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:40:45Z" level=info msg="syncing credentials request" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:40:45Z" level=info msg="timed token access cluster detected: true, so not trying to provision with root secret" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req secret=default/test-sts-secret
time="2023-06-12T13:40:45Z" level=info msg="actuator detected STS enabled cluster making secret" actuator=aws cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:40:45Z" level=info msg="creating secret" actuator=aws cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:40:45Z" level=info msg="clusteroperator status updated" controller=status
time="2023-06-12T13:40:45Z" level=info msg="reconciling clusteroperator status"
time="2023-06-12T13:40:45Z" level=info msg="secret created successfully" actuator=aws cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:40:56Z" level=info msg="calculating metrics for all CredentialsRequests" controller=metrics
time="2023-06-12T13:40:56Z" level=info msg="reconcile complete" controller=metrics elapsed=4.220597ms
time="2023-06-12T13:41:56Z" level=info msg="reconciling clusteroperator status"
time="2023-06-12T13:42:56Z" level=info msg="calculating metrics for all CredentialsRequests" controller=metrics
time="2023-06-12T13:42:56Z" level=info msg="reconcile complete" controller=metrics elapsed=4.133929ms
time="2023-06-12T13:44:45Z" level=info msg="reconciling clusteroperator status"
time="2023-06-12T13:44:45Z" level=info msg="operator set to disabled / manual mode" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:44:45Z" level=info msg="syncing credentials request" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:44:45Z" level=warning msg="no user name set on credentials being deleted, most likely were never provisioned or using passthrough creds" actuator=aws cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:44:45Z" level=info msg="target secret deleted successfully" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req secret=default/test-sts-secret targetSecret=default/test-sts-secret
time="2023-06-12T13:44:45Z" level=info msg="actuator deletion complete, removing finalizer" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req secret=default/test-sts-secret
time="2023-06-12T13:44:45Z" level=info msg="reconciling clusteroperator status"
time="2023-06-12T13:44:45Z" level=info msg="operator set to disabled / manual mode" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:44:45Z" level=info msg="syncing credentials request" controller=credreq cr=openshift-cloud-credential-operator/test-sts-creds-req
time="2023-06-12T13:44:45Z" level=info msg="clusteroperator status updated" controller=status
time="2023-06-12T13:44:56Z" level=info msg="calculating metrics for all CredentialsRequests" controller=metrics
time="2023-06-12T13:44:56Z" level=info msg="reconcile complete" controller=metrics elapsed=3.554297ms
time="2023-06-12T13:46:56Z" level=info msg="calculating metrics for all CredentialsRequests" controller=metrics
time="2023-06-12T13:46:56Z" level=info msg="reconcile complete" controller=metrics elapsed=4.054413ms
time="2023-06-12T13:46:56Z" level=info msg="reconciling clusteroperator status"
```
So I feel more confident the text is really passing.

abutcher (changes requested):
Couple small things. Wondering if some of the new `Info` logs should be moved to `Debug` based on their potential noise level but those can be moved or followed up on.

Makefile
```diff
@@ -149,7 +149,7 @@ test-no-gen: test
 .PHONY: test-no-gen
 
 test-e2e-sts:
-	go test -mod=vendor -race -tags e2e ./test/e2e/aws/sts/...
+	go test -mod=mod -race -tags e2e ./test/e2e/aws/sts/...
```
stevekuznetsov, line 152: This is temporary, right?

pkg/aws/actuator/actuator.go
```diff
@@ -326,47 +336,96 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
...
-	switch credentialsRootSecret.Annotations[constants.AnnotationKey] {
-	case constants.InsufficientAnnotation:
-		msg := "cloud credentials insufficient to satisfy credentials request"
-		logger.Error(msg)
-		return &actuatoriface.ActuatorError{
-			ErrReason: minterv1.InsufficientCloudCredentials,
-			Message:   msg,
+	if stsFeatureGateEnabled && stsDetected {
+		if a.Codec == nil {
+			return fmt.Errorf("invalid codec, nil value")
```
stevekuznetsov, line 349: This seems like a programmer error , can it ever happen normally in runtime?

pkg/aws/actuator/actuator.go
```diff
@@ -326,47 +336,96 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
...
+	return nil
+}
 
+// createSecret makes a time-based token available in a Secret in the namespace of an operator that
+// has supplied the following in the CredentialsRequest:
+// a non-nil spec.CloudTokenString
+// a path to the JWT token: spec.cloudTokenPath
+// a spec.SecretRef.Name
+// a cr.Spec.SecretRef.Namespace
+func (a *AWSActuator) createSecret(awsSTSIAMRoleARN string, cloudTokenPath string, secretRef string,
```
stevekuznetsov, line 411: nit: probably best to name this something STS-specific

pkg/aws/actuator/actuator.go
```diff
@@ -326,47 +336,96 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretRef,
+			Namespace: secretRefNamespace,
+		},
+		StringData: map[string]string{
+			"credentials": fmt.Sprintf(awsSTSCredsTemplate, awsSTSIAMRoleARN, cloudTokenPath),
+		},
+		Type: corev1.SecretTypeOpaque,
+	}
+	err := a.Client.Create(context.TODO(), secret)
```
stevekuznetsov, line 424: The caller of `createSecret` has a context - prefer to pass them through or behavior on leader election failover will be weird

pkg/aws/actuator/actuator_test.go
```diff
@@ -572,3 +573,165 @@ func (a *testAWSError) OrigErr() error {
...
+	codec, err := minterv1.NewCodec()
+	if err != nil {
+		t.Fatalf("failed to set up codec for tests: %v", err)
+	}
+
+	tests := []struct {
+		name               string
+		existing           []runtime.Object
+		wantErr            assert.ErrorAssertionFunc
+		ctx                context.Context
```
stevekuznetsov, line 589: nit: in tests this is always going to be `context.Background()` so perhaps not necessary to parameterize?

pkg/cmd/operator/cmd.go
```diff
@@ -39,8 +39,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	v1 "github.com/openshift/api/config/v1"
```
stevekuznetsov, line 42: nit: prefer `configv1`

pkg/gcp/actuator/actuator.go
```diff
@@ -64,6 +64,10 @@ type Actuator struct {
 	GCPClientBuilder func(string, []byte) (ccgcp.Client, error)
 }
 
+func (a *Actuator) STSFeatureGateEnabled() bool {
```
stevekuznetsov, line 67: nit: I vaguely remember reading this concept being named something not AWS-specific (as opposed to STS) - does that matter?

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -179,15 +175,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	if err != nil {
 		return err
 	}
 
 	// allCredRequestsMapFn simply looks up all CredentialsRequests and requests they be reconciled.
-	allCredRequestsMapFn := handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
+	allCredRequestsMapFn := handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, a client.Object) []reconcile.Request {
 		log.Info("requeueing all CredentialsRequests")
 		crs := &minterv1.CredentialsRequestList{}
 		err := mgr.GetClient().List(context.TODO(), crs)
```
stevekuznetsov, line 189: Pass the context you got in the method signature here

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -231,7 +227,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// for that new namespace. This allows us to be up and running for other components that don't
 	// yet exist, but will.
 
-	namespaceMapFn := handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
+	namespaceMapFn := handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, a client.Object) []reconcile.Request {
 		// Iterate all CredentailsRequests to determine if we have any that target
 		// this new namespace. We are not anticipating huge numbers of CredentialsRequests,
 		// nor namespace creations.
```
stevekuznetsov, line 233: Pass the context you got in the method signature into the List() below

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -374,6 +370,13 @@ func (r *ReconcileCredentialsRequest) Reconcile(ctx context.Context, request rec
 	}()
 
 	mode, conflict, err := utils.GetOperatorConfiguration(r.Client, logger)
+
+	stsDetected := false
```
stevekuznetsov, line 374: We do this block of code as well inside of the AWS actuator - perhaps just pull it out and take the detected bool itself in the actuator constructor?

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -526,94 +532,139 @@ func (r *ReconcileCredentialsRequest) Reconcile(ctx context.Context, request rec
...
-	if err != nil {
-		log.WithError(err).Debug("error retrieving cloud credentials secret, admin can remove root credentials in mint mode")
-	}
-	cloudCredsSecretUpdated := credentialsRootSecret != nil && credentialsRootSecret.ResourceVersion != cr.Status.LastSyncCloudCredsSecretResourceVersion
-	isStale := cr.Generation != cr.Status.LastSyncGeneration
-	hasRecentlySynced := cr.Status.LastSyncTimestamp != nil && cr.Status.LastSyncTimestamp.Add(syncPeriod).After(time.Now())
-	hasActiveFailureConditions := checkForFailureConditions(cr)
+		var syncErr error
+		if !credsExists {
+			syncErr = r.Actuator.Create(context.TODO(), cr)
```
stevekuznetsov, line 547: nit: here and elsewhere - you have a context available to you in `Reconcile()` - no need for TODO

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -526,94 +532,139 @@ func (r *ReconcileCredentialsRequest) Reconcile(ctx context.Context, request rec
...
+		if err != nil {
+			log.WithError(err).Debug("error retrieving cloud credentials secret, admin can remove root credentials in mint mode")
+		}
+		cloudCredsSecretUpdated := credentialsRootSecret != nil && credentialsRootSecret.ResourceVersion != cr.Status.LastSyncCloudCredsSecretResourceVersion
+		isStale := cr.Generation != cr.Status.LastSyncGeneration
+		hasRecentlySynced := cr.Status.LastSyncTimestamp != nil && cr.Status.LastSyncTimestamp.Add(syncPeriod).After(time.Now())
+		hasActiveFailureConditions := checkForFailureConditions(cr)
+
+		if !cloudCredsSecretUpdated && !isStale && hasRecentlySynced && crSecretExists && !hasActiveFailureConditions && cr.Status.Provisioned {
+			logger.Debug("lastsyncgeneration is current and lastsynctimestamp was less than an hour ago, so no need to sync")
```
stevekuznetsov, line 589: nit: I think it might be nice to get all of the inputs to this `if` statement as structured fields in the logger - I can imagine a world where one of those conditions is not correct and the log message today does not provide enough information to figure out which

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -526,94 +532,139 @@ func (r *ReconcileCredentialsRequest) Reconcile(ctx context.Context, request rec
...
-	hasRecentlySynced := cr.Status.LastSyncTimestamp != nil && cr.Status.LastSyncTimestamp.Add(syncPeriod).After(time.Now())
-	hasActiveFailureConditions := checkForFailureConditions(cr)
+		var syncErr error
+		if !credsExists {
+			syncErr = r.Actuator.Create(context.TODO(), cr)
+		} else {
+			syncErr = r.Actuator.Update(context.TODO(), cr)
+		}
+		var provisionErr bool
+		if syncErr != nil {
```
stevekuznetsov, line 552: This block seems to be duplicated below, verbatim - are we repeating ourselves?

pkg/operator/platform/platform.go
```diff
@@ -59,3 +67,55 @@ func getClient(explicitKubeconfig string) (client.Client, error) {
...
+	clientSet, err := configclient.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	configInformers := configinformers.NewSharedInformerFactory(clientSet, 10*time.Minute)
+	desiredVersion := computeClusterOperatorVersions()
+	missingVersion := desiredVersion
+
+	// By default, this will exit(0) the process if the featuregates ever change to a different set of values.
+	// Notice the nil EventRecorder passed, is this okay?
```
stevekuznetsov, line 93: Sounds like something we need an answer to still?

pkg/operator/platform/platform.go
```diff
@@ -59,3 +67,55 @@ func getClient(explicitKubeconfig string) (client.Client, error) {
 
 	return dynamicClient, nil
 }
+
+func GetFeatureGates() (featuregates.FeatureGate, error) {
+	stop := make(chan struct{})
+	ctx := context.Background()
```
stevekuznetsov, line 73: The code in the command's initialization creates a context and binds it to signals, leader election, etc - pefer to pass that in here?

pkg/operator/platform/platform.go
```diff
@@ -59,3 +67,55 @@ func getClient(explicitKubeconfig string) (client.Client, error) {
...
+
+	featureGates, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return nil, err
+	}
+	return featureGates, nil
+}
+
+func computeClusterOperatorVersions() string {
+	currentVersion := os.Getenv("RELEASE_VERSION")
```
stevekuznetsov, line 119: What if this is not set?

pkg/operator/utils/utils.go
```diff
@@ -72,6 +72,32 @@ func LoadInfrastructureTopology(c client.Client, logger log.FieldLogger) (config
...
+//  2. Is serviceAccountIssuer non-empty
+//
+// Both of these conditions must be true for any timed access token enabled clusters for the implementations mentioned above.
+func IsTimedTokenCluster(c client.Client, logger log.FieldLogger) (bool, error) {
+	credentialsMode, _, err := GetOperatorConfiguration(c, logger)
+	if err != nil {
+		logger.WithError(err).Error("error loading CCO configuration to determine mode")
+		return false, err
+	}
+	if credentialsMode != "Manual" {
```
stevekuznetsov, line 87: super-nit: is manual not a constant somewhere?

pkg/operator/utils/utils.go
```diff
@@ -72,6 +72,32 @@ func LoadInfrastructureTopology(c client.Client, logger log.FieldLogger) (config
...
+		return false, nil
+	}
+	authConfig, err := GetAuth(c)
+	if err != nil {
+		logger.WithError(err).Error("error loading authentication config")
+		return false, err
+	}
+	if authConfig.Spec.ServiceAccountIssuer == "" {
```
stevekuznetsov, line 95: nit:

```go
return authConfig.Spec.ServiceAccountIssuer != ""
```

pkg/operator/utils/utils.go
```diff
@@ -112,6 +138,14 @@ func GetInfrastructure(c client.Client) (*configv1.Infrastructure, error) {
 	return infra, nil
 }
 
+func GetAuth(c client.Client) (*configv1.Authentication, error) {
+	auth := &configv1.Authentication{}
+	if err := c.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, auth); err != nil {
```
stevekuznetsov, line 143: let's thread through the context for client calls

test/e2e/aws/sts/actutator_e2e_test.go
stevekuznetsov: The `obj interface{}` passed to this func is the object that was just added - perhaps use that? 2min wait should not be needed here

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -382,7 +385,10 @@ func (r *ReconcileCredentialsRequest) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, fmt.Errorf("configuration conflict")
 	} else if mode == operatorv1.CloudCredentialsModeManual {
 		logger.Infof("operator set to disabled / manual mode")
-		return reconcile.Result{}, err
+		if stsDetected {
+			logger.Infof("operator detects STS enabled cluster")
```
abutcher, line 389: We now log this and the message above for each CR reconcile which is slightly contradictory. Does it make sense to combine the spirit of the message in some way?

```
time="2023-06-23T14:07:13Z" level=info msg="operator set to disabled / manual mode" controller=credreq cr=openshift-cloud-credential-operator/manila-csi-drive
r-operator
time="2023-06-23T14:07:13Z" level=info msg="operator detects STS enabled cluster" controller=credreq cr=openshift-cloud-credential-operator/manila-csi-driver-
operator
```

pkg/operator/platform/platform.go
```diff
@@ -89,12 +93,22 @@ func GetFeatureGates() (featuregates.FeatureGate, error) {
...
 
+	kubeClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create kube client: %w", err)
+	}
+	eventRecorder := events.NewKubeRecorder(kubeClient.CoreV1().Events("openshift-cloud-credential-operator"), "cloud-credential-operator", &corev1.ObjectReference{
+		APIVersion: "apps/v1",
+		Kind:       "Deployment",
+		Namespace:  "openshift-cloud-credential-operator",
+		Name:       "cloud-credential-operator",
```
abutcher, line 104: I hardcoded values for namespace & deployment here but they should probably be constants.

bentito:
> Couple small things. Wondering if some of the new `Info` logs should be moved to `Debug` based on their potential noise level but those can be moved or followed up on.

Changed all 3 instances to Debug in d6bb9169

pkg/aws/actuator/actuator.go
```diff
@@ -326,47 +330,89 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
...
+		logger.Info("actuator detected STS enabled cluster, enabling STS secret brokering for CredentialsRequests providing an IAM Role ARN")
+		awsSTSIAMRoleARN, err := awsSTSIAMRoleARN(a.Codec, cr)
 		if err != nil {
 			return err
 		}
-	case constants.MintAnnotation:
-		logger.Debugf("provisioning with cred minting")
-		err := a.syncMint(ctx, cr, logger)
+		if awsSTSIAMRoleARN != "" {
+			err = a.createSecret(awsSTSIAMRoleARN, cr.Spec.CloudTokenPath, cr.Spec.SecretRef.Name, cr.Spec.SecretRef.Namespace, logger)
```
abutcher, line 357: The check for this `err != nil` is outside the if statement.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -382,7 +383,10 @@ func (r *ReconcileCredentialsRequest) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, fmt.Errorf("configuration conflict")
 	} else if mode == operatorv1.CloudCredentialsModeManual {
 		logger.Infof("operator set to disabled / manual mode")
-		return reconcile.Result{}, err
+		if !stsDetected {
+			logger.Infof("operator detects STS enabled cluster")
+			return reconcile.Result{}, err
```
abutcher, line 390: This return short circuits continuing through the controller to create the secret for the CredentialsRequest.

pkg/apis/cloudcredential/v1/types_aws.go
```diff
@@ -28,6 +28,13 @@ type AWSProviderSpec struct {
 	metav1.TypeMeta `json:",inline"`
 	// StatementEntries contains a list of policy statements that should be associated with this credentials access key.
 	StatementEntries []StatementEntry `json:"statementEntries"`
+	// STSRoleARN is the Amazon Resource Name (ARN) of an IAM Role which was created manually for the associated
+	// CredentialsRequest.
+	// The presence of an STSRoleARN within the AWSProviderSpec initiates creation of a secret containing IAM
+	// Role details necessary for assuming the IAM Role via Amazon's Secure Token Service.
+	// +optional
+	// +kubebuilder:validation:Pattern:="^arn:(?P<Partition>[^:\n]*):(?P<Service>[^:\n]*):(?P<Region>[^:\n]*):(?P<AccountID>[^:\n]*):(?P<Ignore>(?P<ResourceType>[^:\/\n]*)[:\/])?(?P<Resource>.*)$"
```
abutcher, line 36: The role ARN validation is causing CRD generation to fail which for some reason fails silently when ran via `make update`. The silently failing part is possibly related to [the openshift/api workaround](https://github.com/openshift/cloud-credential-operator/blob/f9a6219d066fb61ae77e6d89298bc4cc7c56c37d/Makefile#L89-L111).

```
'_output/tools/bin/controller-gen-v0.2.5' schemapatch:manifests="./manifests" paths="./pkg/apis/cloudcredential/v1" 'output:dir="./manifests"'
/home/abutcher/go/src/github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1/types_aws.go:36:2: invalid char escape (at <input>:1:1)
/home/abutcher/go/src/github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1/types_aws.go:36:2: invalid char escape (at <input>:1:1)
/home/abutcher/go/src/github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1/types_aws.go:36:2: unable to parse string: invalid syntax (at <input>:1:1)
/home/abutcher/go/src/github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1/types_aws.go:36:2: invalid char escape (at <input>:1:1)
/home/abutcher/go/src/github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1/types_aws.go:36:2: invalid char escape (at <input>:1:1)
/home/abutcher/go/src/github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1/types_aws.go:36:2: unable to parse string: invalid syntax (at <input>:1:1)
Error: not all generators ran successfully
```

pkg/operator/credentialsrequest/credentialsrequest_controller.go
abutcher: Worth adding a test to the credentials request controller tests to ensure we go through the motions.

pkg/aws/actuator/actuator.go
```diff
@@ -326,11 +330,18 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
...
+		if a.Codec == nil {
+			return fmt.Errorf("invalid codec, nil value")
+		}
+		logger.Info("actuator detected STS enabled cluster, enabling STS secret brokering for CredentialsRequests providing an IAM Role ARN")
+		awsSTSIAMRoleARN, err := awsSTSIAMRoleARN(a.Codec, cr)
+		if err != nil {
+			return err
+		}
+		if awsSTSIAMRoleARN != "" {
+			err = a.createSecret(awsSTSIAMRoleARN, cr.Spec.CloudTokenPath, cr.Spec.SecretRef.Name, cr.Spec.SecretRef.Namespace, logger)
```
abutcher, line 357: `CloudTokenPath` is an optional field so we should use "/var/run/secrets/kubernetes.io/serviceaccount/token" as the path when it is not found.

pkg/aws/actuator/actuator.go
```diff
@@ -326,11 +330,18 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
...
-			err = a.createSecret(cr.Spec.CloudTokenString, cr.Spec.CloudTokenPath, cr.Spec.SecretRef.Name, cr.Spec.SecretRef.Namespace, logger)
+		if a.Codec == nil {
+			return fmt.Errorf("invalid codec, nil value")
+		}
+		logger.Info("actuator detected STS enabled cluster, enabling STS secret brokering for CredentialsRequests providing an IAM Role ARN")
+		awsSTSIAMRoleARN, err := awsSTSIAMRoleARN(a.Codec, cr)
+		if err != nil {
+			return err
+		}
+		if awsSTSIAMRoleARN != "" {
```
abutcher, line 356: Could invert this if and return when the `awsSTSIAMRoleARN == ""` because there will be nothing to do without `awsSTSIAMRoleARN` having been set.

pkg/aws/actuator/actuator.go
bentito: yeah it seems like this would just catch a test setup error, because it seems like not getting a NewCodec() is handled elsewhere at runtime. Will remove this in upcoming commit.

pkg/aws/actuator/actuator.go
bentito: Yeah I agree, renaming in upcoming commit.

pkg/aws/actuator/actuator.go
bentito: good point, changing in upcoming commit.

pkg/aws/actuator/actuator_test.go
bentito: yep, removing in upcoming commit

pkg/cmd/operator/cmd.go
bentito: done in upcoming commit

pkg/gcp/actuator/actuator.go
bentito: Actually GCP calls it STS too. So it's okay naming I think.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito: done in upcoming commit

pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito: done in upcoming commit

pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito: yep, eliminated `TODO()` usage everywhere in this file

pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito: Not sure if code below looked like this:
```go
stsFeatureGateEnabled := r.Actuator.STSFeatureGateEnabled()
```
when the comment was made, but this seems to answer that, at least enough to not mess with the actuator initialization more...

pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito: Seems reasonable. Done in upcoming commit.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito: Seems like it, pulled it up to a method that gets called from each branch in upcoming commit.

pkg/operator/platform/platform.go
bentito: Okay passing in context now in an upcoming commit.

Can you determine/explain if/why we should keep that `context.WithCancel(ctx)` call and subsequent goroutine that immediately follows?

pkg/operator/utils/utils.go
bentito: yikes! investigating I find that mode referred to as both "manual" and "Manual" in yamls, docs and code.  Yes, there is a constant, that has it as "manual" (https://github.com/openshift/cloud-credential-operator/commit/1b384ccd0a8f587e2bf79281f241983ce9235ef5#diff-f6d93feef6d59ca8ced14b54fe0058d41f14de44450baf6b723cc1c2ec280c54R25) but it doesn't seemed to be used everywhere now....

So I'm a little leery to start using the constant version here until CCO team can review a bit more. I'd say it can wait until another day to get better?

pkg/operator/utils/utils.go
bentito: Done in upcoming commit.

pkg/operator/utils/utils.go
bentito: done in upcoming commit

pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito: this is take care of now seemingly

pkg/operator/platform/platform.go
bentito: moved to constants in upcoming commit

pkg/aws/actuator/actuator.go
bentito: got fixed

pkg/aws/actuator/actuator.go
bentito: this is take care of

pkg/aws/actuator/actuator.go
bentito: got done.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito: It's the end of the day and it's been a long one for me, again, I'm not seeing a good way to create a test on the controller with STS enabled. There might be one, I'm just kind of lost at the moment. Maybe we should add a card to Jira for "Streamline Credentials Request Controller Unit Tests and add STS enabled and feature gate flags"?

pkg/operator/platform/platform.go
```diff
@@ -59,3 +72,64 @@ func getClient(explicitKubeconfig string) (client.Client, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get kube config: %v", err)
+	}
+	clientSet, err := configclient.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	configInformers := configinformers.NewSharedInformerFactory(clientSet, 10*time.Minute)
+	desiredVersion := computeClusterOperatorVersions()
+	missingVersion := desiredVersion
```
jstuever, line 94: I believe this should be `missingVersion := "0.0.1-snapshot"` to match what is in the [03-deployment.yaml](https://github.com/openshift/cloud-credential-operator/blob/ee67cc6b1a8cbd2248a2536b0dfa93e18d1bfda2/manifests/03-deployment.yaml#L71)

pkg/operator/platform/platform.go
jstuever: A string cannot be nil in golang, it will instead be an empty string. So, in such a case it will be looking for the cluster version to be an empty string. In such a case, I expect a lot of other things will also be broken. The good news is RELEASE_VERSION is defined in the [03-deployment.yaml](https://github.com/openshift/cloud-credential-operator/blob/ee67cc6b1a8cbd2248a2536b0dfa93e18d1bfda2/manifests/03-deployment.yaml#L71) and should always be non-empty.

jstuever (commented):
I have general concern about the logic defining what is STS enabled vs feature gate enabled (per platform). Otherwise, a few minor nitpicks.

bentito:
> I have general concern about the logic defining what is STS enabled vs feature gate enabled (per platform). Otherwise, a few minor nitpicks.

@jstuever Could you elaborate a bit? Even given that we'll probably need to come back to fix the logic later, it would be helpful to put the specifcs here for a reference for that?

One thing I noticed as I try to add a unit test on the STS enabled path in credentialsrequest_controller.go, I am finding it easy to test the FeatureGate enabled part, but hard to fake the STS detected part, because that boolean is hard to reach from the tests.

pkg/operator/credentialsrequest/actuator/actuator.go
```diff
@@ -46,6 +46,8 @@ type Actuator interface {
 	Upgradeable(operatorv1.CloudCredentialsMode) *configv1.ClusterOperatorStatusCondition
 	// GetCredentialsRootSecret returns the credentials root secret.
 	GetCredentialsRootSecret(ctx context.Context, cr *minterv1.CredentialsRequest) (*corev1.Secret, error)
+	// Query STSFeatureGateEnabled.
+	STSFeatureGateEnabled() bool
```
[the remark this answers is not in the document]
bentito: yep, noted, I understand what you're saying, if you elaborate a little bit on your top level comment and make comment on the Jira about removing the FeatureGate that should do it for remembering to deal with this.

bentito:
In reply to https://github.com/openshift/cloud-credential-operator/pull/542#discussion_r1244042037
There is another constants file floating around this codebase and it's confusing:

`pkg/operator/constants/constants.go` has it as "manual"

pkg/operator/utils/utils.go
```diff
@@ -72,6 +72,29 @@ func LoadInfrastructureTopology(c client.Client, logger log.FieldLogger) (config
...
+	}
+	if credentialsMode != "Manual" {
+		return false, nil
+	}
+	authConfig, err := GetAuth(ctx, c)
+	if err != nil {
+		logger.WithError(err).Error("error loading authentication config")
+		return false, err
+	}
+	return authConfig.Spec.ServiceAccountIssuer != "", nil
```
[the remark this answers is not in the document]
bentito: cc @abutcher just so you this for later handling

pkg/operator/platform/platform.go
bentito: That seems wrong to hard code to a string which might end up different from YAML contents?

jstuever: Possibly... I just noted the difference from many of the other operators.

- [cluster-ingress-operator](https://github.com/openshift/cluster-ingress-operator/blob/ecb3786b1397cb93865952c98db72d1990587bd0/pkg/operator/operator.go#L106)
- [cloud-network-config-controller](https://github.com/openshift/cloud-network-config-controller/pull/102/files#diff-b52c6361b0946b90f2d35912dbe72c9dc8a5acc2d94a43cc26ad640b45e2be9aR41)
- [azure-disk-csi-driver-operator](https://github.com/openshift/azure-disk-csi-driver-operator/pull/83/files#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6R108)
- [azure-file-csi-driver-operator](https://github.com/openshift/azure-file-csi-driver-operator/pull/54/files#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6R74)
- [cluster-storage-operator](https://github.com/openshift/cluster-storage-operator/blob/ab03ebedd1b43342bf3ddd53cb113cf654f14d6d/pkg/operator/standalone_starter.go#L39)
- [machine-api-provider-azure](https://github.com/openshift/machine-api-provider-azure/blob/d3a1d2861718c6ae0ef0dbafaa7fe7bdbfa9fbef/cmd/manager/main.go#L137)

I believe what exists today (`missingVersion := desiredVersion`) will cause it to treat all versions as if they are the missing version, thereby skipping the version check altogether.

jstuever: Something to keep in mind, the purpose of `missingVersion` is to handle the situation where the 03-deployment.yaml is applied from locally compiled code, and where it is not going through the process of having a RELEASE_VERSION defined via our release process. Or, in other words, it is explicitly to enable locally compiled operator(s) to function as desired by ignoring the version check itself.

jstuever:
Responding to https://github.com/openshift/cloud-credential-operator/pull/542#issuecomment-1610077466

I agree, the constants file is confusing. I had to dig in to understand how it is used. It looks like they are used solely for keys (string) in a mapping in the metrics operator. The mapping appears to be more of a status use case than a desired configuration. There also appear to be additional status options compared to `CloudCredentialsMode` type definition. Why the statuses were selected to be lowercase, I do not know.

pkg/operator/platform/platform.go
bentito: 👍

stevekuznetsov:
/lgtm

@abutcher gave his blessing last week, the new e2e are passing, @bentito let's do this!

2uasimojo:
/approve

proxy for @abutcher

PR #544, opened 2023-06-22
manager: filter the cache of configmaps

stevekuznetsov:
This controller LISTs and WATCHes ConfigMaps in order to react to changes to one and only one ConfigMap - the controller's (legacy) configuration. Previously, this controller would hold the entire cluster's ConfigMaps in cache in order to do this; with this change, we will only ever watch one ConfigMap and hold it in memory. This is unfortunately a cross-cutting concern and it blurs the control flow that previously exited, but we need these options to be passed to the cache constructor at the moment we create the manager, so it's not clear that there is a factoring that keeps this from happening until the controller initialization code runs, like where we add filtering today.

### DIFF
--- a/pkg/cmd/operator/cmd.go
+++ b/pkg/cmd/operator/cmd.go
@@ -29,9 +29,15 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/google/uuid"
+	"github.com/openshift/cloud-credential-operator/pkg/operator/constants"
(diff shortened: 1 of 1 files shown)


bparees:
cc @abutcher

stevekuznetsov:
/retitle PORTENABLE-526: manager: filter the cache of configmaps

PR #545, opened 2023-06-22
OCPBUGS-15365: *: use a filtered LIST + WATCH on Secrets for AWS STS

stevekuznetsov:
The status quo for this controller is to LIST + WATCH all Secrets on the cluster. This consumes
more resources than necessary on clusters where users put other data in Secrets themselves, as we
hold that data in our cache and never do anything with it. The reconcilers mainly need to react to
changes in Secrets created for CredentialRequests, which they control and can label, allowing us
to filter the LIST + WATCH down and hold the minimal set of data in memory. However, two caveats:

- in passthrough and mint mode, admin credentials are also provided by the user through Secrets,
  and we need to watch those, but we can't label them
- on existing clusters, secrets exist from previous work these reconcilers did that will not have
  Secrets labelled

We could solve the second issue with an interim release of this controller that labels all previous
Secrets, but does not restrict the watch stream.

Due to the way that controller-runtime closes over the client/cache concepts, it's difficult to
solve the first issue, though, since we'd need two sets of clients and caches, both for Secrets,
and ensure that we use one for client access to Secrets we're creating or mutating and the other
when we're interacting with admin credentials. Not impossible to do, but tricky to implement and
complex.

Until we undertake that effort, we apply a simplification to the space: only when AWS STS mode is
enabled, we will try to filter the LIST + WATCH. This mode is brand new, so we can be reasonably
sure that there are no previous secrets on the cluster, and, we make the filtering best-effort
in order to check if that assumption held. Second, AWS STS mode only runs in clusters without
admin credentials, so if we apply the filter, we should not see failures downstream from clients
that hope to see those objects but can't.

### DIFF
--- a/pkg/apis/cloudcredential/v1/types_credentialsrequest.go
+++ b/pkg/apis/cloudcredential/v1/types_credentialsrequest.go
@@ -27,6 +27,10 @@ const (
 	// credentials in AWS before allowing the CredentialsRequest to be deleted in etcd.
 	FinalizerDeprovision string = "cloudcredential.openshift.io/deprovision"
 
+	// LabelCredentialsRequest is to mark Secrets created as a target of CredentailsRequests.
+	LabelCredentialsRequest      string = "cloudcredential.openshift.io/credentials-request"
+	LabelCredentialsRequestValue string = "true"
+
 	// AnnotationCredentialsRequest is used on Secrets created as a target of CredentailsRequests.
 	// The annotation value will map back to the namespace/name of the CredentialsRequest that created
 	// or adopted the secret.
--- a/pkg/apis/cloudcredential/v1/types_credentialsrequest.go
+++ b/pkg/apis/cloudcredential/v1/types_credentialsrequest.go
@@ -27,11 +27,11 @@ const (
 	// credentials in AWS before allowing the CredentialsRequest to be deleted in etcd.
 	FinalizerDeprovision string = "cloudcredential.openshift.io/deprovision"
 
-	// LabelCredentialsRequest is to mark Secrets created as a target of CredentailsRequests.
+	// LabelCredentialsRequest is to mark Secrets created as a target of CredentialsRequests.
 	LabelCredentialsRequest      string = "cloudcredential.openshift.io/credentials-request"
 	LabelCredentialsRequestValue string = "true"
 
-	// AnnotationCredentialsRequest is used on Secrets created as a target of CredentailsRequests.
+	// AnnotationCredentialsRequest is used on Secrets created as a target of CredentialsRequests.
 	// The annotation value will map back to the namespace/name of the CredentialsRequest that created
 	// or adopted the secret.
 	AnnotationCredentialsRequest string = "cloudcredential.openshift.io/credentials-request"
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -507,7 +507,7 @@ func (a *AWSActuator) syncPassthrough(ctx context.Context, cr *minterv1.Credenti
 	}
 
 	// userPolicy param empty because in passthrough mode this doesn't really have any meaning
-	err = a.syncAccessKeySecret(cr, accessKeyID, secretAccessKey, existingSecret, "", logger)
+	err = a.syncAccessKeySecret(ctx, cr, accessKeyID, secretAccessKey, existingSecret, "", logger)
 	if err != nil {
 		msg := "error creating/updating secret"
 		logger.WithError(err).Error(msg)
@@ -692,7 +692,7 @@ func (a *AWSActuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequ
 		accessKeyString = *accessKey.AccessKeyId
 		secretAccessKeyString = *accessKey.SecretAccessKey
 	}
-	err = a.syncAccessKeySecret(cr, accessKeyString, secretAccessKeyString, existingSecret, desiredUserPolicy, logger)
+	err = a.syncAccessKeySecret(ctx, cr, accessKeyString, secretAccessKeyString, existingSecret, desiredUserPolicy, logger)
 	if err != nil {
 		log.WithError(err).Error("error saving access key to secret")
 		return err
@@ -995,7 +995,7 @@ func (a *AWSActuator) getLogger(cr *minterv1.CredentialsRequest) log.FieldLogger
 	})
 }
 
-func (a *AWSActuator) syncAccessKeySecret(cr *minterv1.CredentialsRequest, accessKeyID, secretAccessKey string, existingSecret *corev1.Secret, userPolicy string, logger log.FieldLogger) error {
+func (a *AWSActuator) syncAccessKeySecret(ctx context.Context, cr *minterv1.CredentialsRequest, accessKeyID, secretAccessKey string, existingSecret *corev1.Secret, userPolicy string, logger log.FieldLogger) error {
 	sLog := logger.WithFields(log.Fields{
 		"targetSecret": fmt.Sprintf("%s/%s", cr.Spec.SecretRef.Namespace, cr.Spec.SecretRef.Name),
 		"cr":           fmt.Sprintf("%s/%s", cr.Namespace, cr.Name),
@@ -1015,6 +1015,9 @@ func (a *AWSActuator) syncAccessKeySecret(cr *minterv1.CredentialsRequest, acces
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      cr.Spec.SecretRef.Name,
 				Namespace: cr.Spec.SecretRef.Namespace,
+				Labels: map[string]string{
+					minterv1.LabelCredentialsRequest: minterv1.LabelCredentialsRequestValue,
+				},
 				Annotations: map[string]string{
 					minterv1.AnnotationCredentialsRequest:   fmt.Sprintf("%s/%s", cr.Namespace, cr.Name),
 					minterv1.AnnotationAWSPolicyLastApplied: userPolicy,
@@ -1027,7 +1030,7 @@ func (a *AWSActuator) syncAccessKeySecret(cr *minterv1.CredentialsRequest, acces
 			},
 		}
 
-		err := a.Client.Create(context.TODO(), secret)
+		err := a.Client.Create(ctx, secret)
 		if err != nil {
 			sLog.WithError(err).Error("error creating secret")
 			return err
@@ -1039,6 +1042,11 @@ func (a *AWSActuator) syncAccessKeySecret(cr *minterv1.CredentialsRequest, acces
 	// Update the existing secret:
 	sLog.Debug("updating secret")
 	origSecret := existingSecret.DeepCopy()
+	if existingSecret.Labels == nil {
+		existingSecret.Labels = map[string]string{}
+	}
+	existingSecret.Labels[minterv1.LabelCredentialsRequest] = minterv1.LabelCredentialsRequestValue
+
 	if existingSecret.Annotations == nil {
 		existingSecret.Annotations = map[string]string{}
 	}
@@ -1054,7 +1062,7 @@ func (a *AWSActuator) syncAccessKeySecret(cr *minterv1.CredentialsRequest, acces
 
 	if !reflect.DeepEqual(existingSecret, origSecret) {
 		sLog.Info("target secret has changed, updating")
-		err := a.Client.Update(context.TODO(), existingSecret)
+		err := a.Client.Update(ctx, existingSecret)
 		if err != nil {
 			msg := "error updating secret"
 			sLog.WithError(err).Error(msg)
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -69,6 +69,7 @@ var _ actuatoriface.Actuator = (*AWSActuator)(nil)
 // AWSActuator implements the CredentialsRequest Actuator interface to create credentials in AWS.
 type AWSActuator struct {
 	Client                             client.Client
+	RootCredClient                     client.Client
 	Codec                              *minterv1.ProviderCodec
 	AWSClientBuilder                   func(accessKeyID, secretAccessKey []byte, c client.Client) (ccaws.Client, error)
 	Scheme                             *runtime.Scheme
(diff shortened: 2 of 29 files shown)


stevekuznetsov:
Would like some guidance for what kinds of tests we want to add to this :)

pkg/aws/actuator/actuator.go
```diff
@@ -441,7 +441,7 @@ func (a *AWSActuator) syncPassthrough(ctx context.Context, cr *minterv1.Credenti
 	}
 
 	// userPolicy param empty because in passthrough mode this doesn't really have any meaning
-	err = a.syncAccessKeySecret(cr, accessKeyID, secretAccessKey, existingSecret, "", logger)
+	err = a.syncAccessKeySecret(ctx, cr, accessKeyID, secretAccessKey, existingSecret, "", logger)
```
stevekuznetsov, line 444: These changes were not strictly necessary but it seemed like latent tech debt and since I was touching the client calls, I figured we could fix them.

pkg/operator/controller.go
```diff
@@ -62,10 +64,22 @@ func init() {
 var AddToManagerFuncs []func(manager.Manager, string) error
 
 // AddToManagerWithActuatorFuncs is a list of functions to add all Controllers with Actuators to the Manager
-var AddToManagerWithActuatorFuncs []func(manager.Manager, actuator.Actuator, configv1.PlatformType) error
+var AddToManagerWithActuatorFuncs []func(manager.Manager, actuator.Actuator, configv1.PlatformType, corev1client.CoreV1Interface) error
 
 // AddToManager adds all Controllers to the Manager
 func AddToManager(m manager.Manager, explicitKubeconfig string) error {
+	rules := clientcmd.NewDefaultClientConfigLoadingRules()
```
stevekuznetsov, line 71: Controller-runtime does not have support for server-side-apply yet, and by far the best way for us to add labels to existing objects surgically, and without conflict problems, is using server-side-apply. I create a client-go client and thread it through in order to allow that.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -336,6 +381,64 @@ func isAdminCredSecret(namespace, secretName string) bool {
...
+		return reconcile.Result{}, err
+	}
+
+	applyConfig := corev1applyconfigurations.Secret(secret.Name, secret.Namespace)
+	applyConfig.WithLabels(map[string]string{
+		minterv1.LabelCredentialsRequest: minterv1.LabelCredentialsRequestValue,
+	})
+
+	if _, err := r.mutatingClient.Secrets(secret.Namespace).Apply(ctx, applyConfig, metav1.ApplyOptions{
+		Force:        true, // we're the authoritative owner of this field and should not allow anyone to stomp it
```
stevekuznetsov, line 432: @deads2k would love a gut-check that this does what I think it does, given that the applyConfig is set up the way it is - is there a single owner for all of `metadata.labels`?

stevekuznetsov:
/hold

It is not clear to me if labeling the root credentials in `kube-system` is a tenable approach. I need to look more into how `controller-runtime` works with a filtered cache to see if we can use a different approach here.

stevekuznetsov:
https://issues.redhat.com/browse/PORTENABLE-526

stevekuznetsov:
/retitle PORTENABLE-526: *: label the secrets we interact with

stevekuznetsov:
After some chats with @abutcher and @deads2k  : 

 - we cannot mutate the admin credentials in `kube-system`
 - using multiple informers (one for labelled secrets, N for user-provided ones in `kube-system`) is really nasty in `controller-runtime` right now and the cost of an implementation is high
 - STS mode - the specific mode we care about footprint in - will have no root credential
 - I will refactor this to piggy-back on #542 and, as part of that feature flag:
   - run a startup check to ensure that no credential-request-created secrets exist without the label
   - run the controller using a label selector on secrets for the LIST+WATCH

Since STS feature flag does not exist before that PR merges, we don't have to worry about pulling in old data / upgrades / etc.

@sdodson @bparees WDYT?

bparees:
sgtm but why can't we mutate the kube-system cred (just to label it)?  because it may be user provided/user-managed?

alternatively, if it's not associated w/ credreq, why do we need to label it?

jianping-shu:
Verified with the following:
Checked on the OCP cluster without the PR change.
jianpingshu@jshu-mac ~ % oc get secret -A -l "cloudcredential.openshift.io/credentials-request=true"
No resources found

Checked on the OCP cluster with the PR change.
jianpingshu@jshu-mac ~ % oc get secret -A -l "cloudcredential.openshift.io/credentials-request=true"
NAMESPACE                                   NAME                                     TYPE     DATA   AGE
kube-system                                 aws-creds                                Opaque   2      69m
openshift-cloud-credential-operator         cloud-credential-operator-iam-ro-creds   Opaque   3      66m
openshift-cloud-network-config-controller   cloud-credentials                        Opaque   3      66m
openshift-cluster-csi-drivers               ebs-cloud-credentials                    Opaque   3      66m
openshift-image-registry                    installer-cloud-credentials              Opaque   3      66m
openshift-ingress-operator                  cloud-credentials                        Opaque   3      66m
openshift-machine-api                       aws-cloud-credentials                    Opaque   3      66m

stevekuznetsov:
@jianping-shu sorry - we had to change the approach we were taking. I am re-writing this PR and will let you know when it's ready for review.

pkg/apis/cloudcredential/v1/types_credentialsrequest.go
```diff
@@ -27,6 +27,10 @@ const (
 	// credentials in AWS before allowing the CredentialsRequest to be deleted in etcd.
 	FinalizerDeprovision string = "cloudcredential.openshift.io/deprovision"
 
+	// LabelCredentialsRequest is to mark Secrets created as a target of CredentailsRequests.
```
bparees, line 30:
```suggestion
	// LabelCredentialsRequest is to mark Secrets created as a target of CredentialsRequests.
```

pkg/cmd/operator/cmd.go
```diff
@@ -99,6 +105,70 @@ func NewOperator() *cobra.Command {
...
+					for _, secret := range secrets.Items {
+						if credentialsrequest.IsMissingSecretLabel(&secret) {
+							missing = append(missing, types.NamespacedName{
+								Namespace: secret.Namespace,
+								Name:      secret.Name,
+							})
+						}
+					}
+					if len(missing) != 0 {
+						log.WithField("missing", missing).Warnf("%s feature gate enabled but not all secrets labelled, falling back to caching all secrets on cluster", configv1.FeatureGateAWSSecurityTokenService)
```
bparees, line 166: why not just label them if it's missing?

pkg/apis/cloudcredential/v1/types_credentialsrequest.go
stevekuznetsov: bad copy pasta, thanks, fixed

pkg/cmd/operator/cmd.go
```diff
@@ -99,6 +105,70 @@ func NewOperator() *cobra.Command {
...
+				// more resources than necessary on clusters where users put other data in Secrets themselves, as we
+				// hold that data in our cache and never do anything with it. The reconcilers mainly need to react to
+				// changes in Secrets created for CredentialRequests, which they control and can label, allowing us
+				// to filter the LIST + WATCH down and hold the minimal set of data in memory. However, two caveats:
+				// - in passthrough and mint mode, admin credentials are also provided by the user through Secrets,
+				//   and we need to watch those, but we can't label them
+				// - on existing clusters, secrets exist from previous work these reconcilers did that will not have
+				//   Secrets labelled
+				// We could solve the second issue with an interim release of this controller that labels all previous
+				// Secrets, but does not restrict the watch stream.
```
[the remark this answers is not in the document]
stevekuznetsov: > did you look into just always labeling all the secrets at startup time, before going into list/watch mode?

It was not clear to me how to do this without a race / with any guarantee that this process finishes - we would want to make sure that no Secrets are created or mutated between the moment we are done labelling and when we start the LIST+WATCH. Thoughts?

> and/or recording some sort of "labeling complete" indicator in the ClusterOperator status so we can avoid doing it every time we start?

Similarly here, we'd ideally want to keep that field on the OperatorStatus up to date over time, right, so we'd need to know, given a watch even we're processing in the controller at runtime, whether or not we need to update the OperatorStatus. It's not clear how we'd know when we were processing the last un-labelled Secret and therefore could call the job done.

pkg/cmd/operator/cmd.go
```diff
@@ -99,6 +105,70 @@ func NewOperator() *cobra.Command {
...
+				// when we're interacting with admin credentials. Not impossible to do, but tricky to implement and
+				// complex.
+				// Until we undertake that effort, we apply a simplification to the space: only when AWS STS mode is
+				// enabled, we will try to filter the LIST + WATCH. This mode is brand new, so we can be reasonably
+				// sure that there are no previous secrets on the cluster, and, we make the filtering best-effort
+				// in order to check if that assumption held. Second, AWS STS mode only runs in clusters without
+				// admin credentials, so if we apply the filter, we should not see failures downstream from clients
+				// that hope to see those objects but can't.
+				var filteredWatchPossible bool
+				if awsSecurityTokenServiceGateEnabled {
```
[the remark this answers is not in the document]
stevekuznetsov: @bentito

pkg/cmd/operator/cmd.go
stevekuznetsov: The controller I added will do so, I didn't want to add `O(count(secrets))` mutations to the startup of this controller, could make CCO unresponsive on large clusters/

pkg/cmd/operator/cmd.go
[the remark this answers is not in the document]
bparees: > we would want to make sure that no Secrets are created or mutated 

secret creation doesn't matter, right?  because there are two things creating secrets:

1) us - no worries, label will be applied.  also we won't be creating any until after we get through the initialization phase
2a) someone else, but the secret isn't ref'd by a credreq - don't care
2b) someone else, they create a credreq and a secret ref'd by the cred req - we'll pick up this secret when we see the credreq object and go to inject content into the secret, at which point we can label it.

so i think creation is probably handled.

Mutation isn't, i think, a concern, because mutation implies the secret already existed when we were initalizing, in which case we'd have already labeled it.  I guess the one gap might be:

1) we list + label secrets
2) user edits secret and stomps the label away
3) we never reconcile that secret

though in theory that could also be a concern at runtime?  If someone edits a secret to remove our label, will we get notified of the change to the secret, given that it no longer matches our filter?  Certainly if they do it while the controller is down we'd have that problem unless we relist+label all secrets when we start up.

Regardless i don't see how doing this in a previous release helps.  If there is a window between "we listing/watching all secrets and label them" and "we're only watching labeled secrets" in which a secret can be created/mutated such that we never reconcile it again, then that window would exist whether we:
1) ship one release that has a controller looking at all of them, then in a subsequent release we stop doing so
or
2) ship, in a single release, logic that looks at all of them, and then stops doing so

no?

Also the current logic assumes "if we list all secrets and they are all labeled, we're safe to filter", but doesn't that have similar race condition concerns? (a secret could be created/mutated after we finish our listing and think it's safe to filter) At least if we label them all while listing them, assuming we are successful in labeling them, we don't have to wait a release for most clusters to be able to enable the logic.  But that still doesn't address the race concern.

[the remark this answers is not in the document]
stevekuznetsov: Addressed in the latter commits.

pkg/cmd/operator/cmd.go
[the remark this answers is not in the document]
bparees: i looked up how this is defined and it's just a TP flag, has nothing to do w/ whether the cluster is in STS mode or not, just whether the cluster is TP-enabled or not.  Hopefully that aligns w/ how you intended it to behave.

in otherwords, once we move this "feature" out of TP, this logic will run on every cluster, not just AWS-STS clusters.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -316,6 +322,144 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			return IsMissingSecretLabel(e.ObjectNew)
+		},
+		CreateFunc: func(e event.CreateEvent) bool {
+			return IsMissingSecretLabel(e.Object)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return IsMissingSecretLabel(e.Object)
+		},
+	}
+	// Watch Secrets and reconcile if we see an event for an admin credential secret in kube-system.
```
bparees, line 360: i thought we were specifically *not* going to reconcile(add labels to) the admin secret?

pkg/cmd/operator/cmd.go
bparees, line 166: thought you were going to attempt to add labels to secrets that are missing them and only if that failed, fallback to caching all secrets?

pkg/operator/credentialsrequest/credentialsrequest_controller.go
stevekuznetsov: Oops, bad copy-pasta

pkg/cmd/operator/cmd.go
stevekuznetsov: In the controller that runs, not at startup.

stevekuznetsov: I could not find a clean way to thread the context for  the controller down far enough to where we could cancel it and cause self-destruction/restart. Happy to put a `log.Fatal()` in there but it seems like making that control flow abnormal might be hard to reason about for other devs in the future.

pkg/azure/actuator_test.go
```diff
@@ -190,32 +173,6 @@ func TestDecodeToUnknown(t *testing.T) {
 	}
 }
 
-func TestAnnotations(t *testing.T) {
```
stevekuznetsov, line 193: This test did nothing to my eyes, it only ran the constructor. I wasn't sure how to adapt it.

stevekuznetsov:
/hold

The last commit is far-reaching ... would need more :eye: 

/test e2e-aws-manual-oidc

stevekuznetsov:
Was sending a label selector instead of a field selector, oops. Should be good now.

stevekuznetsov:
@abutcher will need an approval as well!

pkg/operator/credentialsrequest/credentialsrequest_controller.go
deads2k: Labels are divided by name, so you only own the labels that you're trying to set, not all the labels available.  At least that's my memory of it.  Is the reality different?

PR #546, opened 2023-06-22
PORTENABLE-526: operator: use a partial metadata watch for Namespaces

stevekuznetsov:
We only watch namespaces to react to their creation, and the only information we need about a namespace is its' name, so we can use a partial metadata watch to winnow down the size of our cache. Namespaces don't really store much in spec and status, so this effect is likely to not be enormous, but every bit helps.


stevekuznetsov:
/retitle PORTENABLE-526: operator: use a partial metadata watch for Namespaces

PR #555, opened 2023-06-27
Move manifests to simplify codegen

2uasimojo:
We spent some time in #511 and again in #542 trying to reconcile CRDs. The problem is that we want to *generate* the CredentialsRequest CRD from code in this repo, but *use* (copy) the CloudCredential CRD from openshift/api, which we vendor. But we invoke controller-gen through build-machinery-go, and it does unexpected things to the latter, which breaks validation.

With this commit, we move the CredentialsRequest CRD to a `generated` subdirectory and the CloudCredential CRD to an `imported` subdirectory. This lets us go back to the simpler invocation of bmg's tooling while keeping everything in the shape we expect.

One more quirk: Because build-machinery-go starts defining dependency chains for targets like `update`, we need to start defining that dependency chain *before* we import the bmg libs to ensure that we copy/generate CRDs *before* we include them in bindata.


2uasimojo:
Sounds good @stevekuznetsov. We'll want to:
- Wait until `controller-gen@master` gets a release
- Submit a PR to build-machinery-go to bump the [controller-gen version](https://github.com/openshift/build-machinery-go/blob/d321ffa04533d88d2fdca5bcf2d5555fdd02a0a3/make/targets/openshift/controller-gen.mk#L19) to that release
- Update the build-machinery-go vendored dep here in CCO to pick it up.

We also may still need to do something akin to the `make` dependency shuffle in this PR, since we still want to make sure we've done the regeneration and copying before we build bindata. Without that, I was finding yesterday that one would have to run `make update*` multiple times if those CRDs had changed.

stevekuznetsov:
@2uasimojo I chatted with the folks doing there, and it seems like the real flow will be:

 - make `build-machinery-go` use `go mod` to install `controller-gen`
 - re-vendor `build-machinery-go` in CCO
 - update the version of `controller-gen` used in CCO to be newer

Looks like they don't want to be pushing releases anymore, so the `curl`-based install is not something we can continue to do anymore in BMG. Even now it's pulling in something only as recent as last September...

stevekuznetsov:
@2uasimojo also good point on bindata - I noticed the same.

PR #558, opened 2023-07-05
CCO-392: BUGFIX Allow for Secret to exist in AWS STS case

bentito:
Update Secret if it exists, or create if it doesn't.

The update part is the fix, wasn't allowing for existing Secrets before, but just creating new ones.


stevekuznetsov:
I have this in #545  - any chance we can hold this?

bentito:
This is covered in #545. closing.

PR #559, opened 2023-07-06
CCO-363: Adding azure identity webhook

tsorya:
1. Change aws pod identity controller to be a generic one
2. Common files moved to common folder in bindata
3. Adding azure webhook deployment
4. Don't add azure webhook in case azure credentials are not added


abutcher:
Note that once we have the OSE based azure pod identity webhook image we will set it [here](https://github.com/openshift/cloud-credential-operator/blob/19fee252a8bc6bda17ce525643f3cf031160ff00/manifests/03-deployment.yaml#L72) alongside the AWS image.

pkg/operator/podidentity/podidentitywebhook_controller.go
```diff
@@ -193,11 +216,12 @@ func Add(mgr, rootCredentialManager manager.Manager, kubeconfig string) error {
 
 	// Create a namespace local cache separate from the Manager cache
 	// A namespace scoped cache can still handle cluster scoped resources
-	cache, err := cache.New(config, cache.Options{Namespaces: namespaces})
+	controllerCache, err := cache.New(config, cache.Options{Namespaces: namespaces})
```
stevekuznetsov, line 219: Why did this get re-named?

pkg/operator/podidentity/podidentitywebhook_controller.go
```diff
@@ -209,7 +233,7 @@ func Add(mgr, rootCredentialManager manager.Manager, kubeconfig string) error {
 		if !ok {
 			return fmt.Errorf("failed to convert runtime.Object to client.Object")
 		}
-		informer, err := cache.GetInformer(context.TODO(), co)
+		informer, err := controllerCache.GetInformer(context.TODO(), co)
```
stevekuznetsov, line 236: Thread through a context instead of using `context.TODO()` - this might be pre-existing code but if everyone pitches in to help refactor this we can squash the tech debt faster.

pkg/operator/podidentity/podidentitywebhook_controller.go
```diff
@@ -293,8 +318,8 @@ func (r *staticResourceReconciler) ReconcileResources(ctx context.Context) error
 		r.logger.Infof("%s reconciled successfully", result.Type)
 	}
 
-	// "v4.1.0/aws-pod-identity-webhook/deployment.yaml"
-	requestedDeployment := resourceread.ReadDeploymentV1OrDie(v410_00_assets.MustAsset("v4.1.0/aws-pod-identity-webhook/deployment.yaml"))
+	// "v4.1.0/<platform_type_files>/deployment.yaml"
```
stevekuznetsov, line 321: If you are closing over the source of the deployment manifest, you can't assume paths like this outside of the implementations. Alternatively, consider removing the interface entirely and just building these paths dynamically. Not sure the interface adds a huge amount of value as it's written today.

pkg/operator/podidentity/podidentitywebhook_controller.go
```diff
@@ -325,8 +350,8 @@ func (r *staticResourceReconciler) ReconcileResources(ctx context.Context) error
 		// Don't deploy the PDB to single replica topology clusters
 		r.logger.Debugf("not deploying PodDisruptionBudget to single replica topology")
 	} else {
-		// "v4.1.0/aws-pod-identity-webhook/poddisruptionbudget.yaml"
-		requestedPDB := resourceread.ReadPodDisruptionBudgetV1OrDie(v410_00_assets.MustAsset("v4.1.0/aws-pod-identity-webhook/poddisruptionbudget.yaml"))
+		// "v4.1.0/common/poddisruptionbudget.yaml"
+		requestedPDB := resourceread.ReadPodDisruptionBudgetV1OrDie(v410_00_assets.MustAsset(pdb))
```
stevekuznetsov, line 354: nit: if `pdb` is used once and only once, the indirection of having a constant is not worth it in my opinion

pkg/operator/podidentity/podidentitywebhook_controller_test.go
```diff
@@ -0,0 +1,317 @@
...
+func init() {
+	log.SetLevel(log.DebugLevel)
+}
+
+func TestPodIdentityWebhookController(t *testing.T) {
+	schemeutils.SetupScheme(scheme.Scheme)
+
+	getDeployment := func(c kubernetes.Interface, name, namespace string) *appsv1.Deployment {
+		deployment, err := c.AppsV1().Deployments(namespace).Get(context.TODO(), name, v1.GetOptions{})
+		if err == nil {
```
stevekuznetsov, line 59: I imagine when someone breaks this test they will want to see the content of `err` when it's not `nil`

pkg/operator/podidentity/podidentitywebhook_controller_test.go
```diff
@@ -0,0 +1,317 @@
...
+			return webhook
+		}
+		l, err := c.AdmissionregistrationV1().MutatingWebhookConfigurations().List(context.TODO(), v1.ListOptions{})
+		fmt.Println("BBBBBBBBBBB", l, err)
+		fmt.Println("AAAAAAAAAAAAAAAA")
+		return nil
+	}
+
+	t.Setenv("AWS_POD_IDENTITY_WEBHOOK_IMAGE", "aws_identity_image")
```
stevekuznetsov, line 84: Can we read these at startup and pass them in to the controller as parameters? `os.Setenv` and `os.Getenv` are not thread-safe and while it's not an issue in this code right now it's close to a sharp edge case, if we read at startup we can be more confident in reasoning about the downstream code.

pkg/operator/podidentity/podidentitywebhook_controller_test.go
```diff
@@ -0,0 +1,317 @@
...
+		podIdentityObj: AzurePodIdentity{},
+		expectedResult: true,
+		expectErr:      false,
+	},
+		{
+			name: "Azure secret with credentials without tenant id",
+			existing: []runtime.Object{
+				&corev1.Secret{
+					ObjectMeta: v1.ObjectMeta{
+						Name: azureCredsSecretName,
```
stevekuznetsov, line 268: These fixtures don't have a `namespace` - why does this work?

pkg/operator/podidentity/podidentitywebhook_controller_test.go
tsorya: Setenv is used only for testing so i don't think we will have any problems with Getenv here.

pkg/operator/podidentity/podidentitywebhook_controller_test.go
```diff
@@ -0,0 +1,317 @@
...
+			}
+
+			if !test.expectErr {
+				expectedImage := fmt.Sprintf("%s_identity_image", strings.ToLower(string(test.platformType)))
+				assert.Equal(t, r.imagePullSpec, expectedImage)
+
+				podIdentityWebhookDeployment := getDeployment(fakeClientset, "pod-identity-webhook", "openshift-cloud-credential-operator")
+				assert.NotNil(t, podIdentityWebhookDeployment, "did not find expected pod-identity-webhook Deployment")
+
+				if test.expectedReplicas != 0 {
```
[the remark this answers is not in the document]
tsorya: Mostly this test was written before and i changed its filename, so i added only the parts i change to say the truth

pkg/operator/podidentity/podidentitywebhook_controller.go
tsorya: Actually i prefer to remove the comments, i assume interface can be helpful in future when more changes will be between webhooks

pkg/operator/podidentity/podidentitywebhook_controller.go
tsorya: It is used twice so i think it is worth const.

pkg/operator/podidentity/podidentitywebhook_controller_test.go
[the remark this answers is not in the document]
tsorya: I assume it only should validate it exists and i added validation that it matches type that i expect

pkg/operator/podidentity/podidentitywebhook_controller_test.go
tsorya: As i don't provide it to shouldBeDeployed, i will add namespace in case it helps

abutcher:
- Validated that AWS webhook continues to mutate pods in STS clusters.
- Validated that Azure webhook can be used following [procedure](https://issues.redhat.com/browse/CCO-363?focusedId=22722950&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22722950) using [quick-start](https://azure.github.io/azure-workload-identity/docs/quick-start.html).

PR #560, opened 2023-07-07
reconcile status when the clusteroperator changes

deads2k:
Found this while investigating https://github.com/openshift/cluster-kube-apiserver-operator/pull/1423.

If status is modified, the operator must set status to proper values. This ensures that accidents cannot permanently reset status and gives a clear indication that the operator is "live".  This came up as important when operators were NOT live, during cert rotations and we had no indication of problems.


deads2k:
Appears to work properly. Need to check audit logs to ensure that it's not hotloop writing before merge.

PR #563, opened 2023-07-12
OCPBUGS-15365: *: use correct clients in the secretannotator

stevekuznetsov:
We were (again) accidentally watching all ConfigMaps in the cluster. This patch should fix the errant code by using the correct client as well as by removing the RBAC to do so in the first place.


pkg/operator/secretannotator/aws/reconciler.go
```diff
@@ -75,12 +75,12 @@ func Add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 		},
 	}
-	err = c.Watch(source.Kind(mgr.GetCache(), &corev1.Secret{}), &handler.EnqueueRequestForObject{}, p)
+	err = c.Watch(source.Kind(rootCredMgr.GetCache(), &corev1.Secret{}), &handler.EnqueueRequestForObject{}, p)
 	if err != nil {
 		return err
 	}
 
-	err = secretutils.WatchCCOConfig(c, types.NamespacedName{
+	err = secretutils.WatchCCOConfig(mgr.GetCache(), c, types.NamespacedName{
```
stevekuznetsov, line 83: This is a bit tricky - we're triggering reconciliation in the controller (based on the root cred manager) by using events for the CCO config that come from the other manager's cache. We did the same in the actuators but in the other direction, one cache with configmap events driving the other reconciler.

PR #564, opened 2023-07-12
OCPBUGS-15906: ccoctl azure delete to also delete role assignments

jstuever:
The ccoctl azure subcommand was previously leaking role assignments. This change ensures all associated roleAssignments are removed prior to deleting the user-assigned managed identities.

### DIFF
--- a/pkg/cmd/provisioning/azure/delete.go
+++ b/pkg/cmd/provisioning/azure/delete.go
@@ -108,7 +108,11 @@ func deleteManagedIdentities(client *azureclients.AzureClientWrapper, name, reso
 		return nil
 	}
 	for _, identity := range managedIdentities {
-		_, err := client.UserAssignedIdentitiesClient.Delete(
+		err := deleteRoleAssignmentsByPrincipal(client, *identity.Properties.PrincipalID, subscriptionID)
+		if err != nil {
+			return err
+		}
+		_, err = client.UserAssignedIdentitiesClient.Delete(
 			context.Background(),
 			resourceGroupName,
 			*identity.Name,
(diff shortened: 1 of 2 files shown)


pkg/cmd/provisioning/azure/delete.go
```diff
@@ -80,6 +81,14 @@ func deleteManagedIdentities(client *azureclients.AzureClientWrapper, name, reso
 	for listManagedIdentities.More() {
 		pageResponse, err := listManagedIdentities.NextPage(context.Background())
 		if err != nil {
+			var respErr *azcore.ResponseError
+			if !(errors.As(err, &respErr)) {
+				return err
+			}
+			if respErr.ErrorCode == "ResourceGroupNotFound" {
+				log.Printf("Found no user-assigned managed identities in resource group %s", resourceGroupName)
```
abutcher, line 89: Prefer if this said something about the resource group already being gone / having been deleted. Something like this similar to the latter `"Found no resource group %s"` log.

```suggestion
				log.Printf("Found no resource group %s. No user-assigned managed identities to list", resourceGroupName)
```

abutcher (commented):
Works in my testing! One comment about the info log for no identities if that could be more specific to the RG not existing. Looks good

pkg/cmd/provisioning/azure/delete.go
jstuever: [Done](https://github.com/openshift/cloud-credential-operator/compare/cc1d63226f953c4ada6539888746cf4cbe463e1f..f83c6b70da024ca5e159d2515ab149227bad58d9)

PR #568, opened 2023-07-14
OCPBUGS-16334: Clarify updateActuatorConditions & update lastSync

bentito:
For `updateActuatorConditions()` usage, see inline comments please, basically this PR just cleans up the function to hopefully show conditions are updated properly.
 
`cr.Status.LastSyncTimestamp` is updated now.

### DIFF
--- a/pkg/operator/credentialsrequest/credentialsrequest_controller.go
+++ b/pkg/operator/credentialsrequest/credentialsrequest_controller.go
@@ -711,6 +711,11 @@ func (r *ReconcileCredentialsRequest) Reconcile(ctx context.Context, request rec
 
 				logger.Errorf("errored with condition: %v", t.Reason())
 				r.updateActuatorConditions(cr, t.Reason(), syncErr)
+				// Update the status of the CredentialsRequest object
+				err := r.Client.Status().Update(ctx, cr)
+				if err != nil {
+					logger.Errorf("failed to update credentialsrequest status: %v", err)
+				}
 			default:
 				logger.Errorf("unexpected error while syncing credentialsrequest: %v", syncErr)
 				return reconcile.Result{}, syncErr
--- a/pkg/operator/credentialsrequest/credentialsrequest_controller.go
+++ b/pkg/operator/credentialsrequest/credentialsrequest_controller.go
@@ -836,28 +836,28 @@ func (r *ReconcileCredentialsRequest) CreateOrUpdateOnCredsExist(ctx context.Con
 }
 
 func (r *ReconcileCredentialsRequest) updateActuatorConditions(cr *minterv1.CredentialsRequest, reason minterv1.CredentialsRequestConditionType, conditionError error) {
-	// Clear all existing conditions first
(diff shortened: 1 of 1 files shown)


pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito, line 1: https://github.com/openshift/cloud-credential-operator/pull/568/files#diff-ba64ecf2acf81969711010121ce3acdef206490c0d5aa1b9ee03a51321c6bdccR713 we're calling `updateActuatorConditions`. This PR cleans up that function and hopefully makes it clear we are setting the actuator conditions properly?

fxierh: Sorry for the confusion in forum-cloud-credential-operator. 

What I meant is, after CR.status.conditions are properly set by `updateActuatorConditions`, we should IMO still call r.Client.Status().Update(ctx, cr) after that.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -711,6 +711,11 @@ func (r *ReconcileCredentialsRequest) Reconcile(ctx context.Context, request rec
 
 				logger.Errorf("errored with condition: %v", t.Reason())
 				r.updateActuatorConditions(cr, t.Reason(), syncErr)
+				// Update the status of the CredentialsRequest object
+				err := r.Client.Status().Update(ctx, cr)
```
bentito, line 715: @fxierh This is where you meant to call Update, right?

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -831,30 +836,30 @@ func (r *ReconcileCredentialsRequest) CreateOrUpdateOnCredsExist(ctx context.Con
 }
 
 func (r *ReconcileCredentialsRequest) updateActuatorConditions(cr *minterv1.CredentialsRequest, reason minterv1.CredentialsRequestConditionType, conditionError error) {
+	// Clear all existing conditions first
+	clearAllConditions(cr)
```
abutcher, line 840: I think we should revert to the original logic in `updateActuatorConditions`. The outcome of this switch case will be the same in terms of which conditions are set but we'll lose fields like `lastTransitionTime` should a condition remain the same.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
fxierh: Yep

pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito: k reverting logic there

abutcher:
Going to need a bug card for this one.

bentito:
closing and reopening to hopefully get e2e-hypershift test unstuck.

PR #570, opened 2023-07-17
Revert "manifests: manually amend resources that get laid down by CVO in HCP"

enxebre:
This reverts commit 3b4d1e2edc904c9215e69d0002c29bbed4d7bf62.


dgoodwin:
/override ci/prow/e2e-upgrade

Going back to known good state, high confidence from Hypershift folks this is the culprit which broke payload stream.

PR #573, opened 2023-07-18
CCO-402: Create Azure AD pod identity webhook config secret manifest in create-all,create-oidc-issuer

abutcher:
This requires providing the Azure tenant ID to put within the webhook config in create-all and create-oidc-issuer subcommands which will create the webhook config. The Azure tenant ID is not made available/discoverable by azure-sdk-for-go and the method for knowing the tenant ID depends on the authentication method used. Because of this, we will now need to require the tenant ID as a parameter and only use it as a value to be set in the webhook config.

e2e-azure-manual-oidc will need to be modified to account for this new required command line parameter.

### DIFF
--- a/docs/azure_workload_identity.md
+++ b/docs/azure_workload_identity.md
@@ -88,6 +88,7 @@ type: Opaque
                              --output-dir <ccoctl_output_dir> \
                              --region <azure_region> \
                              --subscription-id <azure_subscription_id> \
+                             --tenant-id <azure_tenant_id> \
                              --credentials-requests-dir /path/to/credreqs/directory/created/in/step/3 \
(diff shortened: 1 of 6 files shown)


abutcher:
/cc @jstuever

Somewhat significant change to the client set up with the goal of knowing the Azure tenant ID when we create the OIDC issuer. I need to verify how this interacts with the AZURE_TENANT_ID environment variable existing. I have a hunch that the command line arg obviates it needing to be set since it gets passed into the client options object.

abutcher:
Think I'm ready to send this.

/assign @jstuever

jstuever:
/lgtm
/hold
for e2e-azure-manual-oidc

PR #576, opened 2023-07-20
*: use a global codec

stevekuznetsov:
pkg/operator: correctly fetch CA for AWS minter

We just need to read *one* and only *one* ConfigMap, so add a Role to
let us and use a live client to just read the thing.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

cmd: add pprof endpoint

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

*: use a global codec

The codec *is* global and only ever changes at compile-time, so
generally the pattern is to declare it as a package global and
intiialize it in an init() method. 30% of our runtime allocations were
building up the scheme/codec on-demand.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

### DIFF
--- a/manifests/03-deployment.yaml
+++ b/manifests/03-deployment.yaml
@@ -81,7 +81,7 @@ spec:
         resources:
           requests:
             cpu: 10m
-            memory: 150Mi
+            memory: 20Mi
         terminationMessagePolicy: FallbackToLogsOnError
         volumeMounts:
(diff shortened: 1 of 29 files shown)


stevekuznetsov:
Depends on https://github.com/openshift/cloud-credential-operator/pull/575

PR #582, opened 2023-07-26
OCPBUGS-16807: ccoctl azure exit with error when OIDC and installation resource group names are the same

abutcher:
ccoctl should provide an error when installation and OIDC resource group names are the same to prevent ccoctl from creating resources in the resource group meant to be used for cluster install. The OpenShift installer requires that the resource group in the install-config.yaml be empty when running the installation.


jstuever:
/lgtm
/hold
for e2e-azure-manual-oidc

PR #583, opened 2023-07-26
OCPBUGS-16684: Set cr.status.provisioned=false on syncErr path

bentito:
I think this cures the bug as long as I tracked from the correct part of the code that was setting `cr.status.provisioned` to true despite a failure to actually provision the Secret.

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -341,7 +341,10 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
 		}
 		if awsSTSIAMRoleARN == "" {
 			logger.Debug("CredentialsRequest has no awsSTSIAMRoleARN, no reason to sync")
-			return err
+			return &actuatoriface.ActuatorError{
+				ErrReason: minterv1.CredentialsProvisionFailure,
+				Message:   "an empty roleARN was found so no Secret was created",
+			}
 		}
 		cloudTokenPath := cr.Spec.CloudTokenPath
 		if cr.Spec.CloudTokenPath == "" {
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -341,7 +341,7 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
 		}
 		if awsSTSIAMRoleARN == "" {
 			logger.Debug("CredentialsRequest has no awsSTSIAMRoleARN, no reason to sync")
-			return nil
+			return err
 		}
 		cloudTokenPath := cr.Spec.CloudTokenPath
 		if cr.Spec.CloudTokenPath == "" {
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -343,7 +343,7 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
 			logger.Debug("CredentialsRequest has no awsSTSIAMRoleARN, no reason to sync")
 			return &actuatoriface.ActuatorError{
 				ErrReason: minterv1.CredentialsProvisionFailure,
-				Message:   "an empty roleARN was found so no Secret was created",
+				Message:   "an empty awsSTSIAMRoleARN was found so no Secret was created",
 			}
 		}
 		cloudTokenPath := cr.Spec.CloudTokenPath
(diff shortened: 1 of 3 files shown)


bentito:
@fxierh I think I could have fixed it with this change, but I'm not totally sure. If you can run this PR, set conditions again, and we'll see if it is fixed. If it's not fixed, I'll see how I can debug on this PR, perhaps with the STS workflow with clusterbot.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -688,20 +688,17 @@ func (r *ReconcileCredentialsRequest) Reconcile(ctx context.Context, request rec
 
 		var syncErr error
 		syncErr = r.CreateOrUpdateOnCredsExist(ctx, credsExists, syncErr, cr)
```
fxierh, line 690: When CR.spec.providerSpec.stsIAMRoleARN is unset, syncErr is still nil
https://github.com/openshift/cloud-credential-operator/blob/859de532440846fabddf5dd5a07417a45b47b7fb/pkg/aws/actuator/actuator.go#L342-L345 

So we fall into the else branch below where we call
https://github.com/openshift/cloud-credential-operator/blob/859de532440846fabddf5dd5a07417a45b47b7fb/pkg/operator/credentialsrequest/credentialsrequest_controller.go#L724

In other words CR.status.provisioned is still set to true when CR.spec.providerSpec.stsIAMRoleARN is unset and there is no Secret provisioned.

fxierh: The comment above aligns with what I observed today in a test (on this PR).

fxierh:
For debugging I guess you can try to scale down CVO & CCO then run the controllers locally.

fxierh:
Btw just wanted to check if you're planning to address the comment I left under https://issues.redhat.com/browse/OCPBUGS-16684 in this PR ?

bentito:
> For debugging IMO you can try to install an STS cluster, scale down CVO & CCO and then run the controllers locally with the latest commit.

Yeah, that's basically what's happening with clusterbot, but with automation, but it's way slow for a quick test to see if I got something wrong or right.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
bentito: Thanks! I think my next commit fixes it then.

pkg/aws/actuator/actuator.go
```diff
@@ -341,7 +341,7 @@ func (a *AWSActuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest)
 		}
 		if awsSTSIAMRoleARN == "" {
 			logger.Debug("CredentialsRequest has no awsSTSIAMRoleARN, no reason to sync")
-			return nil
```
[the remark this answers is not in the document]
bentito: Yep, new commit, didn't think that through very well. Thanks!

fxierh:
@bentito With the current implementation the .status field of CRs without .stsIAMRoleARN (including the ones extracted via `oc adm release extract $IMG --cloud=aws --credentials-requests`) looks like the following:
```
status:
  conditions:
  - lastProbeTime: "2023-07-28T16:08:52Z"
    lastTransitionTime: "2023-07-28T16:08:52Z"
    message: 'failed to grant creds: an empty roleARN was found so no Secret was created'
    reason: CredentialsProvisionFailure
    status: "True"
    type: CredentialsProvisionFailure
  lastSyncGeneration: 0
  lastSyncTimestamp: "2023-07-28T16:16:09Z"
  provisioned: false
```
 Just wanted to make sure this is expected or not.

fxierh:
On the other hand, the aforementioned failing condition (`CredentialsProvisionFailure`) is not reflected at the CO level (i.e. we won't see "%d of %d credentials requests are failing to sync" when calling `oc get co cloud-credential -o yaml`:

https://github.com/openshift/cloud-credential-operator/blob/859de532440846fabddf5dd5a07417a45b47b7fb/pkg/operator/credentialsrequest/status.go#L107-L108

This makes me wonder if we should display Secret provision failures on CO/cloud-credential in STS mode just like in mint.
cc. @abutcher

fxierh:
> If you could make a new card about the `lastSyncGeneration` problems that would be good. Thanks.

Done.

bentito:
>Just wanted to make sure this is expected behavior.
@fxierh Yes, I think this is as desired...
And if so, @abutcher , I think this can get an LGTM?

abutcher:
I'll make a card for the CO level status.
/lgtm

bentito:
@abutcher `ci/prow/e2e-azure-manual-oidc` is showing up as a required test here and it definitely shouldn't be here on an AWS related PR I'd think. Or at least not as a required...

abutcher:
Related to a recent change we made to the tests which should have made e2e-azure-manual-oidc required for changes which impact specific pathing but are instead making the test required.

/override ci/prow/e2e-azure-manual-oidc

PR #585, opened 2023-08-01
OCPBUGS-17049: update lastSyncGeneration in STS flow sync success

bentito:
Should now be properly setting the `lastSyncGeneration` as per the non-STS flow.


fxierh:
@bentito Looks like r.Client.Status().Update(context.Background(), cr) is not called on the cr object with lastSyncGeneration updated.

bentito:
> @bentito Looks like r.Client.Status().Update(context.Background(), cr) is not called on the cr object with lastSyncGeneration updated.

@fxierh yep, thanks, added the same call to the utility function UpdateStatus() that the non-STS flow uses.

newtonheath:
@bentito - we ready?

bentito:
> @bentito - we ready?

if it looks good to @abutcher then yes

PR #587, opened 2023-08-02
CCO-372: Azure Workload Identity info in CredsRequests creates a Secret

bentito:
Adds Azure Workload Identity (Timed Access Token) support and e2e test and unit tests.

`pkg/azure/actuator_test.go` should get a review for changes that were needed to pre-existing unit tests. The changes were needed because of the requirements of the new detection for Workload Identity in-use flow.

Known issues:

- [x] - ~Still need to verify OCPBUGS filed against~ the AWS STS work, particularly around properly setting status, conditions and lastSync times that were fixed for AWS STS are not brought back here for Azure.
      * Reviewed the several bug fix PRs against the AWS STS work, most fixes were in credentialsrequest_controller so their all present for this Azure branch as well, just OCPBUGS-16684 had an actuator change that seemed to be needed here.
- [x] - This should be e2e testable with a PR in `release` which actually calls the new Make target: `test-e2e-azident` with the `e2e-azure-manual-oidc` workflow
      * PR to add calling make target to azure e2e is here: https://github.com/openshift/release/pull/42126

### DIFF
--- a/pkg/apis/cloudcredential/v1/types_azure.go
+++ b/pkg/apis/cloudcredential/v1/types_azure.go
@@ -47,6 +47,24 @@ type AzureProviderSpec struct {
 	// and RoleBindings.
 	// +optional
 	DataPermissions []string `json:"dataPermissions,omitempty"`
+
+	// The following fields are only required for Azure Workload Identity.
+	// AzureClientID is the ID of the specific application you created in Azure
+	// +optional
+	AzureClientID string `json:"azureClientID,omitempty"`
+
+	// AzureRegion is the geographic region of the Azure service.
+	// +optional
+	AzureRegion string `json:"azureRegion,omitempty"`
+
+	// Each Azure subscription has an ID associated with it, as does the tenant to which a subscription belongs.
+	// AzureSubscriptionID is the ID of the subscription.
+	// +optional
+	AzureSubscriptionID string `json:"azureSubscriptionID,omitempty"`
+
+	// AzureTenantID is the ID of the tenant to which the subscription belongs.
+	// +optional
+	AzureTenantID string `json:"azureTenantID,omitempty"`
 }
 
 // RoleBinding models part of the Azure RBAC Role Binding
--- a/pkg/azure/actuator.go
+++ b/pkg/azure/actuator.go
@@ -36,6 +36,7 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 
 	minterv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+	"github.com/openshift/cloud-credential-operator/pkg/cmd/provisioning"
 	"github.com/openshift/cloud-credential-operator/pkg/operator/constants"
 	actuatoriface "github.com/openshift/cloud-credential-operator/pkg/operator/credentialsrequest/actuator"
 	"github.com/openshift/cloud-credential-operator/pkg/operator/utils"
@@ -68,20 +69,6 @@ func NewFakeActuator(c, rootCredClient client.Client,
 	}
 }
 
-func (a *Actuator) IsValidMode() error {
-	mode, err := a.client.Mode(context.Background())
-	if err != nil {
-		return err
-	}
-
-	switch mode {
-	case constants.PassthroughAnnotation:
-		return nil
-	default:
-		return errors.New("invalid mode")
-	}
-}
-
 func isAzureCredentials(providerSpec *runtime.RawExtension) (bool, error) {
 	var err error
 	unknown := runtime.Unknown{}
@@ -107,6 +94,19 @@ func (a *Actuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequ
 	}
 	if !exists {
 		return true, nil
+	}
+
+	// Manual mode just update
+	credentialsMode, _, err := utils.GetOperatorConfiguration(a.client, logger)
+	if err != nil {
+		logger.WithError(err).Error("error loading CCO configuration to determine valid mode")
+		return true, err
+	}
+	if credentialsMode == operatorv1.CloudCredentialsModeManual {
+		return true, nil
+	}
+	if credentialsMode == operatorv1.CloudCredentialsModeMint {
+		return false, errors.New("mint mode is invalid")
 	}
 	// passthrough-specifc checks here (now the only kind of checks...)
 
@@ -160,11 +160,17 @@ func (a *Actuator) Delete(ctx context.Context, cr *minterv1.CredentialsRequest) 
 	if isAzure, err := isAzureCredentials(cr.Spec.ProviderSpec); !isAzure {
 		return err
 	}
-	if err := a.IsValidMode(); err != nil {
+	logger := a.getLogger(cr)
+	credentialsMode, _, err := utils.GetOperatorConfiguration(a.client, logger)
+	if err != nil {
+		logger.WithError(err).Error("error loading CCO configuration to determine valid mode")
 		return err
 	}
+	if credentialsMode == operatorv1.CloudCredentialsModeManual {
+		logger.Debug("running delete in manual mode")
+		return nil
+	}
 
-	logger := a.getLogger(cr)
 	logger.Debug("running delete")
 
 	credentialsRootSecret, err := a.GetCredentialsRootSecret(ctx, cr)
@@ -230,6 +236,52 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
 		logger.Debug("credentials already up to date")
 		return nil
 	}
+	stsDetected, err := utils.IsTimedTokenCluster(a.client, ctx, logger)
+	if err != nil {
+		return err
+	}
+	if stsDetected {
+		logger.Debug("actuator detected Azure AD Workload Identity enabled cluster, enabling Workload Identity secret brokering for CredentialsRequests providing a Managed Identity")
+		azureProviderSpec, err := decodeProviderSpec(minterv1.Codec, cr)
+		if err != nil {
+			return err
+		}
+		azureFederatedTokenFile := cr.Spec.CloudTokenPath
+		if cr.Spec.CloudTokenPath == "" {
+			logger.Debugf("CredentialsRequest has no cloudTokenPath, defaulting azure_federated_token_file to %s", provisioning.OidcTokenPath)
+			azureFederatedTokenFile = provisioning.OidcTokenPath
+		}
+		// Check for old Manual Mode where all 4 fields are empty - defaulting to old behavior
+		// where CCO exists and the secret is created manually
+		if azureProviderSpec.AzureClientID == "" && azureProviderSpec.AzureTenantID == "" && azureProviderSpec.AzureSubscriptionID == "" && azureProviderSpec.AzureRegion == "" {
+			return nil
+		}
+		err = validateAzureProviderSpec(*azureProviderSpec)
+		if err != nil {
+			// At least one of the fields was set indicating that the new workload identity
+			// behavior of creating the secret is desired but not all fields required were
+			// provided.
+			msg := "error validating credentials request Azure AD Workload Identity fields"
+			return &actuatoriface.ActuatorError{
+				ErrReason: minterv1.CredentialsProvisionFailure,
+				Message:   fmt.Sprintf("%v: %v", msg, err),
+			}
+		}
+		desiredSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      cr.Spec.SecretRef.Name,
+				Namespace: cr.Spec.SecretRef.Namespace,
+			},
+			StringData: map[string]string{
+				AzureClientID:           azureProviderSpec.AzureClientID,
+				AzureTenantID:           azureProviderSpec.AzureTenantID,
+				AzureRegion:             azureProviderSpec.AzureRegion,
+				AzureSubscriptionID:     azureProviderSpec.AzureSubscriptionID,
+				AzureFederatedTokenFile: azureFederatedTokenFile,
+			},
+		}
+		return a.syncCredentialSecrets(ctx, cr, desiredSecret, logger)
+	}
 
 	credentialsRootSecret, err := a.GetCredentialsRootSecret(ctx, cr)
 	if err != nil {
@@ -259,7 +311,6 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
 			Message:   msg,
 		}
 	}
-
 	return nil
 }
 
@@ -370,7 +421,25 @@ func (a *Actuator) cleanupAfterPassthroughPivot(ctx context.Context, cr *minterv
 
 	return nil
 }
-func (a *Actuator) syncCredentialSecrets(ctx context.Context, cr *minterv1.CredentialsRequest, cloudCredsSecret *corev1.Secret, logger log.FieldLogger) error {
+
+// For Azure Workload Identity, the generated Secret needs to look like this:
+/*
+	apiVersion: v1
+	stringData:
+	  azure_client_id: 0420bfd1-ab26-4b80-a9ac-deadbeeff1f9
+	  azure_tenant_id: 6047c7e9-b2ad-488d-a54e-deadbeefa7ee
+	  azure_region: centralus
+	  azure_subscription_id: 8c20ec23-8478-4f46-96f4-deadbeeff185
+	  azure_federated_token_file: /var/run/secrets/openshift/serviceaccount/token
+	kind: Secret
+	metadata:
+	  name: azure-cloud-credentials
+	  namespace: openshift-machine-api
+	type: Opaque
+*/
+// The first 4 fields need to come from `spec.ProviderSpec` in the CredentialsRequest with
+// spec.cloudTokenPath matching up to: `azure_federated_token_file`
+func (a *Actuator) syncCredentialSecrets(ctx context.Context, cr *minterv1.CredentialsRequest, desiredSecret *corev1.Secret, logger log.FieldLogger) error {
 	sLog := logger.WithFields(log.Fields{
 		"targetSecret": fmt.Sprintf("%s/%s", cr.Spec.SecretRef.Namespace, cr.Spec.SecretRef.Name),
 		"cr":           fmt.Sprintf("%s/%s", cr.Namespace, cr.Name),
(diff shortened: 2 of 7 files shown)


pkg/azure/actuator.go
```diff
@@ -109,43 +103,56 @@ func (a *Actuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequ
...
+		}
+
+		if azureStatus.ServicePrincipalName != "" {
+			return true, nil
+		}
 
-	return false, nil
+		return false, nil
+	case operatorv1.CloudCredentialsModeManual:
+		// not sure
```
gallettilance, line 151: @abutcher how should the updates behave in Manual mode?

pkg/apis/cloudcredential/v1/types_azure.go
```diff
@@ -40,6 +40,14 @@ type AzureProviderSpec struct {
...
 	Permissions []string `json:"permissions,omitempty"`
 
+	AzureClientID string `json:"azureClientID"`
+
+	AzureRegion string `json:"azureRegion"`
+
+	AzureSubscriptionID string `json:"azureSubscriptionID"`
+
+	AzureTenantID string `json:"azureTenantID"`
+
```
jstuever, line 50: These need the following:
- comments explaining their usage
- `+optional`
- `omitempty`

pkg/azure/actuator.go
```diff
@@ -109,43 +103,52 @@ func (a *Actuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequ
...
+
+		if azureStatus.ServicePrincipalName != "" {
+			return true, nil
+		}
 
-	return false, nil
+		return false, nil
+	} else {
+		return false, nil
+	}
```
jstuever, line 151: This can be flattened....
```diff
-		return false, nil
-	} else {
-		return false, nil
	}
+	return false, nil

```

pkg/azure/actuator.go
```diff
@@ -230,36 +230,56 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
...
 	}
-
-	switch credentialsRootSecret.Annotations[constants.AnnotationKey] {
-	case constants.InsufficientAnnotation:
-		msg := "cloud credentials insufficient to satisfy credentials request"
-		logger.Error(msg)
-		return &actuatoriface.ActuatorError{
-			ErrReason: minterv1.InsufficientCloudCredentials,
-			Message:   msg,
+	logger.Infof("stsDetected: %v", stsDetected)
```
jstuever, line 237: Can we make this output more customer friendly, or drop it altogether?

pkg/azure/actuator.go
```diff
@@ -230,36 +230,56 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
...
 			return err
 		}
-	default:
-		msg := fmt.Sprintf("unexpected value or missing %s annotation on admin credentials Secret", constants.AnnotationKey)
-		logger.Info(msg)
-		return &actuatoriface.ActuatorError{
-			ErrReason: minterv1.CredentialsProvisionFailure,
-			Message:   msg,
+	} else {
+		credentialsRootSecret, err := a.GetCredentialsRootSecret(ctx, cr)
```
jstuever, line 254: This can be flattened...
```diff
-		if err != nil {
-			return err
-		}
+		return err
-	} else {
-		credentialsRootSecret, err := a.GetCredentialsRootSecret(ctx, cr)
+	credentialsRootSecret, err := a.GetCredentialsRootSecret(ctx, cr)
<truncated>
```

pkg/azure/actuator_test.go
```diff
@@ -245,6 +246,7 @@ func TestActuator(t *testing.T) {
 				}
 
 				objects = append(objects, targetSecret)
+				objects = append(objects, testOperatorConfig(operatorv1.CloudCredentialsModeMint))
```
jstuever, line 249: Mint mode is an invalid mode for Azure; the default is manual mode.
https://github.com/openshift/installer/blob/d7138a1f324432147ad0b1f5baf43d1db5b4249a/pkg/types/defaults/installconfig.go#L86-L87

pkg/azure/actuator_test.go
```diff
@@ -406,6 +409,7 @@ func TestActuator(t *testing.T) {
...
```
jstuever, line 412: Mint mode is an invalid mode for Azure; the default is manual mode.
https://github.com/openshift/installer/blob/d7138a1f324432147ad0b1f5baf43d1db5b4249a/pkg/types/defaults/installconfig.go#L86-L87

pkg/azure/actuator_test.go
```diff
@@ -521,6 +526,153 @@ func TestActuator(t *testing.T) {
...
+				cr := testCredentialsRequest(t)
+
+				azureSpec := &minterv1.AzureProviderSpec{}
+				err := minterv1.Codec.DecodeProviderSpec(cr.Spec.ProviderSpec, azureSpec)
+				require.NoError(t, err, "error decoding provider spec")
+
+				azureSpec.AzureClientID = "0420bfd1-ab26-4b80-a9ac-deadbeeff1f9"
+				azureSpec.AzureTenantID = "6047c7e9-b2ad-488d-a54e-deadbeefa7ee"
+				azureSpec.AzureRegion = "centralus"
+				azureSpec.AzureSubscriptionID = "/var/run/secrets/openshift/serviceaccount/token"
```
jstuever, line 574: Can these be changed to `fake-*` like the data map above?

jstuever (changes requested):
I have some concerns regarding the removal of `IsValidMode()`; I am investigating the implications of doing so. For now, I am recommending a few changes. Also, the `prow` and `unit` tests are failing and need to be resolved.

jstuever:
> I have some concerns regarding the removal of `IsValidMode()`; I am investigating the implications of doing so. For now, I am recommending a few changes. Also, the `prow` and `unit` tests are failing and need to be resolved.

I think there is merit in continuing parallel logic where `IsValidMode()` was being used before. That is to say, I think we should continue returning from these locations in the various functions where credentialsMode is `Mint` or where mode is `Manual` (or undefined) and STS is not detected. This will create the parallel where these functions only continue in `Passthrough` or `STS` mode.

pkg/azure/actuator.go
```diff
@@ -77,6 +86,8 @@ func (a *Actuator) IsValidMode() error {
 	switch mode {
 	case constants.PassthroughAnnotation:
 		return nil
+	case constants.MintAnnotation:
+		return errors.New("mint mode is invalid")
 	default:
```
gallettilance, line 91: Note to reviewers: I'm assuming in Manual mode there's either no secret or no need to annotate it?

gallettilance: otherwise please let me know what the manual annotation is

pkg/apis/cloudcredential/v1/types_azure.go
```diff
@@ -47,6 +47,24 @@ type AzureProviderSpec struct {
 	// and RoleBindings.
 	// +optional
 	DataPermissions []string `json:"dataPermissions,omitempty"`
+
+	// The following fields are required only for STS.
```
perdasilva, line 51: nit: STS is Amazon lingo
```suggestion
	// The following fields are only required for Azure Workload Identity.
```

pkg/apis/cloudcredential/v1/types_azure.go
```diff
@@ -47,6 +47,24 @@ type AzureProviderSpec struct {
...
+	// AzureClientID is the ID of the ID of the specific application you created in Azure
```
perdasilva, line 52:
```suggestion
	// AzureClientID is the ID of the specific application you created in Azure
```

pkg/azure/actuator.go
```diff
@@ -230,36 +241,53 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
 		logger.Debug("credentials already up to date")
 		return nil
 	}
-
-	credentialsRootSecret, err := a.GetCredentialsRootSecret(ctx, cr)
+	stsDetected, err := utils.IsTimedTokenCluster(a.client, ctx, logger)
```
perdasilva, line 244: nit on sts - maybe tokenServiceDetected?

pkg/azure/actuator.go
```diff
@@ -492,3 +520,102 @@ func (a *Actuator) getLogger(cr *minterv1.CredentialsRequest) log.FieldLogger {
...
+	sLog.Infof("processing secret")
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cr.Spec.SecretRef.Name,
+			Namespace: cr.Spec.SecretRef.Namespace,
+		},
+	}
+	err := validateAzureProviderSpec(azureProviderSpec)
+	if err != nil {
+		logger.WithError(err).Error("error determining whether a credentials update is needed")
```
perdasilva, line 559:
```suggestion
		logger.WithError(err).Error("error validating AzureProviderSpec")
```

pkg/azure/actuator.go
```diff
@@ -492,3 +520,102 @@ func (a *Actuator) getLogger(cr *minterv1.CredentialsRequest) log.FieldLogger {
...
+		msg := "error validating credentials request Azure AD Workload Identity fields"
```
perdasilva, line 560: maybe we don't need this message? the error coming from the validate method already includes everything. Wdyt?

pkg/azure/actuator.go
jstuever: Usually we handle manual mode as the default by specifying Passthrough and Mint mode in prior case statements, as you have done here. I think we just need to ensure the default case does not return an error when stsDetected. This will enable the `Exists` and `Delete` functions to function normally for stsDetected clusters.

pkg/azure/actuator.go
```diff
@@ -230,36 +241,53 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
...
+		if err != nil {
+			return err
 		}
-	case constants.PassthroughAnnotation:
-		logger.Debugf("provisioning with passthrough")
-		err := a.syncPassthrough(ctx, cr, credentialsRootSecret, logger)
+		azureFederatedTokenFile := cr.Spec.CloudTokenPath
+		if cr.Spec.CloudTokenPath == "" {
+			logger.Debug("CredentialsRequest has no cloudTokenPath, defaulting azure_federated_token_file to /var/run/secrets/kubernetes.io/serviceaccount/token")
+			azureFederatedTokenFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
jstuever, line 257: Do we want to match the default to what is used by the `ccoctl` command? 
/var/run/secrets/openshift/serviceaccount/token

https://github.com/openshift/cloud-credential-operator/blob/3c72c02d745112f07d9ccd3675ac017e864a2c76/pkg/cmd/provisioning/constants.go#L17

pkg/azure/actuator.go
```diff
@@ -492,3 +520,102 @@ func (a *Actuator) getLogger(cr *minterv1.CredentialsRequest) log.FieldLogger {
...
+		return nil
+	})
+	sLog.WithField("operation", op).Info("processed secret")
+	if err != nil {
+		return &actuatoriface.ActuatorError{
+			ErrReason: minterv1.CredentialsProvisionFailure,
+			Message:   "error processing secret",
+		}
+	}
+	return nil
```
jstuever, line 595: This is mostly identical to `syncCredentialSecrets`. I would prefer if we re-use that block so as to not duplicate code. The provider spec can be validated prior to passing in.

jstuever (changes requested):

jstuever (changes requested):
I believe there are still two significant issues to work out, both related to `IsValidMode()`

- I expect the [sync()](https://github.com/openshift/cloud-credential-operator/blob/b116fa8f3135cde45aa8c404b9a9af4f93c427d1/pkg/azure/actuator.go#L223) function is currently returning immediately after the [NeedsUpdate()](https://github.com/openshift/cloud-credential-operator/blob/b116fa8f3135cde45aa8c404b9a9af4f93c427d1/pkg/azure/actuator.go#L232-L239) call. That is because `NeedsUpdate()` calls [Exists()](https://github.com/openshift/cloud-credential-operator/blob/b116fa8f3135cde45aa8c404b9a9af4f93c427d1/pkg/azure/actuator.go#L116) which then calls [IsValidMode()](https://github.com/openshift/cloud-credential-operator/blob/b116fa8f3135cde45aa8c404b9a9af4f93c427d1/pkg/azure/actuator.go#L543-L545). It currently returns an error for all modes except PassThrough, which is passed all the way back to `sync()` where we immediately [return](https://github.com/openshift/cloud-credential-operator/blob/b116fa8f3135cde45aa8c404b9a9af4f93c427d1/pkg/azure/actuator.go#L235).
- I expect the `delete()` function will fail to remove secrets when their associated credentialRequest is removed. This is because the `delete()` function also [returns](https://github.com/openshift/cloud-credential-operator/blob/b116fa8f3135cde45aa8c404b9a9af4f93c427d1/pkg/azure/actuator.go#L175-L177) immediately when `IsValidMode()` returns an error.

gallettilance:
[cred-req-secret-creationdeletion.webm](https://github.com/openshift/cloud-credential-operator/assets/20077170/bedda29d-6150-41d6-9afd-602416013f0f)

@jstuever here is a demo of the secret being created and deleted

gallettilance:
[cred-req-workload-identity.webm](https://github.com/openshift/cloud-credential-operator/assets/20077170/3ee2bd50-0200-4b49-8521-d7592cd2588d)

Update based on our convo @jstuever 
- no errors for "old" manual mode
- updates of the secret to make it look exactly as the ccoctl tool does

pkg/azure/actuator.go
```diff
@@ -160,11 +160,17 @@ func (a *Actuator) Delete(ctx context.Context, cr *minterv1.CredentialsRequest)
...
-	if err := a.IsValidMode(); err != nil {
+	logger := a.getLogger(cr)
+	credentialsMode, _, err := utils.GetOperatorConfiguration(a.client, logger)
+	if err != nil {
+		logger.WithError(err).Error("error loading CCO configuration to determine valid mode")
 		return err
 	}
+	if credentialsMode == operatorv1.CloudCredentialsModeManual {
+		logger.Debug("running delete in manual mode")
+		return nil
```
gallettilance, line 171: Note to reviewers: the delete works because of the annotation on the secret

pkg/azure/actuator.go
```diff
@@ -108,6 +95,19 @@ func (a *Actuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequ
...
 	}
+
+	// Manual mode just update
+	credentialsMode, _, err := utils.GetOperatorConfiguration(a.client, logger)
+	if err != nil {
+		logger.WithError(err).Error("error loading CCO configuration to determine valid mode")
+		return true, err
+	}
+	if credentialsMode == operatorv1.CloudCredentialsModeManual {
+		return true, nil
```
gallettilance, line 106: Note to reviewers: what's in the credReq should be the source of truth propagated down to the secret.

pkg/azure/actuator.go
```diff
@@ -230,6 +236,52 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
...
+			return err
+		}
+		azureFederatedTokenFile := cr.Spec.CloudTokenPath
+		if cr.Spec.CloudTokenPath == "" {
+			logger.Debugf("CredentialsRequest has no cloudTokenPath, defaulting azure_federated_token_file to %s", provisioning.OidcTokenPath)
+			azureFederatedTokenFile = provisioning.OidcTokenPath
+		}
+		// Check for old Manual Mode where all 4 fields are empty - defaulting to old behavior
+		// where CCO exists and the secret is created manually
+		if azureProviderSpec.AzureClientID == "" && azureProviderSpec.AzureTenantID == "" && azureProviderSpec.AzureSubscriptionID == "" && azureProviderSpec.AzureRegion == "" {
```
gallettilance, line 256: Note to reviewers: if none of the fields are provided, we assume the creator of the credReq is intending to use the old manual mode where the secrets are created by the ccoctl tool and CCO doesn't do anything (as is the case with openshift operators today).

pkg/azure/actuator.go
```diff
@@ -230,6 +236,52 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
...
+			return nil
+		}
+		err = validateAzureProviderSpec(*azureProviderSpec)
```
gallettilance, line 259: Note to reviewers: if some but not all fields required for this new manual mode workload identity workflow are set then we should error

pkg/azure/actuator.go
```diff
@@ -391,16 +460,24 @@ func (a *Actuator) syncCredentialSecrets(ctx context.Context, cr *minterv1.Crede
 			secret.Annotations = map[string]string{}
 		}
 		secret.Annotations[minterv1.AnnotationCredentialsRequest] = fmt.Sprintf("%s/%s", cr.Namespace, cr.Name)
+		if desiredSecret.Data == nil {
```
gallettilance, line 463: Note to reviewers: this if statement checks that the desiredSecret is from the new manual mode workload identity workflow since it does not set the Data field

pkg/azure/actuator_test.go
```diff
@@ -463,9 +509,31 @@ func TestActuator(t *testing.T) {
 		},
 		{
 			name:        "Mint annotation",
-			expectedErr: fmt.Errorf("error determining whether a credentials update is needed"),
+			expectedErr: fmt.Errorf("unexpected value or missing cloudcredential.openshift.io/mode annotation on admin credentials Secret"),
```
gallettilance, line 512: Note to reviewers: this change is a result of erroring earlier in the process as a result of checking for the cloudcredentials mode.

pkg/azure/const.go
```diff
@@ -1,11 +1,12 @@
...
-	AzureSubscriptionID = "azure_subscription_id"
-	AzureTenantID       = "azure_tenant_id"
+	AzureClientID           = "azure_client_id"
+	AzureClientSecret       = "azure_client_secret"
+	AzureRegion             = "azure_region"
+	AzureResourceGroup      = "azure_resourcegroup"
+	AzureResourcePrefix     = "azure_resource_prefix"
+	AzureSubscriptionID     = "azure_subscription_id"
+	AzureTenantID           = "azure_tenant_id"
+	AzureFederatedTokenFile = "azure_federated_token_file"
```
gallettilance, line 11: Note to reviewers: just adding the `AzureFederatedTokenFile`

gallettilance (commented):
Some notes for reviewers

pkg/azure/actuator.go
```diff
@@ -108,6 +95,19 @@ func (a *Actuator) needsUpdate(ctx context.Context, cr *minterv1.CredentialsRequ
...
+	credentialsMode, _, err := utils.GetOperatorConfiguration(a.client, logger)
+	if err != nil {
+		logger.WithError(err).Error("error loading CCO configuration to determine valid mode")
+		return true, err
+	}
+	if credentialsMode == operatorv1.CloudCredentialsModeManual {
+		return true, nil
+	}
+	if credentialsMode == operatorv1.CloudCredentialsModeMint {
+		return true, errors.New("mint mode is invalid")
```
jstuever, line 109: Should this return `false`?

pkg/azure/actuator.go
jstuever, line 106: Ideally, we would actually check if the secret differs from the source of truth similar to AWS mint mode. However, this can be taken away as tech debt.

jstuever (changes requested):

jstuever (approved):
/lgtm
/approve

PR #590, opened 2023-08-14
pkg/aws/actuator: Drop comment which suggested passthrough permission verification

wking:
17bd3eecfa237 (#228) moved the `CheckPermissionsUsingQueryClient` down to a `!=Passthrough` block, so drop the comment which suggested passthrough creds were checked to confirm sufficient permissions.


abutcher:
Overriding labels for a comment correction.

/label px-approved
/label qe-approved
/label docs-approved

PR #592, opened 2023-08-15
CCO-421: *: split pod identity webhooks from core controller

stevekuznetsov:
Managing these objects is a disjoint operation from the normal CCO mode, and requires different permissions. Splitting them lets us deploy these as needed.


stevekuznetsov:
@abutcher can you help me test this? Which optional tests should we trigger?

abutcher:
@stevekuznetsov I'll launch the sts + azwi ones but I doubt we have any CI testing of the webhooks -- these should result in a healthy operator state for CCO.

jstuever (approved):
/lgtm
/approve

PR #594, opened 2023-08-28
OCPVE-647: annotate manifests with capability name

qJkee:
add CloudCredential capability name to all manifests to make component optional

Also, i had to modify the `render` function. Now, if capability is disabled, we do render nothing


pkg/cmd/render/render_test.go
```diff
@@ -1,8 +1,9 @@
 package render
 
 import (
+	"bytes"
 	"fmt"
-	"io/ioutil"
+	v1 "github.com/openshift/api/config/v1"
```
fxierh, line 6: Ditto

go.mod
```diff
@@ -163,3 +163,5 @@ require (
 	sigs.k8s.io/kube-storage-version-migrator v0.0.4 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
+
+replace github.com/openshift/api => github.com/qjkee/api v0.0.0-20231110150827-4e8a5ac63e64
```
jstuever, line 167: Needs to be removed.

jstuever (changes requested):

jstuever:
Depends on https://github.com/openshift/api/pull/1571

manifests/0000_03_cloud-credential-operator_00_namespace.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: Namespace
```
wking, line 2: I don't understand what's happening with 9491d6070b4756a90's shift to runlevel 3.  Can you explain that, ideally in the commit message where future developers will be able to find it?

pkg/cmd/render/render.go
```diff
@@ -212,9 +221,19 @@ func render() error {
 	return nil
 }
 
+func isDisabledViaCapability(capabilities *v1.ClusterVersionCapabilitiesSpec) bool {
+	if capabilities == nil {
+		return false
+	}
+
+	enabledCaps := sets.New[v1.ClusterVersionCapability](v1.ClusterVersionCapabilitySets[capabilities.BaselineCapabilitySet]...)
```
wking, line 229: [Installer API][1]:

```go
	// baselineCapabilitySet selects an initial set of
	// optional capabilities to enable, which can be extended via
	// additionalEnabledCapabilities. The default is vCurrent.
	// +optional
	BaselineCapabilitySet configv1.ClusterVersionCapabilitySet `json:"baselineCapabilitySet,omitempty"`
```

I dunno how often folks take advantage of that optionality; at least so far all available caps have been part of `vCurrent`, so it seems unlikely the default is meaningful for anyone.  But you may want to either hard-code `vCurrent` as the default here, or get a knob in openshift/api about what the current default set is, or something, just in case a user leaves `baselineCapabilitySet` unset in their install-config.

[1]: https://github.com/openshift/installer/blob/8dc309c31b1cdc97b383f2514c7ed9ee1828a59c/pkg/types/installconfig.go#L527-L529

fxierh:
Why do we annotate the manifests under `./bindata/bootstrap` ? 
AFAIK those are to be rendered and applied by cluster-bootstrap which does not respect CVO annotations like `"[capability.openshift.io/name](http://capability.openshift.io/name): CloudCredential"`.

fxierh:
> I'd like to keep this manifests annotated just in case if someone would use them somewhere else

This makes me wonder why we maintain two sets of manifests, i.e. the `./bindata/bootstrap` ones, and the `./manifests` ones.

PR #607, opened 2023-10-17
OCPBUGS-21745: azure create-managed-identites to add cloud controller manager to network resource group

jstuever:
The Azure cloud controller manager needs access to modify the load balancers, which may exist in a network resource group. This change adds that access to the network resource group when using azure create-manage-identities.


sdodson:
/hold cancel
Lets go ahead and get this into a 4.15 nightly, if QE doesn't like it there they can put the brakes on the 4.14 backport.

PR #610, opened 2023-10-25
OCPBUGS-22369: explicitly set azure oidc bucket to allow public blob access

jstuever:
The cloud provider is changing the default behavior of the buckets to not be publicly available. This change explicitly sets the bucket created by 'ccoctl azure create' to allow public blob access.


huangmingxia:
Works well now, verified with 4.14.0-0.ci.test-2023-10-26-012238-ci-ln-5s021z2-latest.

PR #611, opened 2023-10-25
CCO-430: Use per-project custom roles instead of per-cluster custom roles

dlom:
Because CCO now manages custom roles per-project instead of per-cluster, the name format is fixed instead of random so that it all clusters within the same project will use the same custom roles.

Additionally, by default, custom roles will not be deleted.  To delete per-project custom roles, pass the `--force-delete-custom-roles` flag to `ccoctl`.

xref: https://issues.redhat.com/browse/CCO-430

/assign @jstuever 
/cc @abutcher

### DIFF
--- a/pkg/cmd/provisioning/gcp/create_all.go
+++ b/pkg/cmd/provisioning/gcp/create_all.go
@@ -60,7 +60,7 @@ func createAllCmd(cmd *cobra.Command, args []string) {
 // validationForCreateAllCmd will validate the arguments to the command, ensure the destination directory
 // is ready to receive the generated files, and will create the directory if necessary.
 func validationForCreateAllCmd(cmd *cobra.Command, args []string) {
-	if len(CreateWorkloadIdentityPoolOpts.Name) > 32 {
+	if len(CreateAllOpts.Name) > 32 {
 		log.Fatalf("Name can be at most 32 characters long")
 	}
 
--- a/pkg/cmd/provisioning/gcp/create_service_accounts.go
+++ b/pkg/cmd/provisioning/gcp/create_service_accounts.go
@@ -128,14 +128,17 @@ func createServiceAccount(ctx context.Context, client gcp.Client, name string, c
 	}
 
 	// The role ID has a max length of 64 chars and can include only letters, numbers, period and underscores
-	// we sanitize infraName and crName to make them alphanumeric and then
-	// split role ID into 29_28_5 where the resulting string becomes:
-	// <infraName chopped to 29 chars>_<crName chopped to 28 chars>_<random 5 chars>
-	roleID, err := actuator.GenerateRoleID(name, credReq.Name)
+	// we sanitize projectName and crName to make them alphanumeric and then
+	// split role ID into 32_31 where the resulting string becomes:
+	// <projectName chopped to 32 chars>_<crName chopped to 31 chars>
+	roleID, err := actuator.GenerateRoleID(client.GetProjectName(), credReq.Name)
+	if err != nil {
+		return "", fmt.Errorf("error generating custom role id: %v", err)
+	}
 
 	// The role name field has a 100 char max, so generate a name consisting of the
-	// infraName chopped to 50 chars + the crName chopped to 49 chars (separated by a '-').
-	roleName, err := utils.GenerateNameWithFieldLimits(name, 50, credReq.Name, 49)
+	// projectName chopped to 50 chars + the crName chopped to 49 chars (separated by a '-').
+	roleName, err := actuator.GenerateRoleName(client.GetProjectName(), credReq.Name)
 	if err != nil {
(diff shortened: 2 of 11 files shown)


dlom:
/override ci/prow/security 

lets see if I did that right...

pkg/cmd/provisioning/gcp/create_all.go
```diff
@@ -60,7 +60,7 @@ func createAllCmd(cmd *cobra.Command, args []string) {
 // validationForCreateAllCmd will validate the arguments to the command, ensure the destination directory
 // is ready to receive the generated files, and will create the directory if necessary.
 func validationForCreateAllCmd(cmd *cobra.Command, args []string) {
-	if len(CreateWorkloadIdentityPoolOpts.Name) > 32 {
+	if len(CreateAllOpts.Name) > 32 {
```
dlom, line 63: this is unrelated to the topic of the PR, but the wrong variable was used here (and in a few other places)

pkg/gcp/actuator/utils_test.go
```diff
@@ -0,0 +1,39 @@
...
+			expectedAdded:   []string{"d"},
+			expectedRemoved: []string{"b"},
+		},
+		{
+			name:            "extreme",
+			original:        []string{"a", "b", "c", "d", "e", "f"},
+			new:             []string{"a", "c", "d", "g", "h", "q", "z"},
+			expectedAdded:   []string{"q", "z", "h", "g"},
+			expectedRemoved: []string{"f", "e", "b"},
+		},
```
jstuever, line 29: I think it would be more concise to test the following:

- There are no differences
- There are expectedAdded
- There are expectedRemoved
- There are expectedAdded and expectedRemoved

and also situations where
- original is empty
- new is empty

pkg/cmd/provisioning/gcp/create_service_accounts.go
```diff
@@ -128,14 +128,14 @@ func createServiceAccount(ctx context.Context, client gcp.Client, name string, c
...
 
 	// The role ID has a max length of 64 chars and can include only letters, numbers, period and underscores
-	// we sanitize infraName and crName to make them alphanumeric and then
-	// split role ID into 29_28_5 where the resulting string becomes:
-	// <infraName chopped to 29 chars>_<crName chopped to 28 chars>_<random 5 chars>
-	roleID, err := actuator.GenerateRoleID(name, credReq.Name)
+	// we sanitize projectName and crName to make them alphanumeric and then
+	// split role ID into 32_31 where the resulting string becomes:
+	// <projectName chopped to 32 chars>_<crName chopped to 31 chars>
+	roleID, err := actuator.GenerateRoleID(client.GetProjectName(), credReq.Name)
```
jstuever, line 134: We should probably add handling the err here.

pkg/cmd/provisioning/gcp/create_service_accounts.go
```diff
@@ -264,8 +262,15 @@ func createServiceAccount(ctx context.Context, client gcp.Client, name string, c
 				}
 			} else {
 				log.Printf("Existing IAM custom role %s found, updating permissions", role.Title)
-				if !actuator.AreSlicesEqualWithoutOrder(role.IncludedPermissions, gcpProviderSpec.Permissions) {
-					role.IncludedPermissions = gcpProviderSpec.Permissions
+				addedPermissions, removedPermissions := actuator.CalculateSliceDiff(role.IncludedPermissions, gcpProviderSpec.Permissions)
+
+				if len(removedPermissions) > 0 {
+					allRemovedPermissions := strings.Join(removedPermissions, ", ")
+					log.Printf("The following permissions for IAM custom role %s will NOT be removed: [%s] (permissions deltas can only be additive)", role.Title, allRemovedPermissions)
```
jstuever, line 269: Reword the log... something along the lines of "Unexpected permissions found on existing custom role..." etc.

pkg/gcp/actuator/actuator.go
```diff
@@ -426,9 +420,16 @@ func (a *Actuator) syncMint(ctx context.Context, cr *minterv1.CredentialsRequest
 			}
 			roles = append(roles, role.Name)
 		} else {
-			if !AreSlicesEqualWithoutOrder(role.IncludedPermissions, gcpSpec.Permissions) {
+			addedPermissions, removedPermissions := CalculateSliceDiff(role.IncludedPermissions, gcpSpec.Permissions)
+
+			if len(removedPermissions) > 0 {
```
jstuever, line 425: Just a note, I believe this will only be called if also there are added permissions, because otherwise needsUpdate will return false due to only checking for added permissions there.

jstuever (changes requested):

dlom:
/override ci/prow/security

Security check is new and requires a seperate PR to get master in a clean state

PR #613, opened 2023-10-30
snyk: exclude vendor/

2uasimojo:
...as we expect vulns in deps to be caught by the dep scan.


2uasimojo:
/assign @dlom 

This is working. `security` fail is expected at this point:
- grpc dep update still needed.
- low/med exclusions coming via release repo PR.

PR #634, opened 2023-11-30
NO-JIRA: Update docs/sts.md

shirady:
1. Add a comment about the name flag in step 3 of in-place migration.


shirady:
HI,
I would also note that step 6 is not clear "Wait for the kube-apiserver pods to be updated with the new config" - what exactly are we waiting for? (what is the state before and what is the state after?)

PR #652, opened 2024-01-11
azwi doc: apply webhook manifest to cluster

jstuever:
Update the azure workload identity document to replace the techpreview step with the application of the azure webhook config manifest.


jstuever:
/skip
docs only

PR #656, opened 2024-01-21
OCPBUGS-27445: Use cached clients to avoid client side throttling

fxierh:
Client-side throttling was observed after switching to the live client in the metrics controller in #645. 

This PR:
- Adds a (cached) root client to the Calculator struct for querying the root credential only. 
- Switches back to the default (cached) client (i.e. `mgr.GetClient()`) for querying anything else.


fxierh:
/hold

Work in progress.

fxierh:
/assign @jstuever @dlom 

jstuever (approved):
/lgtm
/approve

PR #665, opened 2024-01-26
OCPBUGS-28388: Fix the ClusterOperator watch of the status controller

fxierh:
Previously the status controller erroneously:
- watches all ClusterOperators. 
- enqueues ClusterOperator/XXX instead of CloudCredential/cluster when informed by ClusterOperator-related events

This PR restricts the watch to ClusterOperator/cloud-credential and make sure the EventHandler always enqueues the same namespaced name (i.e. CloudCredential/cluster). 

See the comment section of the attached JIRA card for a comparison of metrics pre vs post bug fix.

### DIFF
--- a/pkg/operator/status/status_controller.go
+++ b/pkg/operator/status/status_controller.go
@@ -112,13 +112,13 @@ func Add(mgr, rootCredentialManager manager.Manager, kubeConfig string) error {
 	// always reconcile status when the clusteroperator/cloud-credential changes.
 	if err := c.Watch(source.Kind(operatorCache, &configv1.ClusterOperator{}), handler.EnqueueRequestsFromMapFunc(alwaysReconcileCCOConfigObject), predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
-			return e.Object.GetName() == constants.CloudCredClusterOperatorName
+			return e.Object != nil && e.Object.GetName() == constants.CloudCredClusterOperatorName
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			return e.ObjectNew.GetName() == constants.CloudCredClusterOperatorName
+			return e.ObjectNew != nil && e.ObjectNew.GetName() == constants.CloudCredClusterOperatorName
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
-			return e.Object.GetName() == constants.CloudCredClusterOperatorName
+			return e.Object != nil && e.Object.GetName() == constants.CloudCredClusterOperatorName
 		},
 	}); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


fxierh:
/hold

Work in progress.

fxierh:
/assign @jstuever @dlom 

Could you please take a look ? Thanks.

pkg/operator/status/status_controller.go
```diff
@@ -110,12 +110,15 @@ func Add(mgr, rootCredentialManager manager.Manager, kubeConfig string) error {
...
 	// always reconcile status when the clusteroperator/cloud-credential changes.
-	if err := c.Watch(source.Kind(operatorCache, &configv1.ClusterOperator{}), &handler.EnqueueRequestForObject{}, predicate.Funcs{
-		GenericFunc: func(genericEvent event.GenericEvent) bool {
-			if genericEvent.Object == nil {
-				return false
-			}
-			return genericEvent.Object.GetName() == "cloud-credential"
+	if err := c.Watch(source.Kind(operatorCache, &configv1.ClusterOperator{}), handler.EnqueueRequestsFromMapFunc(alwaysReconcileCCOConfigObject), predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return e.Object.GetName() == constants.CloudCredClusterOperatorName
```
dlom, line 115: I'm not super familiar with how these callbacks work, but is there any chance of a nil dereference here? Maybe add the nil check back to all 3 of the functions, just to be safe

dlom (changes requested):

pkg/operator/status/status_controller.go
fxierh: Thanks for pointing this out. After a some digging, it looks like `e.Object` or `e.ObjectNew` come from `watch.Event` objects which in turn are sourced from a client-go-style `.Watch().ResultChan()`. These objects are essentially part of the API server's response and should not be nil without the decoder erroring out. 

Nevertheless, it is definitely good to be cautious. nil checks are added in the callbacks for safety.

jstuever:
I'm looking further into this. I'm concerned that this may change the behavior such that it's not reconciling in all the situations where it maybe should be.

fxierh:
> I'm concerned that this may change the behavior such that it's not reconciling in all the situations where it maybe should be.

I believe it does what it is expected to do -- "always reconcile status when the clusteroperator/cloud-credential changes" (says the unchanged comment [here](https://github.com/openshift/cloud-credential-operator/blob/07a77966c7f158c22afeea4f4ad33d4f82636424/pkg/operator/status/status_controller.go#L112)). 

Here's why:
1) `source.Kind`, once started, gets (or adds if necessary) an informer for the gvk to watch (ClusterOperator in our case), then adds an event handler to the aforementioned informer as follows:
https://github.com/openshift/cloud-credential-operator/blob/07a77966c7f158c22afeea4f4ad33d4f82636424/vendor/sigs.k8s.io/controller-runtime/pkg/internal/source/kind.go#L82
The event handler dispatches (and only dispatches) create, update and delete events, see this func:
https://github.com/openshift/cloud-credential-operator/blob/07a77966c7f158c22afeea4f4ad33d4f82636424/vendor/k8s.io/client-go/tools/cache/shared_informer.go#L960
Callbacks are provided to all 3 types of events. See CreateFunc, UpdateFunc or DeleteFunc in the changed code.
2) The same pattern is also found elsewhere in this repo, e.g.  
https://github.com/openshift/cloud-credential-operator/blob/07a77966c7f158c22afeea4f4ad33d4f82636424/pkg/operator/status/status_controller.go#L138-L157

jstuever (approved):
/lgtm
/approve

PR #667, opened 2024-01-29
OCPBUGS-28535: stop LoadInfrastructureRegion from panicking when being called on a non-AWS cluster

fxierh:
This PR:
- Fixes `LoadInfrastructureRegion` so it won't panic when being called on a non-AWS cluster (this is possible, see the attached card for details)
- Increase unit test coverage for the `secretannotator` controller
- Remove (another) `LoadInfrastructureRegion` which is no longer used


fxierh:
/assign @jstuever @dlom 

Could you please take a look ? Thanks.

PR #669, opened 2024-01-31
NO-JIRA: Fix update-codegen.sh

fxierh:
The following issues are found on update-codegen.sh:
- On MacOS, `sed` interprets the string following `-i` as the `extension` passed to this flag, resulting in the following:
```shell
$ ./hack/update-codegen.sh
...
+ sed -i 's,^exec \(".*/generate-internal-groups.sh"\),bash \1,g' ./vendor/k8s.io/code-generator/generate-groups.sh
sed: 1: "./vendor/k8s.io/code-ge ...": invalid command code .
```
- Two traps could be set on the same signal (EXIT)

This PR:
- Use gsed (aka gnu-sed) for MacOS, sed otherwise. Also see: https://github.com/openshift/hive/pull/2183. 
- Aggregate all cleanup logics into the `cleanup` function so we end up with one single trap

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -17,6 +17,17 @@ ensure-temp-gopath() {
 	ln -s "$REPO_FULL_PATH" "${fake_repopath}"
 }
 
+cleanup() {
+    # Restore modified files, or `verify` will puke.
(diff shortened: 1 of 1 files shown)


fxierh:
Actually there seems to be another option: 
`sed -i 's/bar/foo/g' text.txt -> sed -i'' 's/bar/foo/g' text.txt `

fxierh:
/assign @2uasimojo 

PR #671, opened 2024-02-05
OCPBUGS-29027: dynamically obtain permissions required for passthrough

fxierh:
On AWS, when CCO is in the default mode and the root credential is not sufficient for mint mode, the secret annotator controller checks if the root credential is good enough for passthrough against a static list of permissions
https://github.com/openshift/cloud-credential-operator/blob/3cf3099b60fbd886dd9582c55373f36419c96fff/pkg/aws/utils.go#L35-L93
This list of perms is outdated.

### DIFF
--- a/pkg/aws/utils.go
+++ b/pkg/aws/utils.go
@@ -232,7 +232,6 @@ func CheckPermissionsAgainstActions(awsClient Client, actionList []string, param
 // CheckCloudCredPassthrough will see if the provided creds are good enough to pass through
 // to other components as-is based on the static list of permissions needed by the various
 // users of CredentialsRequests
-// TODO: move away from static list (to dynamic passthrough validation?)
 func CheckCloudCredPassthrough(awsClient Client, params *SimulateParams, logger log.FieldLogger) (bool, error) {
 	return CheckPermissionsAgainstActions(awsClient, credPassthroughActions, params, logger)
 }


fxierh:
I personally prefer the 2nd option above.

fxierh:
It looks like the installer is the only user of `func CheckCloudCredPassthrough`: 
https://github.com/openshift/installer/blob/3d78a2270aad5f71034bed9b7ab0ae992c6c83cd/pkg/asset/installconfig/aws/permissions.go#L310

Considering the above:
- Option 1 has no impact on the installer. 
- Option 2 would require the installer: 
   - to be refactored in a way that extracts all CredentialsRequests (for the platform to install the cluster) from the release image and glean the permissions from them OR
   - to remove the entire passthrough-permission-checking logic above
- Option 3 would require the installer to remove the passthrough-permission-checking logic above. 

I guess option 1 is the way to go. 

@jstuever @dlom WDTY ?

jstuever:
Sorry for the delay, I had to look into this one. I think we are stuck with option 1 for now. The credentialRequests can be obtained for a specific version with the following:
```
oc adm release extract --credentials-requests --cloud=aws to=./creds  --from <release>
```

Note: This will become complicated as these operators become optional components, so the installer might need to rework this in the future.

fxierh:
@jstuever 

Sorry for the late reply. 

> Note: This will become complicated as these operators become optional components, so the installer might need to rework this in the future.

Thanks for the reminder. This seems to be a common issue for both the CCO and the installer. Currently, among the 7 operators which request for cloud credentials on AWS, 4 are already optional: CCO, MAPI, ImageRegistry and Storage. 

With the above taken into account, and if we go with option 1, we'd need to make the static list of permissions minimal, meaning that it will only contain permissions required by the non-optional operators. This makes CCO's secret annotator controller and installer's platform permission check kind of in-precise, even if we keep the static list of permissions up-to-date. 

I'm thus wondering if it's good timing to solve this issue once and for all by
- either passing an abstraction (e.g. interface) to `CheckCloudCredPassthrough` so the CCO and the installer could implement their own logic to obtain the list of permissions;
- or letting both components obtain the list of permissions themselves somehow and then call `CheckPermissionsAgainstActions` directly.

PR #677, opened 2024-02-07
NO-JIRA:docu change: fix find command

Unfaced85:
After testing the Migration part out I found out that the find command does not find the secrets, instead it finds the CredentialsRequest Objects.

This change in the Documentation will reflect how it should work.


docs/sts.md
```diff
@@ -176,7 +176,7 @@ $ ./ccoctl aws create-iam-roles --output-dir ./newstscluster/ --name newstsclust
 
 13. Apply the Secrets generated by the above command:
 ```bash
-$ find ./cred-reqs -iname "*yaml" -print0 | xargs -I {} -0 -t oc replace -f {}
+$ find ./newstscluster -iname "*yaml" -print0 | xargs -I {} -0 -t oc replace -f {}
```
jstuever, line 179: This can be more precise by using `find ./newstscluster/manifests`.

jstuever (changes requested):
Thank you for submitting this pull request. One minor adjustment and then we should be good to go.

docs/sts.md
Unfaced85: updated as requested

jstuever (approved):
/lgtm
/approve

PR #685, opened 2024-03-06
OCPBUGS-28787: gcp actuator update check to consider custom roles

jstuever:
Prior to this change, when the root credential was removed while the cluster was in mint mode, the cloud credentail operator would become degraded after aproximately an hour. This is because when checking if the service account needed to be updated, it mistakenly considered the custom role to be an excessive role that needed to be removed. Then, it would fail to perform the updates due to missing the root credential.

This change ensures the custom role is considred when checking if the service accounts need to be updated. As a result, the update is no longer happening in this scenario, thus avoiding the need for a root credential.

### DIFF
--- a/pkg/gcp/actuator/policy.go
+++ b/pkg/gcp/actuator/policy.go
@@ -19,6 +19,7 @@ package actuator
 import (
 	"fmt"
 
+	log "github.com/sirupsen/logrus"
 	"google.golang.org/api/cloudresourcemanager/v1"
 	"google.golang.org/api/iam/v1"
 	iamadminpb "google.golang.org/genproto/googleapis/iam/admin/v1"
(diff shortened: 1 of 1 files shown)


jianping-shu:
Verified with cluster-bot build.
1. GCP Mint cluster installed successfully
jianpingshu@jshu-mac ~ % oc get clusterversion
NAME      VERSION                                                   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.15.0-0.ci.test-2024-03-06-021033-ci-ln-7ql7iwb-latest   True        False         53m     Cluster version is 4.15.0-0.ci.test-2024-03-06-021033-ci-ln-7ql7iwb-latest
jianpingshu@jshu-mac ~ % oc get secret -n kube-system gcp-credentials -o=jsonpath={.metadata.annotations}
{"cloudcredential.openshift.io/mode":"mint"}
2. Delete the root credential
jianpingshu@jshu-mac ~ % oc delete secret -n kube-system gcp-credentials
secret "gcp-credentials" deleted
4. Wait > 3h, no degrade
jianpingshu@jshu-mac ~ % oc get co cloud-credential -o=jsonpath={.status.conditions}
[{"lastTransitionTime":"2024-03-06T03:37:53Z","message":"All is well","reason":"AsExpected","status":"True","type":"Available"},{"lastTransitionTime":"2024-03-06T03:39:11Z","message":"All is well","reason":"AsExpected","status":"False","type":"Degraded"},{"lastTransitionTime":"2024-03-06T03:51:10Z","message":"All is well","reason":"AsExpected","status":"False","type":"Progressing"},{"lastTransitionTime":"2024-03-06T05:01:10Z","message":"Parent credentials secret must be restored prior to upgrade: kube-system/gcp-credentials","reason":"MissingRootCredential","status":"False","type":"Upgradeable"}]

PR #695, opened 2024-04-23
OCPBUGS-32026: Docs: Fix migration to azure workload identy to apply the webhook config

jstuever:
This documentation update changes the migration to azure workload identity workflow to apply the webhook configuration instead of replacing it. In a pristine cluster that was installed without azure workload identity, this configuration doesn't exist.


huangmingxia:
Hi @jstuever  `oc apply `  works, but there is another issue, please see comment. thanks.

huangmingxia:
@jstuever It works well now.

Verified after migration:
The issues in Description have been fixed. Secrets contain an azure_federated_token_file key. It is expected.

PR #708, opened 2024-06-12
CCO-572: STS OLM for GCP (WIF)

dlom:
/hold
/assign @jstuever 
xref: [CCO-572](https://issues.redhat.com//browse/CCO-572)

### DIFF
--- a/pkg/gcp/actuator/actuator.go
+++ b/pkg/gcp/actuator/actuator.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
@@ -54,6 +55,20 @@ const (
 	roGCPCredsSecret          = "cloud-credential-operator-gcp-ro-creds"
 
 	gcpSecretJSONKey = "service_account.json"
+
+	gcpSTSCredsTemplate = `{
+  "type": "external_account",
+  "audience": "%s",
+  "subject_token_type": "urn:ietf:params:oauth:token-type:jwt",
+  "token_url": "https://sts.googleapis.com/v1/token",
+  "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/%s:generateAccessToken",
+  "credential_source": {
+    "file": "%s",
+    "format": {
+      "type": "text"
+    }
+  }
+}`
 )
 
 var _ actuatoriface.Actuator = (*Actuator)(nil)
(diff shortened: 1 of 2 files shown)


pkg/gcp/actuator/actuator.go
```diff
@@ -861,6 +954,31 @@ func checkServicesEnabled(gcpClient ccgcp.Client, permList []string, logger log.
 	return serviceAPIsEnabled, nil
 }
 
+func validateWIFProviderSpec(providerSpec minterv1.GCPProviderSpec) error {
+	var errors []error
+	isEmptyProjectNumber := providerSpec.ProjectNumber == ""
+	isEmptyPoolID := providerSpec.PoolID == ""
+	isEmptyProviderID := providerSpec.ProviderID == ""
+	isEmptyServiceAccountEmail := providerSpec.ServiceAccountEmail == ""
```
jstuever, line 962: nit: I'm not a huge fan of splitting this out into variables... this can be done inline below while reducing lines of code without really affecting readability.
```
if providerSpec.ProjectNumber == "" {
```

jstuever (changes requested):

pkg/gcp/actuator/actuator.go
```diff
@@ -238,6 +287,50 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
 	return nil
 }
 
+// syncWIFSecret makes a time-based token available in a Secret in the namespace of an operator that
+// has supplied the following in the CredentialsRequest:
+// a non-nil spec.CloudTokenString
+// a path to the JWT token: spec.cloudTokenPath
```
RomanBednar, line 293: `cloudTokenPath` is not a hard requirement since we have a default value, correct?

pkg/gcp/actuator/actuator.go
```diff
@@ -238,6 +287,50 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
...
+// a spec.SecretRef.Name
+// a cr.Spec.SecretRef.Namespace
```
RomanBednar, line 295:
```suggestion
// a spec.SecretRef.Namespace
```

pkg/gcp/actuator/actuator.go
```diff
@@ -238,6 +287,50 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
 	return nil
 }
 
+// syncWIFSecret makes a time-based token available in a Secret in the namespace of an operator that
+// has supplied the following in the CredentialsRequest:
+// a non-nil spec.CloudTokenString
```
RomanBednar, line 292: I don't see `CloudTokenString` in the API - remove this comment line?

pkg/gcp/actuator/actuator.go
```diff
@@ -238,6 +286,59 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
...
 }
 
+// syncSTSSecret makes a time-based token available in a Secret in the namespace of an operator that
+// has supplied the following in the CredentialsRequest:
+// a non-empty cr.CloudTokenPath
+// a non-empty cr.ProviderSpec.ServiceAccountEmail
+// a non-empty cr.ProviderSpec.PoolID
+// a non-empty cr.ProviderSpec.ProviderID
+func (a *Actuator) syncSTSSecret(poolName, providerId, serviceAccount, cloudTokenPath string, cr *minterv1.CredentialsRequest, logger log.FieldLogger, ctx context.Context) error {
+	gcpClient, err := a.buildRootGCPClient(cr)
```
jstuever, line 296: I don't think we can assume STS clusters have the Root client... they probably have `buildReadGCPClient`, though.

jstuever: Actually, can this be derived from the PoolID or ProviderID? That would be more accurate to what it is being used for.

jstuever: Even so, you would still need to `buildReadGCPClient`... and that client probably doesn't have the permissions to check the PoolID or ProviderID. It might be best to revert to asking for the ProjectID like you originally did.

jstuever: Alternatively, you could ask for the audience directly instead of computing it based on multiple inputs.

jstuever (changes requested):

pkg/gcp/actuator/actuator.go
```diff
@@ -195,6 +210,40 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
...
+			msg := "error validating credentials request GCP Workload Identity Federation fields"
+			return &actuatoriface.ActuatorError{
+				ErrReason: minterv1.CredentialsProvisionFailure,
+				Message:   fmt.Sprintf("%v: %v", msg, err),
+			}
+		}
+		cloudTokenPath := cr.Spec.CloudTokenPath
+		if cr.Spec.CloudTokenPath == "" {
+			logger.Debug("CredentialsRequest has no cloudTokenPath, defaulting cloudTokenPath to /var/run/secrets/kubernetes.io/serviceaccount/token")
+			cloudTokenPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
RomanBednar, line 241: Should this default to `"/var/run/secrets/openshift/serviceaccount/token"` instead? 

https://github.com/openshift/cloud-credential-operator/blob/a7f311bc118fb79d78501939d3fe392020dae1ed/pkg/apis/cloudcredential/v1/types_credentialsrequest.go#L76

https://github.com/openshift/cloud-credential-operator/blob/a7f311bc118fb79d78501939d3fe392020dae1ed/docs/gcp_workload_identity.md?plain=1#L115

jstuever: /lgtm cancel
Oh! Good catch!
https://github.com/openshift/cloud-credential-operator/blob/a68b344ed8ee1f1d27721760345781b06db2da13/pkg/cmd/provisioning/constants.go#L17

dlom: It's the same in the AWS actuator, but I'll leave that for a follow up PR

jstuever: /lgtm

jstuever: I think there is merit to having the non-core OpenShift operators default to a more Kubernetes native path. I would be tempted to keep it as is and updater the Azure to match.

PR #711, opened 2024-06-28
Fix Makefile for developer install

### LINKED ISSUE #710: make install failed
When we run `make install` as given here in https://github.com/openshift/cloud-credential-operator?tab=readme-ov-file#developer-instructions , it fails with the following error:
```
error: the path "manifests/00-crd.yaml" does not exist
make: *** [Makefile:130: install] Error 1
```

khrm:
Fixes #710 so that we can run `make install`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ run: build
 
 # Install CRDs into a cluster
 install: update-codegen
-	kubectl apply -f manifests/00-crd.yaml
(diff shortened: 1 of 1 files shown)


khrm (commented):
/test security

jstuever (approved):
/lgtm
/approve

jstuever:
/retitle NO-JIRA: Fix Makefile for developer install

PR #725, opened 2024-07-15
CCO-580: Upgrade to Kubernetes 1.30

jstuever:
This change upgrades the dependencies to Kubernetes 1.30. This requires a controller-runtime update to be compatible with the newer client-go.


jstuever:
Depends on https://github.com/openshift/cloud-credential-operator/pull/719

dlom:
@jstuever here's how I got code generation/verify working again in hive: https://github.com/dlom/hive/commit/b9e81e1bd3c77f89e7b875db76f1c7b6e6bfdb96

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -284,41 +284,82 @@ func add(mgr, adminMgr manager.Manager, r reconcile.Reconciler) error {
...
-		allCredRequestsMapFn,
-		configMapPredicate)
+		source.Kind(operatorCache, &corev1.ConfigMap{},
+			configMapAllCredRequestsMapFn,
+			configMapPredicate))
 	if err != nil {
 		return err
 	}
 
+	// configMapAllCredRequestsMapFn simply looks up all CredentialsRequests and requests they be reconciled.
```
dlom, line 338: s/config/credential

pkg/operator/credentialsrequest/credentialsrequest_controller.go
```diff
@@ -284,41 +284,82 @@ func add(mgr, adminMgr manager.Manager, r reconcile.Reconciler) error {
...
+	credentialMapAllCredRequestsMapFn := handler.TypedEnqueueRequestsFromMapFunc(func(ctx context.Context, a *operatorv1.CloudCredential) []reconcile.Request {
```
dlom, line 339: I think this function could be made generic and combined with the similarly named function on line 295.  Neither function actually references the `a` parameter anyways

jstuever: They don't reference the parameter, but the parameter is of different type.

jstuever: I wanted to break this out into its own function, but they use `mgr` inline, which can't really be passed any other way.

pkg/operator/credentialsrequest/credentialsrequest_controller.go
jstuever: Good catch! [Done](https://github.com/openshift/cloud-credential-operator/compare/a59a050e97e61d6ec59ae4eaa75ea581dbb93de5..e8ed3b77640f58d32912a590527bc7b105962c08).

PR #752, opened 2024-09-04
OCPBUGS-39531: List secrets in batches to avoid api timeout

jstuever:
Prior to this change, the operator listed all of the secrets in every namespace using a single list call. The results were then used to determine if the operator can limit the secrets it is watching by label. Unfortunately, when there is a huge number of secrets, the kube API may timeout prior to returning the list.

This change modifies the behavior to pull the list in smaller batches to avoid reaching the timeout.


jianping-shu:
Regression test with AWS and bare metal cluster install, both were successful.
Will try to reproduce the issue with large number of secrets.

Issue #761, opened 2024-09-24
AWS with --cli-input-json shows invalid 


owenmather:
The extra comma in the generated `--dry-run` `07-block-public-access-to-oidc-bucket.json` file prevents aws cli from processing using `--cli-input-json`

```
ccoctl aws create-identity-provider --name testcluster --public-key-file /home/ec2-user/sts/serviceaccount-signer.public --region eu-west-1 --dry-run --create-private-s3-bucket

aws s3api put-public-access-block --cli-input-json file://./07-block-public-access-to-oidc-bucket.json

Error parsing parameter 'cli-input-json': Invalid JSON received.

cat 07-block-public-access-to-oidc-bucket.json
{
        "Bucket": "testcluster-oidc",
        "PublicAccessBlockConfiguration": {
        "BlockPublicAcls": true,
        "IgnorePublicAcls": true,
        "BlockPublicPolicy": true,
        "RestrictPublicBuckets": true
    }, #<-- invalid JSON 
}

cat 07-block-public-access-to-oidc-bucket.json  | jq
jq: parse error: Expected another key-value pair at line 9, column 1

```

Works as expected with this comma removed.

https://github.com/openshift/cloud-credential-operator/blob/d668c182b0e3523f496936586afa82bcff917a6d/pkg/cmd/provisioning/aws/create_identity_provider.go#L59

PR #762, opened 2024-09-24
NO-JIRA: Fix JSON for 07-block-public-access-to-oidc-bucket.json

owenmather:
Proposed fix for https://github.com/openshift/cloud-credential-operator/issues/761


dlom:
/retitle NOJIRA: Fix JSON for 07-block-public-access-to-oidc-bucket.json

PR #771, opened 2024-10-21
OCPBUGS-36644: Only attempt timed token credentials on supported platforms.

jstuever:
Prior to this change, all platforms attempted to reconcile credentials
when in manual mode as if they were in mint mode. At one point, a per
platform featuregate happened to limit which ones were attempting to do
so. However, once the featuregate was removed, there was no further
logic to limit this.
   
This change adds a function to the actuator in order to enable each
platform to decide for itself if time token credentials are supported in
addition to determining if they are configured. This ensures each
platform only attempts to sync in manual mode if it already has the
logic to do so.

The following platforms have timed token credentials enabled:
AWS, Azure, GCP

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -1398,7 +1398,7 @@ func awsSTSIAMRoleARN(codec *minterv1.ProviderCodec, credentialsRequest *minterv
 }
 
 func (a *AWSActuator) IsTimedTokenCluster(c client.Client, ctx context.Context, logger log.FieldLogger) (bool, error) {
-	return false, nil
+	return utils.IsTimedTokenCluster(c, ctx, logger)
 }
 
 // Upgradeable returns a ClusterOperator status condition for the upgradeable type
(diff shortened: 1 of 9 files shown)


jianping-shu:
Checked on version 4.18.0-0.nightly-2024-10-21-132049 for all token based platforms(aws, azure, gcp), all of them don't have those logs on bug description.
Particularly for gcp, seems it was fixed by [OCPBUGS-38951](https://issues.redhat.com/browse/OCPBUGS-38951) (back ported to 4.17), so 4.16 and previous releases still have the issue.
Had a preliminary test with the cluster-bot build for gcp wif cluster.
Cluster installed successfully and healthy.
Regression tests OCP-34470/50869/75429 passed.

jstuever:
/jira-backport release-4.17
STS functionality for GCP was added in 4.17, so 4.16 and prior should not include the gcp commit.

jstuever:
/jira backport release-4.17
STS functionality for GCP was added in 4.17, so 4.16 and prior should not include the gcp commit.

jstuever:
/hold
for e2e-azure and e2e-azure-manual-oidc

PR #772, opened 2024-10-22
OCPBUGS-43645: Only attempt timed token credentials on supported platforms.

jstuever:
Prior to this change, all platforms attempted to reconcile credentials
when in manual mode as if they were in mint mode. At one point, a per
platform featuregate happened to limit which ones were attempting to do
so. However, once the featuregate was removed, there was no further
logic to limit this.

This change adds a function to the actuator in order to enable each
platform to decide for itself if time token credentials are supported in
addition to determining if they are configured. This ensures each
platform only attempts to sync in manual mode if it already has the
logic to do so.

The following platforms have timed token credentials enabled:
AWS, Azure

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -1374,6 +1374,10 @@ func awsSTSIAMRoleARN(codec *minterv1.ProviderCodec, credentialsRequest *minterv
 	return awsSpec.STSIAMRoleARN, nil
 }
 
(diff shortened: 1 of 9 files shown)


jstuever:
This PR is a cherry-pick of the original commits with the exception of the one enabling OLM STS in GCP, which didn't exist in 4.16 and prior.

jianping-shu:
Tested with cluster-bot build 4.16.0-0.ci.test-2024-10-24-021827-ci-ln-hdixpdt-latest.
The issue is fixed in gcp wif cluster and regression tested with aws sts/gcp wif/azure entra Id/vsphere clusters.
/label cherry-pick-approved

PR #777, opened 2024-10-25
OCPBUGS-43821: pkg/gcp/actuator: Guard API-enable check with skipServiceCheck

wking:
The `sync()` `not all required service APIs are enabled` check landed back in 9e1d7946c4 (#86).  There's a similar check in `syncMint`. The `syncMint` check grew a `skipServiceCheck` guard in 9d0bee6fd2 (#259).

This commit lifts the `skipServiceCheck` guard up into `sync()` as well, to avoid erroring out on `skipServiceCheck` CredentialRequests that request APIs that have not been enabled ([OCPBUGS-43821][1]).

[1]: https://issues.redhat.com/browse/OCPBUGS-43821

### DIFF
--- a/pkg/gcp/actuator/actuator.go
+++ b/pkg/gcp/actuator/actuator.go
@@ -181,7 +181,12 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
 		}
 	}
 
-	if !servicesAPIsEnabled {
+	gcpSpec, err := decodeProviderSpec(minterv1.Codec, cr)
+	if err != nil {
+		return true, false, fmt.Errorf("unable to decode ProviderSpec: %v", err)
+	}
+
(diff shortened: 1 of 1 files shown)


dlom:
/retitle OCPBUGS-43821: pkg/gcp/actuator: Guard API-enable check with skipServiceCheck

pkg/gcp/actuator/actuator.go
```diff
@@ -181,7 +181,12 @@ func (a *Actuator) sync(ctx context.Context, cr *minterv1.CredentialsRequest) er
 		}
 	}
 
-	if !servicesAPIsEnabled {
+	gcpSpec, err := decodeProviderSpec(minterv1.Codec, cr)
+	if err != nil {
+		return true, false, fmt.Errorf("unable to decode ProviderSpec: %v", err)
```
dlom, line 186: this function only returns `error`

PR #818, opened 2025-02-11
NO-JIRA: Add dry-run on aws create-all command

danilo-gemoli:
As per title, enable `--dry-run` when generating all AWS assets by running `ccoctl aws create-all`.
The change should be quite straightforward as the vast majority of the code is there, already implemented.
The sole relevant change is in `func getIssuerURLFromIdentityProvider`.


jstuever:
/override ci/prow/security
This is covered by existing bugs.

jstuever:
/override ci/prow/security
There are existing bugs for the reported issues.

PR #820, opened 2025-02-25
NO-JIRA: add example of how to build from source

pruan-rht:
for users who want to build from source.


jstuever:
/title NO-JIRA: add example of how to build from source

PR #827, opened 2025-03-10
NO-JIRA: Report the error when syncing STS secrets

danilo-gemoli:
When the controller fails to create an STS secret, it doesn't surface the error generated by the k8s client.
The error message is instead quite generic and difficult to troubleshoot:
```
time="2025-03-10T08:58:38Z" level=error msg="error syncing credentials: error syncing creds in mint-mode: error processing secret" controller=credreq cr=openshift-cloud-credential-operator/cluster-init secret=ci/cluster-init-cloud-credentials
```

### DIFF
--- a/pkg/aws/actuator/actuator.go
+++ b/pkg/aws/actuator/actuator.go
@@ -430,11 +430,11 @@ func (a *AWSActuator) syncSTSSecret(awsSTSIAMRoleARN string, cloudTokenPath stri
 		secret.Type = corev1.SecretTypeOpaque
 		return nil
 	})
-	sLog.WithField("operation", op).Info("processed secret")
+	sLog.WithField("operation", op).WithError(err).Info("processed secret")
 	if err != nil {
 		return &actuatoriface.ActuatorError{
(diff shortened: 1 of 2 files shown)


pkg/aws/actuator/actuator.go
```diff
@@ -434,7 +434,7 @@ func (a *AWSActuator) syncSTSSecret(awsSTSIAMRoleARN string, cloudTokenPath stri
 	if err != nil {
```
wking, line 434: the `sLog` above here seems like it could grow a `.WithField("error", err)` or some such as well, so the error is present in logs instead of the generic `processed secret` which currently doesn't distinguish between "hooray, success!" and "uhoh, failure!".

dlom:
/retitle NO-JIRA: Report the error when syncing STS secrets

pkg/aws/actuator/actuator.go
dlom: @danilo-gemoli I agree with Trevor here

danilo-gemoli: Agreed, done.

PR #850, opened 2025-04-17
OCPBUGS-55777: Sync OpenStack CA Bundles from legacy location

stephenfin:
In #780, we added support for syncing CA bundles from the root credential secret to the generated credential secrets when running on OpenStack clouds. This is a big improvement for OpenShift developers, since it hugely simplifies how we obtain these in other operators and components and removes the need for a number of controllers. However, as things stand, it doesn't help users as it introduces a second place that they must consider when rotating credentials.

Long-term, we would like to remove the CA bundle from the cloud-providers config map. Doing so requires some rework of CCCMO as well as investigation into potential issues caused by CCMs role in early stage bootstrapping. This isn't going to happen in OpenShift 4.19, so for now we opt to leave the current documentation around cert rotation as-is and simply sync the CA cert from the CCM config map to the root credential, if it's set. We can then revert this down the line if needed.

/hold

### DIFF
--- a/pkg/operator/secretannotator/openstack/reconciler.go
+++ b/pkg/operator/secretannotator/openstack/reconciler.go
@@ -163,13 +163,30 @@ func (r *ReconcileCloudCredSecret) Reconcile(ctx context.Context, request reconc
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


jstuever (approved):
/lgtm
/approve

PR #848, opened 2025-04-09
OCPBUGS-53429: Render: configure proxy on bootstrap static pod

jstuever:
Previously, the static pod on the bootstrap machine was not aware of the
cluster-wide proxy configuration. This caused the static pod to be
unable to reach sites that were not allowed directly from the bootstrap
host. This caused installation failure in Mint and Passthrough mode as
the bootstrap process depends on some credentials being populated.

This change injects the proxy env variables into the static pod manifest
when rendered. It also mounts the ca-trust from the bootstrap host into
the pod. This enables the cloud-credential-operator binary to use the
proxy as configured in the environment variables on the static pod.

Previously, the render command used string templates to generate some
manifests. This made it difficult to modify the values, forcing all
changes to adhere to the template structure.

This change migrates those templates to the relevant object structs and
uses yaml marshal to generate the manifests. This enables the values to
be changed easily on the objects. It purposefully uses the sigs yaml
module because it is capable of adhering to the json structure tags
producing concise yaml manifests as a result.

PR #855, opened 2025-04-28
OCPBUGS-53429: Render: configure proxy on bootstrap static pod

jstuever:
Previously, the render command used string templates to generate the
static pod. This made it difficult to modify the values, forcing all
changes to adhere to the templating interface.

This change migrates the static pod template to a pod struct and uses
yaml marshal to generate the manifest. This enables the values to be set
on the struct prior to generating the manifest. The sigs yaml module is
used purposefully because it is capable of adhering to the json metadata
labels and therefore produce more concise yaml manifests.

Previously, the static pod on the bootstrap machine was not aware of the
cluster-wide proxy configuration. This caused the static pod to be
unable to reach sites that were not allowed directly from the bootstrap
host. This caused installation failure in Mint and Passthrough mode as
the bootstrap process depends on some credentials being populated.

This change injects the proxy env variables into the static pod manifest
when rendered. It also mounts the ca-trust from the bootstrap host into
the pod. This enables the cloud-credential-operator binary to use the
proxy as configured in the environment variables on the static pod.

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -31,7 +31,9 @@ import (
 	"github.com/spf13/cobra"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	sigsyaml "sigs.k8s.io/yaml"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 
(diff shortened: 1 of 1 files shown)


jstuever:
This is a scaled back version of the original PR #848, where I attempted to convert both the static pod and the operator manifests to use structures. Unfortunately, one of the fields under the operator struct (in the openshift API) does not have the "omitempty" json metadata tag and is rendered as an empty string, which is an invalid value. This caused the operator manifest application to be rejected, leaving the credentials mode to be the default. As a result, passthrough and manual mode annotations were effectively ignored.

As opposed to trying to resolve this issue, I scaled back the effort. The original goal was to have the static pod as a struct so I could easily add the proxy components when applicable. I wanted to also do the same with the operator manifest simply to keep them in sync in terms of how they were handled. Due to the above issue, I have opted to ignore the operator manifest, leaving it as a template, and only focus on the static pod changes needed to accomplish the primary goal.

jstuever:
/override ci/prow/security
There is an existing bug to handle this.

jstuever:
/assign @dlom
Same concept as before, just less things I'm touching this time.

jstuever:
/hold
for manual oidc jobs

PR #864, opened 2025-05-28
OCPBUGS-57296: Remove duplicate cluster role rules

stephenfin:
We have a duplicate rule with a larger set of verbs defined later in the file.


stephenfin:
/test e2e-hypershift

This was failing last week but seems to be green now :crossed_fingers:

jianping-shu:
The regression tests passed on aws mint and aws sts clusters with cluster-bot built image.

PR #865, opened 2025-05-28
OCPBUGS-59937: secretannotator: Only pass clients to Reconciler, remove use of non-caching client

stephenfin:
Per $subject. This makes the clients in use a little more obvious (IMO).


jstuever:
@stephenfin I'm hesitant to mere this. While it may be beneficial to make the client use more clear, I don't see a functional reason to do so and am concerned that the change might have unintended consequences. I'm not sure it is worth the risk.

jstuever:
Disregard my last statement, I suddenly remember why we did this. Can you create and link a Jira bug for this?

jstuever (changes requested):
I apologize for taking so long to review this one. I wanted to make sure I fully understood how the LiveClient was being used, specifically in the aws platform.

The AWS actuator(s) need the live client because the functions that they use to build the aws clients need to access the infrastructure resource in the cluster, which is not cached by the RootCredClient. This is acceptable because the live client is only called when the actuator is created within NewAWSActuator(). My recommendation is to revert all of the changes relevant to that.

In contrast, the OpenStack reconciler had started using the LiveClient within the reconcile() function. As a result, it was using a non-cached client on a regular basis. I would keep all of the changes that enable it to use the RootCredClient within the reconcile function (such as adding the resources to the filter).

PR #876, opened 2025-06-18
OCPBUGS-53454: ccoctl: only add owned tag to azure resources on create

jstuever:
Prior to this change, ccoctl always added the owned tag to the resource groups and storage accounts. This caused errors when the number of owned tags from multpiple clusters exceeded the maximum tag limit.

This change makes it so the owned tag is only added when a resource group or storage account is actually created. Any additional clusters using these resources will not attempt to add their own owned tag. This enables significantly more clusters to share the same resources without reaching the limit.

### DIFF
--- a/pkg/cmd/provisioning/azure/create_managed_identities.go
+++ b/pkg/cmd/provisioning/azure/create_managed_identities.go
@@ -622,7 +622,7 @@ func createManagedIdentities(client *azureclients.AzureClientWrapper, credReqDir
 
 	// Ensure the installation resource group exists
 	if !dryRun {
-		err := ensureResourceGroup(client, installationResourceGroupName, region, resourceTags)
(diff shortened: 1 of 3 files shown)


jstuever:
/retest
flaky

jianping-shu:
pre-merge testing passed with cluster-bot build.
1. Azure Microsoft Entra Id cluster installed successfully and regression test passed.
In this case the OIDC RG and storage account were still created by ccoctl and added with openshift.io_cloud-credential-operator tag. 
2. Tested with OCPBUGS-53454 part of [OCP-64542](https://polarion.engineering.redhat.com/polarion/redirect/project/OSE/workitem?id=OCP-64542) , created the OIDC RG and storage account in advance then run the ccoctl, then OIDC RG and storage account didn't have the openshift.io_cloud-credential-operator tag any more

PR #896, opened 2025-08-04
CORS-4161: Add GCP Service Endpoint overrides

barbacbd:
pkg/cmd/provisioning:

** Temporarily updated to pass nil for service endpoints. This is a placeholder while ccoctl does not support user supplied service endpoints.

pkg/gcp/actuator:

** Actuator will get the infrastructure object and pass the endpoints during client creation.

pkg/operator/secretannotator/gcp:

** Reconciler will get the infrastructure object and pass the endpoints during client creation.

pkg/gcp/client.go:

** Client will now accept endpoints when creating the service cleints (ex: compute).


jstuever (changes requested):
In addition to the inline comments, this appears to be missing feature gate.

pkg/gcp/actuator/actuator.go
```diff
@@ -729,8 +729,18 @@ func (a *Actuator) buildReadGCPClient(cr *minterv1.CredentialsRequest) (ccgcp.Cl
...
+	infra, err := utils.GetInfrastructure(a.Client)
+	if err != nil {
+		return nil, err
+	}
+
+	endpoints := []configv1.GCPServiceEndpoint{}
+	if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.GCP != nil {
+		endpoints = infra.Status.PlatformStatus.GCP.ServiceEndpoints
+	}
+
```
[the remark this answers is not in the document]
jstuever: Having looked at this code more, this path is called every time reconcile() is called. As a result, this is causing unnecessary API requests quite regularly. I recommend moving this closer to the NewActuator() so that it's only pulled once when the operator starts. I understand the field is currently not expected to change. If we decide it can change at a later date, we should add functionality to restart the operator when the field changes.

pkg/util/featuregate.go
```diff
@@ -0,0 +1,100 @@
...
+	currentFeatureGates, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return nil, err
+	}
+
+	enabledFeatureGates := map[configv1.FeatureGateName]bool{}
+	for _, fg := range featureGates {
+		enabledFeatureGates[fg] = currentFeatureGates.Enabled(fg)
+	}
+	return enabledFeatureGates, nil
```
jstuever, line 56: Instead, let's have this function `GetCurrentFeatureGates` and return the `currentFeatureGates` from line 47. This will change the behavior in pkg/operator/controller.go as mentioned in the correlated comment.
```
return featureGateAccessor.CurrentFeatureGates()
```

pkg/operator/controller.go
```diff
@@ -107,7 +110,12 @@ func AddToManager(m, rootM manager.Manager, explicitKubeconfig string, coreClien
 			if infraStatus.PlatformStatus == nil || infraStatus.PlatformStatus.GCP == nil {
 				log.Fatalf("missing GCP configuration in platform status")
 			}
-			a, err = gcpactuator.NewActuator(m.GetClient(), rootM.GetClient(), infraStatus.PlatformStatus.GCP.ProjectID)
+			enabledFeatures, err := util.GetEnabledFeatureGates([]configv1.FeatureGateName{features.FeatureGateGCPCustomAPIEndpoints})
+			if err != nil {
+				return fmt.Errorf("error getting enabled feature gates: %v", err)
+			}
+			gcpCustomEndpointsEnabled, _ := enabledFeatures[features.FeatureGateGCPCustomAPIEndpoints]
```
jstuever, line 117: In conjunction with the comment in pkg/util/featuregate.go, modify this to get currentFeatureGates, and then check for gcpCustomEndpointsEnabled separately.
```
	currentFeatureGates, err := util.GetCurrentFeatureGates()
	if err != nil {
		return err
	}
```
	
That block should be included in the prior commit so it remains after the featuregate is removed. And then for this specific feature gate, you would add
```
	gcpCustomEndpointsEnabled := currentFeatureGates.Enabled(features.FeatureGateGCPCustomAPIEndpoints)
```

pkg/util/featuregate.go
```diff
@@ -0,0 +1,100 @@
...
+		log.Info("FeatureGates initialized", "knownFeatures", currentFeatureGates.KnownFeatures())
+	case <-time.After(1 * time.Minute):
+		log.Error(nil, "timed out waiting for FeatureGate detection")
+		return nil, fmt.Errorf("timed out waiting for FeatureGate detection")
+	}
+
+	currentFeatureGates, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return nil, err
+	}
```
jstuever, line 50: nit: this feels replicated from the case above.

jstuever (changes requested):

pkg/operator/secretannotator/gcp/reconciler.go
```diff
@@ -41,12 +43,18 @@ const (
 )
 
 func NewReconciler(c client.Client, mgr manager.Manager, projectName string) reconcile.Reconciler {
+	endpoints, err := gcputils.GetServiceEndpoints(c)
+	if err != nil {
+		return nil
+	}
+
```
jstuever, line 50: This is failing. Please add a log here so it's more obvious why.
```diff
	endpoints, err := gcputils.GetServiceEndpoints(c)
	if err != nil {
+        log.WithField("controller", constants.SecretAnnotatorControllerName).WithError(err).Error("unable to get service endpoints")
		return nil
	}
```

With the above logging in place, the issue is that the `c` client is failing on cache.
```
time="2025-09-26T22:27:09Z" level=info msg="Setting up secret annotator. Platform Type is GCP"
time="2025-09-26T22:27:09Z" level=error msg="unable to get service endpoints" controller=secretannotator error="error retrieving infrastructure containing service endpoints: the cache is not started, can not rea
d objects"
time="2025-09-26T22:27:09Z" level=fatal msg="unable to register controllers to the manager" error="must specify Reconciler"
```

pkg/gcp/actuator/actuator.go
```diff
@@ -78,16 +79,23 @@ type Actuator struct {
...
+	GCPEndpoints     []configv1.GCPServiceEndpoint
 }
 
 // NewActuator initializes and returns a new Actuator for GCP.
 func NewActuator(c, rootCredClient client.Client, projectName string) (*Actuator, error) {
+	endpoints, err := gcputils.GetServiceEndpoints(c)
+	if err != nil {
+		return nil, err
+	}
+
```
jstuever, line 92: I expect this would be failing similar to the secretannotator controller if we weren't failing on that controller.

jstuever (changes requested):

PR #901, opened 2025-08-25
OCPBUGS-29900:fix the Metric cco_credentials_mode issue

jianping-shu:
See [OCPBUGS-29900](https://issues.redhat.com/browse/OCPBUGS-29900) for the details.

Try to use AI tools to facilitate the bug fix.
(1) Tried deepwiki, its answers were positioning related code functions and providing the fix suggestions
https://deepwiki.com/search/for-aws-sts-cluster-metric-cco_6fe2f619-3e29-4d31-8aeb-5c85a8f589db

### DIFF
--- a/pkg/operator/metrics/metrics.go
+++ b/pkg/operator/metrics/metrics.go
@@ -2,6 +2,7 @@ package metrics
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jianping-shu:
To use utils.IsTimedTokenCluster is a better and more efficient solution than func credRequestIsPodIdentity which is called for every CR in func processCR. And it did fix the issue.
Need check if it is safe to remove func credRequestIsPodIdentity.

jianping-shu:
@jstuever @dlom please help review this PR when you got time, TIA

jstuever (changes requested):
My apologies for taking so long to review this PR. I wanted to make sure I fully understood the `manualpodidentity` use case. I don't think AI provided the right solution here. It is clear that `manualpodidentity` is intended to show that at least one secret is actually using short term tokens. However, this solution instead shows that the cluster is configured to use short term tokens without verifying that any of the secrets are actually configured to do so.

Instead, this PR needs to figure out if `credRequestIsPodIdenty()` is returning false when there are short term tokens being used, and if so, resolve why that is happening. As it is currently written, there is no logic for the GCP platform.

PR #922, opened 2025-09-29
CCO-714: Add preserve-existing-roles flag to ccoctl auzre

jstuever:
By default, ccoctl reconciles the roles assigned to a managed identity, removing any roles that are not explicitly defined in the CredentialsRequest. This can be problematic if a user has manually assigned additional roles to the identity for other purposes, as they will be removed on the next run.

This commit introduces a new `--preserve-existing-roles` flag to the Azure provisioning commands. When this flag is enabled, ccoctl will no longer remove existing role assignments. It will only ensure that the roles specified in the CredentialsRequest are present, making the operation additive rather than a full reconciliation.

Assisted-by: gemini-2.5-pro


huangmingxia:
Hi @jstuever, you’re right — the job failure is unrelated to this PR. I’ll continue working on this test today ^ ^

PR #986, opened 2026-03-11
CCO-770: Docs: consolidate and simplify short-term token workflows

jstuever:
Previously, the documentation for the short-term token workflows was split into three docs, one for each supported cloud. They were also more complicated than they needed to be. This caused unnecessary burden on both the maintainers of the repo as well as users attempting to follow them.

This change combines the three docs into a single workflow. This reduces the maintenance burden on repo maintainers. It also simplifies the installation workflows by having the oc extract use the --included and --install-config flags in addition to outputting the manifests directly into the installation folder directly

## Summary by CodeRabbit

* **Documentation**
  * Consolidated short-term credentials content into a single unified guide covering AWS, Azure, and GCP short-lived credentials.
  * Removed separate per-cloud workload-identity and STS guides in favor of the unified documentation.
  * Replaced the multi-link cloud list in the README with a single link to the short-term credentials guide.
  * Updated the Support Matrix: expanded entries, added IBM Cloud and Nutanix, and added a footnote about Azure Mint mode removal.
  * No changes to public/exported APIs.


jstuever:
LOL. I forgot to do a spellcheck pass, and CodeRabbit called me out on it.

PR #1016, opened 2026-04-22
CCO-711: docs: improve the filter when getting the capi pod

jstuever:
Previously, a validation step in sts-migrate-to-private-bucket.md fetched the name of the CAPI pod from the cluster. However, it was fetching more than the desired pod, which caused a followup command to fail.

This change adds additional filters to the command fetching the pod in order to ensure it is only fetching the desired pod. As a result, the followup command will now succeed.

## Summary by CodeRabbit

* **Documentation**
  * Enhanced STS private bucket migration documentation with additional pod selection constraints. Users migrating to private buckets will now find more precise guidance for identifying the correct Kubernetes pod when extracting authentication tokens, ensuring more reliable and accurate token retrieval during the migration process.

### DIFF
--- a/docs/sts-migrate-to-private-bucket.md
+++ b/docs/sts-migrate-to-private-bucket.md
@@ -121,6 +121,7 @@ TOKEN_PATH=$(oc get secrets aws-cloud-credentials \
 # Get Controler's pod
 CAPI_POD=$(oc get pods -n openshift-machine-api \
     -l api=clusterapi \
+    -l k8s-app=controller \
     -o jsonpath='{.items[*].metadata.name}')
 
 # Extract tokens from the pod


docs/sts-migrate-to-private-bucket.md
```diff
@@ -121,6 +121,7 @@ TOKEN_PATH=$(oc get secrets aws-cloud-credentials \
 # Get Controler's pod
 CAPI_POD=$(oc get pods -n openshift-machine-api \
     -l api=clusterapi \
+    -l k8s-app=controller \
     -o jsonpath='{.items[*].metadata.name}')
```
damdo, line 125: @jstuever I was reading PRs and just noticed this. Which is fine as a change per se.

This is called CAPI_POD but I suspect this is in reality MAPI, and the `api=clusterapi` bit is because of legacy reasons where MAPI has been forked out of upstream CAPI originally.

I'd maybe rename `CAPI_POD` to `MAPI_POD` or something, and maybe add a comment to explain this

PR #1043, opened 2026-06-25
OCPBUGS-87829: Scope minted AWS IAM policies to cluster-owned resources

dlom:
xref: OCPBUGS-87829

Add an implicit `StringEquals` condition to every `Allow` statement in minted IAM user policies, requiring the standard `kubernetes.io/cluster` `infraName` tag (with a value of `"owned"`). This restricts minted credentials to resources tagged as belonging to the cluster, reducing the blast radius if credentials are compromised (`CVE-2026-10843`).

The CVE advisory suggests tightening `Resource` ARNs from "*" to specific ARNs, but this isn't feasible: `CredentialsRequest` manifests are authored before cluster infrastructure exists, so resource ARNs for specific clusters are not known when the CR is authored.

the `aws:ResourceTag` condition key is not evaluated by all AWS api actions. For actions that do not support it, AWS silently ignores the condition and the action is allowed unconditionally. These actions are never destructive. This means that the fix only affects destructive actions, the same category of actions explicitely called out by the CVE.

The condition is not applied to `Deny` statements (this would actually narrow and weaken the policy), nor is it applied to the auto-generated `iam:GetUser` self-lookup statement, which is already scoped to just the own user's ARN.

## Summary by CodeRabbit

* **New Features**
  * Generated minted-user IAM policies are now infrastructure-aware, adding `aws:ResourceTag/kubernetes.io/cluster/<infra>` `StringEquals`-scoped permissions alongside unscoped ones where applicable.
* **Bug Fixes**
  * Cluster tag validation and user tagging now consistently use the expected tag value, and statement conditions are preserved while `Deny` rules remain unchanged.
* **Tests / Automation**
  * Added unit tests covering infra scoping behavior, action classification, and statement splitting.
  * Added an automation script to verify payload AWS action coverage against the expected scoped/unscoped action sets.

### DIFF
--- a/hack/verify-action-maps.sh
+++ b/hack/verify-action-maps.sh
@@ -1,0 +1,165 @@
+#!/usr/bin/env bash
+#
+# Extracts AWS CredentialsRequest manifests from an OpenShift release payload
+# and compares the IAM actions they declare against the hardcoded classification
+# maps in pkg/aws/utils.go.
+#
+# Usage:
+#   ./hack/verify-action-maps.sh [RELEASE_IMAGE]
+#
+# If RELEASE_IMAGE is omitted, the script pulls the latest 5.0.0 nightly.
+# Set REGISTRY_AUTH_FILE to point to a pull secret (default: $HOME/.pull-secret).
+# Requires: oc, jq, yq, curl
+
+set -euo pipefail
+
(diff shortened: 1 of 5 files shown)


dlom:
/verified by 4.22 and 5.0 payload nightly tests

PR #1045, opened 2026-06-30
OCPBUGS-87828: Scope minted AWS IAM policies to cluster-owned resources 

amitesh1201:
xref: OCPBUGS-87828
Backport of https://github.com/openshift/cloud-credential-operator/pull/1043 from master to release-4.22 to fix CVE-2026-10843

---

Add an implicit `StringEquals` condition to every `Allow` statement in minted IAM user policies, requiring the standard `kubernetes.io/cluster` `infraName` tag (with a value of `"owned"`). This restricts minted credentials to resources tagged as belonging to the cluster, reducing the blast radius if credentials are compromised (`CVE-2026-10843`).

The CVE advisory suggests tightening `Resource` ARNs from "*" to specific ARNs, but this isn't feasible: `CredentialsRequest` manifests are authored before cluster infrastructure exists, so resource ARNs for specific clusters are not known when the CR is authored.

Not all AWS actions support the `aws:ResourceTag` condition. Actions that create new resources fail (because the resource doesn't exist yet). Describe and List actions lack resource-level permissions. Some services don't support it at all. However, within the payload, the list is both finite and rarely changed. As such, we can easily classify the entire list into supported/incompatible categories. Unrecognized actions produce an error state, with directions pointing the developer to update the classification list in the CCO.

The condition is not applied to `Deny` statements (this would actually narrow and weaken the policy), nor is it applied to the auto-generated `iam:GetUser` self-lookup statement, which is already scoped to just the own user's ARN.

### Changes
- `pkg/aws/utils.go`: Exports infra resource tag prefix/value constants, defines scoped and unscoped AWS action classification maps, adds `SupportsInfraResourceTagCondition()` function
- `pkg/aws/actuator/actuator.go`: `getDesiredUserPolicy` now partitions actions by `aws:ResourceTag` compatibility and injects infrastructure-scoped conditions on supported Allow statements
- `pkg/aws/actuator/actuator_test.go`: Unit tests covering infra scoping behavior, action classification, and statement splitting
- `pkg/aws/utils_test.go`: Tests for payload action classification and map disjointness
- `hack/verify-action-maps.sh`: Verification script to validate action map coverage against the release payload (adjusted to 4.22.0 nightly stream)

### Testing
- Cherry-pick applied cleanly from master (commit bdf1b53)
- Unit tests pass: `go test ./pkg/aws/...` and `go test ./pkg/aws/actuator/...`
- Original PR was verified against 4.22 and 5.0 payload nightly blocking tests

### DIFF
--- a/hack/verify-action-maps.sh
+++ b/hack/verify-action-maps.sh
@@ -1,0 +1,165 @@
+#!/usr/bin/env bash
+#
+# Extracts AWS CredentialsRequest manifests from an OpenShift release payload
+# and compares the IAM actions they declare against the hardcoded classification
+# maps in pkg/aws/utils.go.
+#
+# Usage:
+#   ./hack/verify-action-maps.sh [RELEASE_IMAGE]
+#
+# If RELEASE_IMAGE is omitted, the script pulls the latest 4.22.0 nightly.
+# Set REGISTRY_AUTH_FILE to point to a pull secret (default: $HOME/.pull-secret).
+# Requires: oc, jq, yq, curl
+
+set -euo pipefail
+
+AUTHFILE="${REGISTRY_AUTH_FILE:-$HOME/.pull-secret}"
+RELEASE_IMAGE="${1:-}"
+TMPDIR=$(mktemp -d)
+trap 'rm -rf "$TMPDIR"' EXIT
+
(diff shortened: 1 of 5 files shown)


dlom:
/verified by 4.22 payload test on original PR

PR #1048, opened 2026-07-01
CCO-847: Resolve ci/prow/security job failures

dlom:
xref: [CCO-847](https://redhat.atlassian.net/browse/CCO-847)

Persistent ci/prow/security failures should be resolved by these dependency updates

## Summary by CodeRabbit

* **Chores**
  * Updated the Go toolchain version and refreshed several indirect dependencies to newer releases.
  * Removed an unnecessary dependency override, simplifying version management.

PR #1062, opened 2026-07-07
MULTIARCH-5990: IBM Cloud Staging Endpoint Support for `ccoctl`

taliandre49:
Add a `--endpoint flag to `ccoctl ibmcloud` subcommands to support IBM Cloud credential provisioning against Staging and other API endpoints. This enables creation of credential creation workflows on alternative endpoints.

## Changes

**New `--api-endpoint` flag** added to the following `ccoctl ibmcloud` subcommands:
- `create-service-id`
- `delete-service-id`
- `refresh-keys`

Accepts a base domain (e.g. `test.cloud.ibm.com`) and derives the full service
URLs from it:
- IAM authenticator, identity, and policy management -> `https://iam.<endpoint>`
- Resource Manager -> `https://resource-controller.<endpoint>`

When omitted, all clients use their SDK production defaults unchanged.

**`pkg/ibmcloud/client.go`**:
- Added `APIEndpoint string` to `ClientParams`
- `NewClient` constructs service URLs from the endpoint when set

**`.gitignore`**: Added `/ccoctl-*` to ignore platform-specific build artifacts

### Notes

- The `--api-endpoint` flag is opt-in and has no effect on normal production usage
- Alternative Endpoints are only meaningful with the corresponding platform's IBM Cloud API key (i.e staging API endpoint and an API key from the staging platform)
- Tested against current repository tests (all pass) and locally to confirm that service IDs do indeed get created on staging.

## Summary by CodeRabbit

* **New Features**
  * Added an `--api-endpoint` option/flag to IBM Cloud provisioning commands (create service ID, delete service ID, refresh keys) to override the IBM Cloud API base domain used for requests.
  * IBM Cloud client now supports configurable API endpoint routing for IAM and resource-manager calls.
* **Tests**
  * Added coverage for API endpoint normalization and client URL construction behavior.
* **Chores**
  * Updated version control ignore rules to exclude paths matching `/ccoctl-*` in addition to the existing `/ccoctl` entry.

### DIFF
--- a/pkg/ibmcloud/client.go
+++ b/pkg/ibmcloud/client.go
@@ -2,6 +2,7 @@ package ibmcloud
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/IBM/go-sdk-core/v5/core"
 	identityv1 "github.com/IBM/platform-services-go-sdk/iamidentityv1"
@@ -80,6 +81,23 @@ func (i *ibmcloudClient) CreatePolicy(options *pmv1.CreatePolicyOptions) (*pmv1.
 	return i.pmClient.CreatePolicy(options)
 }
 
+// validateAPIEndpoint ensures the provided API endpoint is a bare hostname/domain with no
+// scheme or trailing slash (e.g. "test.cloud.ibm.com"), returns a cleaned value or an error.
+func validateAPIEndpoint(endpoint string) (string, error) {
+	// Strip a scheme the user may have accidentally included.
+	for _, scheme := range []string{"https://", "http://"} {
+		endpoint = strings.TrimPrefix(endpoint, scheme)
+	}
+	// Strip trailing slashes.
+	endpoint = strings.TrimRight(endpoint, "/")
+
+	// After stripping, the value must not still contain "://"
+	if strings.Contains(endpoint, "://") {
+		return "", fmt.Errorf("invalid api-endpoint %q: must be a bare domain (e.g. test.cloud.ibm.com), not a full URL", endpoint)
+	}
+	return endpoint, nil
+}
+
 func NewClient(apiKey string, params *ClientParams) (Client, error) {
 	authenticator := &core.IamAuthenticator{
 		ApiKey: apiKey,
@@ -92,8 +110,13 @@ func NewClient(apiKey string, params *ClientParams) (Client, error) {
 			agentText = params.InfraName
 		}
 		if params.APIEndpoint != "" {
-			iamURL = fmt.Sprintf("https://iam.%s", params.APIEndpoint)
-			resourceManagerURL = fmt.Sprintf("https://resource-controller.%s", params.APIEndpoint)
+			endpoint, err := validateAPIEndpoint(params.APIEndpoint)
+			if err != nil {
+				return nil, err
+			}
+			// Both the IAM Identity and IAM Policy Management APIs are served from the same iam.* host
+			iamURL = fmt.Sprintf("https://iam.%s", endpoint)
+			resourceManagerURL = fmt.Sprintf("https://resource-controller.%s", endpoint)
 		}
 	}
 
--- a/pkg/ibmcloud/client.go
+++ b/pkg/ibmcloud/client.go
@@ -27,6 +27,9 @@ type Client interface {
 
 type ClientParams struct {
 	InfraName string
+	// APIEndpoint is the base domain of the IBM Cloud API to target (e.g. "test.cloud.ibm.com").
+	// When empty, the SDK defaults (production cloud.ibm.com endpoints) are used.
+	APIEndpoint string
 }
 
 type ibmcloudClient struct {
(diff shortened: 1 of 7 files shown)


taliandre49:
Snyk code scan passed `Total issues: 0`, my  source changes are clean
The dependency scan failed on packages like golang.org/x/net@0.49.0, std/crypto/x509@1.26.3, github.com/microsoft/kiota-http-go@1.5.4, none of which this PR touched. I believe this test fails due to merging against master@97109748 which is before #1048 (the PR that resolves this issue) lands. Can someone with the necessary permissions override this please!

clnperez:
I'm wondering if this should be, instead of a staging flag, an API endpoint flag or something similar.

taliandre49:
@clnperez I did a staging flag so you could specify when you run any ccoctl command for example:

`
./ccoctl ibmcloud create-service-id --credentials-requests-dir ./credreqs --name <name> --resource-group-name <resource-group-name> --staging
`
But open to hear your approach! 

--update: ahhh i see what you're getting at, an API (url) endpoint flag for the IAM/resource manager!!

dlom:
/retest

flaky test

dlom:
/retitle [MULTIARCH-5990](https://redhat.atlassian.net/browse/MULTIARCH-5990): IBM Cloud Staging Endpoint Support for `ccoctl`

taliandre49:
Status: Refactoring code to use more flexible approach of an API endpoint flag, will push shortly

pkg/ibmcloud/client.go
```diff
@@ -25,8 +25,17 @@ type Client interface {
 	ListResourceGroups(*resourcemanagerv2.ListResourceGroupsOptions) (*resourcemanagerv2.ResourceGroupList, *core.DetailedResponse, error)
 }
 
+// Staging endpoint defaults. These are the IBM Cloud pre-production (staging) API endpoints.
+const (
+	StagingIAMURL              = "https://iam.test.cloud.ibm.com"
+	StagingPolicyManagementURL = "https://iam.test.cloud.ibm.com"
```
dlom: are these two URLs supposed to be the same?

pkg/ibmcloud/client.go
```diff
@@ -113,6 +138,9 @@ func NewClient(apiKey string, params *ClientParams) (Client, error) {
 	resourceManagerV2Options := &resourcemanagerv2.ResourceManagerV2Options{
 		Authenticator: authenticator,
 	}
+	if staging {
+		resourceManagerV2Options.URL = StagingResourceManagerURL
+	}
 
 	resourceManagerV2Client, err := resourcemanagerv2.NewResourceManagerV2(resourceManagerV2Options)
```
dlom, line 165: maybe out of scope here, but I noticed that this `err` is never actually checked

pkg/ibmcloud/client.go
taliandre49: this is omitted in the latest code changes! Now appends the base domain for the API endpoint provided

pkg/ibmcloud/client.go
taliandre49: latest code changes includes a catch for this err

dlom:
(no AWS-related changes in this PR, and this is a known flaky test)

pkg/ibmcloud/client.go
```diff
@@ -81,20 +84,36 @@ func NewClient(apiKey string, params *ClientParams) (Client, error) {
...
-	if params != nil && params.InfraName != "" {
-		agentText = params.InfraName
-	}
 	userAgentString := fmt.Sprintf("OpenShift/4.x Cloud Credential Operator: %s", agentText)
 
 	serviceClientOptions := &pmv1.IamPolicyManagementV1Options{
 		Authenticator: authenticator,
 	}
+	if iamURL != "" {
+		serviceClientOptions.URL = iamURL
```
dlom, line 138: Are the policy management API and the identity API always going to live at the same URL? I don't know enough about IBM's APIs here tbh

pkg/ibmcloud/client.go
```diff
@@ -81,20 +84,36 @@ func NewClient(apiKey string, params *ClientParams) (Client, error) {
...
 	}
+
+	agentText := "defaultAgent"
+	var iamURL, resourceManagerURL string
+	if params != nil {
+		if params.InfraName != "" {
+			agentText = params.InfraName
+		}
+		if params.APIEndpoint != "" {
+			iamURL = fmt.Sprintf("https://iam.%s", params.APIEndpoint)
```
dlom: I agree with the automated review's findings here, though I don't think it's a nitpick. Would definitely like something more defensive here if the user passes a bad URL (e.g. leading `https://` or trailing slash)

pkg/ibmcloud/client.go
dlom, line 101: Nit: now that this function is doing some more complicated things, I'd love a unit test for checking things like verifying constructed URLs are sane and verifying that URLs aren't overridden when the endpoint is empty. I don't think this code is wrong, but I would like to prevent potential future regressions

pkg/ibmcloud/client.go
taliandre49: Correct yes they are both served by the same URL. Both the IAM Identity and IAM Policy Management APIs are served from the same iam.* host on IBM Cloud!

taliandre49: added a comment in the code to clarify this!

pkg/ibmcloud/client.go
taliandre49: Latest code changes include unit tests

pkg/ibmcloud/client.go
taliandre49: I agree as well, implemented this!

dlom:
@taliandre49 verify job is a real failure, the security job is failing again for unrelated reasons to this pr (I recommend running `make verify` before pushing up to github)

taliandre49:
@dlom running the update!

taliandre49:
Ran the update!

dlom:
/verified by this is backwards compatible/new functionality, no old behavior is affected

PR #1075, opened 2026-08-17
SPLAT-2899: support per-component credential overrides in openshift-config

rvanderp3:
Allow administrators to provide per-component vSphere credentials by placing override secrets in the openshift-config namespace. Override secrets are mapped to their target CredentialsRequest via annotations:

  cloudcredential.openshift.io/target-secret-namespace: <namespace>
  cloudcredential.openshift.io/target-secret-name: <secret-name>

When a CredentialsRequest is processed, the actuator lists secrets in openshift-config and looks for one whose target annotations match the CredentialsRequest's spec.secretRef. If found, the override secret's data is used; otherwise the shared root credential in kube-system/vsphere-creds is used as a fallback.

This enables least-privilege, per-component credentials for vSphere without breaking backward compatibility: clusters without override secrets continue to work exactly as before, with all components receiving the shared root credential.

Override secret example:

  apiVersion: v1
  kind: Secret
  metadata:
    name: anything-admin-wants
    namespace: openshift-config
    annotations:
      cloudcredential.openshift.io/target-secret-namespace: openshift-machine-api
      cloudcredential.openshift.io/target-secret-name: vsphere-cloud-credentials
      cloudcredential.openshift.io/mode: passthrough
  data:
    <vcenter>.username: ...
    <vcenter>.password: ...

Override secrets must carry the cloudcredential.openshift.io/mode annotation (same as the root credential secret) to be accepted.

The credentials request controller is also updated to watch for override secret changes in openshift-config so that create/update/delete events trigger reconciliation of all CredentialsRequests.

Assisted-by: Claude Code (claude-opus-4-20250725)

## Summary by CodeRabbit

* **New Features**
  * Added support for per-component vSphere credential override Secrets.
  * Automatically detects matching override credentials and reconciles affected credential requests.
  * Uses component-specific credentials when available, with fallback to shared root credentials.
  * Validates override targeting information and reports errors for incomplete configurations.

* **Bug Fixes**
  * Improved credential refresh behavior when override Secrets are added or updated.
  * Added clearer handling for missing or invalid credential Secrets.


rvanderp3:
/hold

cc @jcpowermac @vr4manta

PR #1068, opened 2026-07-20
CORS-4524: Enable GCP custom universe domain support for CCO

patrickdillon:
This PR adds plumbing to set universe domain on GCP clients as well as updating assumptions on service account name formatting. Doing this enables openshift operation on Google Cloud Dedicated, GCP's sovereign cloud offering. 

Additionally, this PR updates the relevant packages to move off of deprecated functions that were being used for authentication.

For more details, check the relevant commit messages.

For testing, we can check GCP e2e for any regressions. For testing GCD, once all the operator PRs are open, which should happen very soon, we can use `/testwith` to test in the environment.

Due to the priority of this, we intend to backport this functionality to 4.22.

See https://github.com/openshift/enhancements/pull/1977 for more details.

## Summary by CodeRabbit

* **New Features**
  * Added support for Google Cloud “universe domains” in the GCP client.

* **Bug Fixes**
  * Fixed service account email formatting for domain-scoped project identifiers.
  * Improved credential handling by deriving/preserving credential types when loading JSON credentials.

* **Tests**
  * Added unit tests for domain-aware service account email formatting.
  * Updated GCP client mocks and related tests to include universe domain lookups.

* **Chores**
  * Updated local build to Go 1.26 and refreshed Go dependency versions.

PR #1083, opened 2026-08-27
OCPBUGS-114669: Enable GCP custom universe domain support for CCO

rochacbruno:
Backport of #1068 to release-4.22.

Cherry-picks the GCP client and service-account changes that add support for custom (alternate) universe domains:
- Set the universe domain on GCP API clients
- Support domain-scoped service account email formatting
- Replace the deprecated `CredentialsFromJSON` usage

It also bumps `google.golang.org/api` v0.252.0 -> v0.258.0, the minimal version providing `option.WithAuthCredentialsJSON` and `option.CredentialsType` used by the cherry-picked code. `k8s.io/api` stays at v0.35.2 and `github.com/openshift/api` is unchanged.

Note: the companion ccoctl change (#1073) is intentionally NOT included here. It depends on the `GCPPlatformStatus.UniverseDomain` field from openshift/api #2963 ([CORS-4417](https://redhat.atlassian.net/browse/CORS-4417)), which has not yet been backported to openshift/api release-4.22.


Dockerfile.local
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.21 AS builder
+FROM golang:1.26 AS builder
```
tthvo:
```suggestion
FROM golang:1.25 AS builder
```

I wonder if we can use go v1.25 instead here since 4.22 release is built with go v1.25? My guess is "yes" based on: https://github.com/googleapis/google-api-go-client/blob/182ae992632d7ba19b5d3c60c71bda28fbfabca1/go.mod#L3

rochacbruno: Good catch, applied. Switched Dockerfile.local to `golang:1.25` since release-4.22 is a Go 1.25 release (the go.mod `go` directive is 1.25.0, and google.golang.org/api v0.258.0 only requires Go 1.25). The `golang:1.26` came from the #1068 cherry-pick, which targets master/5.0.

rochacbruno:
On the failing `ci/prow/security` (Snyk) check: the 22 findings are all pre-existing on release-4.22 and are not introduced by this PR.

- This PR only changes `google.golang.org/api` (0.252.0 -> 0.258.0), `enterprise-certificate-proxy` (0.3.6 -> 0.3.7), `genproto/rpc` (date bump), and `protobuf` (1.36.10 -> 1.36.11). None of these appear in Snyk's findings.
- Every flagged package is at the identical version on the release-4.22 baseline (e.g. `grpc` v1.79.3, `otel` v1.40.0, `golang.org/x/net` v0.49.0 which is pinned via a `replace` directive, `aws-sdk s3` v1.96.0), plus a set of Go stdlib CVEs tied to the CI toolchain (Go 1.25.12, fixed in 1.25.13). A fresh Snyk scan of release-4.22 HEAD would report the same issues; earlier PRs show green only because Snyk uses a live vulnerability DB and these CVEs were published more recently.

Addressing these belongs in a separate, repo-wide dependency bump rather than this targeted backport.

tthvo (commented):
/lgtm

tthvo:
/payload-job periodic-ci-openshift-release-main-ci-4.22-upgrade-from-stable-4.21-e2e-gcp-ovn-upgrade

Dockerfile.local
rochacbruno: @tthvo I will have to bump to a new golang api version, 0.258.0 has a bug that was fixed on 0.288

rochacbruno:
The `e2e-gcp-manual-oidc` failure was `ccoctl` erroring with:

```
Failed to initiate GCP client: dialing: multiple credential options provided
```

Root cause is the `google.golang.org/api` version, not the cherry-picked CCO code. This branch had pulled in `v0.258.0`, whose `DialSettings.Validate()` counts **both** `AuthCredentials` and `AuthCredentialsJSON` toward the credential-option total. `NewClient` passes `option.WithAuthCredentialsJSON(...)`, and for the new-auth gRPC/storage clients that JSON is auto-converted into an `auth.Credentials`, so both fields end up set and validation rejects the dial as having "multiple credential options".

`v0.288.0` fixes this by no longer counting `AuthCredentials` in that total (it is treated as a special case for the oauth2 -> `cloud.google.com/go/auth` conversion). This is the same `google.golang.org/api` version already on `master`, where the identical `NewClient` code passes `e2e-gcp-manual-oidc`.

I bumped `google.golang.org/api` to `v0.288.0` (`k8s.io/api` stays pinned at `v0.35.2`, `github.com/openshift/api` unchanged) and cleaned up the intermediate commits. Build is green locally.

Dockerfile.local
tthvo: Ah, right, thanks for checking that :pray: The version `v0.258.0` was indeed retracted (see [here](https://github.com/googleapis/google-api-go-client/blob/182ae992632d7ba19b5d3c60c71bda28fbfabca1/go.mod#L5)) due to that bug you described below :D

tthvo (commented):
/lgtm 

+1 for the failure analysis https://github.com/openshift/cloud-credential-operator/pull/1083#issuecomment-5484733885 🚀. This exact issue was filed and addressed in https://github.com/googleapis/google-cloud-go/issues/13503.

tthvo:
/payload-job periodic-ci-openshift-release-main-ci-4.22-upgrade-from-stable-4.21-e2e-gcp-ovn-upgrade

rochacbruno:
The `google.golang.org/api` v0.288.0 bump fixed the credential error. In this run the cluster provisioned cleanly, `ccoctl` created the GCP resources without issue, and 2017 tests passed (the "multiple credential options provided" failure is gone).

The suite went red on two blocking failures that are unrelated to this backport and to CCO:

- `[sig-api-machinery] CRDValidationRatcheting ... MUST fail to update a resource due to JSONSchema errors on unchanged uncorrelatable fields`
- `[sig-node] [DRA] kubelet [Feature:DynamicResourceAllocation] on single node retries pod scheduling after creating device class`

Both are generic Kubernetes/OpenShift platform conformance tests. The only CCO-related failure (`CCO metrics endpoint validation`, PolarionID 88196) is an informing/non-blocking test and is unrelated to the universe-domain change, which only affects `ccoctl` GCP client construction, not the operator's metrics endpoint.

Re-running the job to clear the flakes.

/test e2e-gcp-manual-oidc

PR #1065, opened 2026-07-10
CCO-849: bump go toolchain to 1.26.5 to resolve snyk complaints

dlom:
ci/prow/security is failing... again...

## Summary by CodeRabbit

* **Chores**
  * Updated the Go toolchain to version 1.26.5.

PR #1085, opened 2026-09-01
OCPBUGS-112282: In testing skip pod-identity-webhook checks on external platform

nader-ziada:
On platform-external clusters CCO runs as a no-op and does not deploy the pod-identity-webhook, but getIaasPlatform resolves External to the underlying IaaS name (e.g. aws), causing the PSA and readOnlyRootFilesystem conformance tests to assert on webhook pods that never exist.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated platform-specific pod identity webhook checks to correctly exclude external infrastructure platforms.
  * Improved AWS webhook reconciliation test handling for unsupported platform types.

### DIFF
--- a/test/extend/cloudcredential.go
+++ b/test/extend/cloudcredential.go
@@ -292,7 +292,7 @@ data:
 		o.Expect(seccompProfileType).To(o.Equal("RuntimeDefault"))
 		//Check IAAS platform type
 		iaasPlatform := checkPlatform(oc)
-		if iaasPlatform == "aws" || (iaasPlatform == "azure" && isWorkloadIdentityCluster(oc)) || iaasPlatform == "gcp" {
+		if !isPlatformExternal(oc) && (iaasPlatform == "aws" || (iaasPlatform == "azure" && isWorkloadIdentityCluster(oc)) || iaasPlatform == "gcp") {
 			g.By(fmt.Sprintf("2.Check pod-identity-webhook pod for %s", iaasPlatform))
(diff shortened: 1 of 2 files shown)


mtulio (commented):
Hi, LGTM for surgical fix for https://redhat.atlassian.net/browse/OCPBUGS-112282, althrough for long term we need to track the root cause: review the [function](https://github.com/openshift/cloud-credential-operator/blame/b187feee66f4ce0f992059b21a97a2ae88e4cdd9/test/extend/cloudcredential_util.go#L165-L178) `getIaasPlatform()` which uses `platformName` on platform type External to define the infrastructure, which isn't the better approach as `platformName` can assume any value, including "aws", which will change the behavior of entire test cases, which isn't expected.

Example of this jobs are those for [aws](https://prow.ci.openshift.org/?job=*openshift-release-*e2e-external-aws*), which is expected to be no-op but will assume platform type `aws`.

/lgtm

nader-ziada:
the `security` job failure is because of a pre-existing security-gate failure. All runs since July 30 have failed the same way. The findings are primarily from Go stdlib 1.26.4 and [google.golang.org/grpc](http://google.golang.org/grpc) 1.82.1; the log indicates fixes in Go 1.26.6+ and gRPC 1.83.1+.

There is already a fix in:  https://github.com/openshift/cloud-credential-operator/pull/1065 but its still open

nader-ziada:
Hi @dlom, are you able to approve please?

